PR #8, opened 2026-08-03
Bundle the frontend into one binary with production CLI startup

luhe19001:
## Context

Running the product needed Node, Vite, and a separate API process. This change produces one executable that serves the compiled React app and the Go API from the same loopback origin, so a release build is enough to run the full product locally.

Host/origin hardening, CI for the release path, and published artifacts are intentionally out of scope.

## Changes

- `make release` builds the frontend, stages it into the Go module for embedding, and writes a single `bin/coslash` binary; missing `index.html` fails the build clearly.
- The binary serves real frontend files (including hashed assets), falls back to `index.html` for client routes like `/coslash`, and returns 404 for missing asset paths and unrouted `/api/*` paths.
- Startup listens on loopback only, binds before opening the browser, and supports `--version`, `--port`, and `--no-open`, with actionable errors for port conflicts and browser-open failures.
- Root README is a short quick start for build, run, and local develop.

## Test

- `make release` — passed
- `go build ./...` — passed
- Manual: release binary on `--port 8899 --no-open`
  - `/` and `/coslash` → 200 HTML
  - hashed `/assets/*` → 200
  - missing hashed asset → 404
  - `/api/sessions` → 200 JSON
  - `/api/bogus` → 404
  - `--version`, bad `--port`, and port conflict → actionable errors


milanshen:
Reduce inline comments, only use inline comments for todos or special cases, not as overall logic or plan

collector/Makefile
```diff
@@ -1,8 +1,27 @@
...
+# Go cannot embed $(FRONTEND)/dist directly, so copy it into the module first.
+stage:
+	npm --prefix $(FRONTEND) ci
+	npm --prefix $(FRONTEND) run build
+	@test -f $(FRONTEND)/dist/index.html || { \
+		echo "error: the frontend build produced no $(FRONTEND)/dist/index.html"; exit 1; }
+	rm -rf $(STAGED)
+	mkdir -p $(STAGED)
+	cp -R $(FRONTEND)/dist/. $(STAGED)/
+	touch $(STAGED)/.gitkeep
```
milanshen, line 23: [P2] After this target runs, `internal/web/dist` remains populated. Neither `build` nor `run` removes it, so a later `make build` or `make run` embeds the previous release frontend. I reproduced `make release && make build`: `/` still returns 200 instead of the intended 503; after frontend edits this is silently stale. Please clear staged assets before non-release builds/runs, or otherwise isolate release-only embeds.

README.md
```diff
@@ -1,2 +1,32 @@
...
+| `--port N` | serve on port N (default `8787`) |
+| `--no-open` | do not open the browser |
+| `--version` | print the version and exit |
+
+## Develop
+
+In two terminals:
+
+```sh
+cd collector && make run
```
milanshen, line 28: [P2] `make run` now starts a server that opens `http://127.0.0.1:8787` by default. On a clean checkout that endpoint serves the intentional 503 because no frontend is embedded; the Vite UI is at port 5173 and Vite is not configured to open it. Please make the dev target use `--no-open` and document or open the Vite URL.

collector/Makefile
luhe19001: Good catch, fixed. Added an `unstage` target that empties `internal/web/dist`; `build`, `run`, and `clean` depend on it. Your repro now gives the intended 503 after `make build`.

README.md
luhe19001: Fixed. `make run` now passes `--no-open`, and the README points dev at Vite on http://localhost:5173. I documented the URL rather than setting `server.open` — happy to switch if you would rather it open.

luhe19001:
Trimmed the comments in `main.go`, `web.go`, and `browser.go` down to the non-obvious bits. The ones in `api.go` are pre-existing text this PR only moved, so I left them as they were.

PR #9, opened 2026-08-03
Speed up session loading with windowed collection

milanshen:
## Context

The session page parsed every historical transcript before applying client-side time filtering. This made bounded views pay the full-history collection cost and scaled poorly as transcript history grew.

## Changes

- Pass the active UI time window to `/api/sessions`, while keeping All unbounded.
- Prefilter Claude session groups and Codex descendant graphs before full parsing, preserving live sessions and complete subagent data.
- Read one extra day for fork context; forks whose parents fall outside that buffer retain conservative full usage.
- Use a bounded collection for the startup synthesis sweep.

## Test

- `cd collector && go build ./...` — passed
- `cd frontend && npx oxlint src/pages/coslash/CoslashPage.tsx src/pages/coslash/hooks/use-sessions.ts` — passed
- `cd frontend && npx vite build` — passed
- Manual: This Week returned 6 matching sessions in 291 ms versus 1.72 s for All, with zero record mismatches.

### DIFF
--- a/collector/cmd/coslash/main.go
+++ b/collector/cmd/coslash/main.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"log"
 	"net/http"
+	"strconv"
 	"time"
 	"unicode/utf8"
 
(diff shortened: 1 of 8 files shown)


luhe19001 (changes requested):
minor issues:

the "matches outside this window" empty state is now unreachable.

CoslashPage.tsx:209 counts sessions with status == null && mtime < windowStart,
but collector.go:54 now drops exactly those server-side (mtime is
LastActivityTime, session.go:51). The count is structurally 0, so
page-copy.ts:17 never renders.

Effect: searching for last month's work with "This week" selected shows "Nothing
matches X in this window" — with no hint that matches exist elsewhere.

Fix by returning an out-of-window count from the API, fetching unbounded while a
search term is active, or dropping the affordance deliberately.

luhe19001 (approved):

PR #10, opened 2026-08-04
Publish tagged release artifacts and a Homebrew formula

luhe19001:
## Context

coSlash has no published binaries, so the only way to get it is to clone the
repository and build it with Go and Node. This adds a tagged release path that
produces macOS binaries and a Homebrew formula, making install and upgrade a
single command. Binaries ship unsigned and unnotarized for the first release,
with Homebrew rather than direct download as the supported install path.

Depends on #8, which adds `--version` and the embedded frontend. This branch
builds and packages on its own, but the packaged binary is only complete once
#8 lands, so do not cut a tag before then. The two branches touch the same
Makefile and README sections with identical content, so the second merge will
need a trivial conflict resolution.

## Changes

- A `v*` tag builds macOS arm64 and amd64 binaries, publishes SHA-256
  checksums, and creates the GitHub release. Release notes lead with the
  Homebrew install command rather than the raw artifacts.
- A second job renders the formula for the tag, re-verifies the published
  checksums, and commits it to the tap repository. It runs only for stable
  semver tags, so prereleases publish artifacts without moving the tap.
- Before publishing, the workflow extracts the packaged binary and checks
  `--version`, the embedded UI, and the API, so a build that produces a broken
  binary cannot become a release.
- `make dist` cross-compiles both architectures from a single frontend build
  and writes `checksums.txt`. `make release` still builds one local binary.
- The formula resolves architecture with `Hardware::CPU` rather than
  `on_arm`/`on_intel` blocks, which `brew audit` rejects for `url` and `sha256`.
- `workflow_dispatch` builds and smoke-tests without publishing, so the
  pipeline can be exercised without spending a tag.

Publishing requires the tap repository to exist and a token secret with write
access to it; without both, the formula job fails after the release succeeds.

## Test

Validated against #8 merged in locally, since the packaged binary needs it.

- `make -C collector dist VERSION=v1.2.3-test` — passed; two archives plus
  `checksums.txt`, verified with `shasum -a 256 -c`
- Manual: extracted the packaged arm64 binary and confirmed `--version`, the
  embedded UI at `/`, and `/api/sessions`
- Manual: rendered the formula into a local tap — `brew style` and
  `brew audit` clean, then install, `brew test`, upgrade to a newer version,
  and uninstall all succeeded
- Manual: confirmed the installed binary carries no `com.apple.quarantine`
  attribute and launches with no Gatekeeper override

Not covered: Intel install (no Intel hardware available), the HTTPS download
path (local testing used `file://`), and the tap-publish job.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -149,8 +149,6 @@ jobs:
             fi
           fi
 
-          # Homebrew is the supported install path, so the notes lead with it
-          # rather than with the raw artifacts.
           notes=$(mktemp)
           cat >"$notes" <<'NOTES'
           ## Install
@@ -162,10 +160,23 @@ jobs:
 
           Already tapped? `brew upgrade coslash`.
 
-          The `.tar.gz` assets below are for manual installs. Verify them with
-          `shasum -a 256 -c checksums.txt`. The binaries are unsigned and
-          unnotarized, so macOS may warn about an archive downloaded through a
-          browser; the Homebrew install above is not affected.
+          The `.tar.gz` assets below are for manual installs:
+
+          ```sh
+          VERSION=v0.0.1
+          ARCH=arm64  # or amd64 on Intel
+          ASSET="coslash_${VERSION}_darwin_${ARCH}.tar.gz"
+          BASE_URL="https://github.com/centauri-ai/coslash/releases/download/${VERSION}"
+          curl -fLO "${BASE_URL}/${ASSET}"
+          curl -fLO "${BASE_URL}/checksums.txt"
+          grep -F "  ${ASSET}" checksums.txt | shasum -a 256 -c -
+          tar -xzf "${ASSET}"
+          "${ASSET%.tar.gz}/coslash"
+          ```
+
+          Substitute the release version and architecture. The binaries are
+          unsigned and unnotarized, so macOS may warn about an archive downloaded
+          through a browser; the Homebrew install above is not affected.
           NOTES
 
           if ! gh release view "$TAG" >/dev/null 2>&1; then
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -254,7 +254,7 @@ jobs:
           if [[ ! -f "$formula" ]]; then
             exit 0
           fi
-          current_version="$(ruby -ne 'if $_ =~ /^\s*version "([^"]+)"/; puts $1; exit; end' "$formula")"
+          current_version="$(ruby -ne 'if $_ =~ %r{/releases/download/v([^/]+)/}; puts $1; exit; end' "$formula")"
           if [[ -z "$current_version" ]]; then
             echo "error: could not read the current tap version" >&2
             exit 1
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,19 +13,15 @@ permissions:
   contents: read
 
 jobs:
-  release:
+  build:
     runs-on: macos-15
-    permissions:
-      contents: write
     steps:
-      - uses: actions/checkout@v5
-        with:
-          fetch-depth: 0
-      - uses: actions/setup-go@v6
+      - uses: actions/checkout@fbc6f3992d24b796d5a048ff273f7fcc4a7b6c09 # v5
+      - uses: actions/setup-go@924ae3a1cded613372ab5595356fb5720e22ba16 # v6
         with:
           go-version-file: collector/go.mod
           cache: false
-      - uses: actions/setup-node@v6
+      - uses: actions/setup-node@249970729cb0ef3589644e2896645e5dc5ba9c38 # v6
         with:
           node-version-file: frontend/.nvmrc
           cache: npm
@@ -44,7 +40,7 @@ jobs:
         env:
           VERSION: ${{ inputs.version || github.ref_name }}
         run: make -C collector dist VERSION="$VERSION"
-      - name: Smoke-test packaged binaries
+      - name: Smoke-test Apple silicon package
         env:
           VERSION: ${{ inputs.version || github.ref_name }}
         run: |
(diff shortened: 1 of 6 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,181 @@
...
+        required: true
+
+permissions:
+  contents: read
+
+jobs:
+  release:
+    runs-on: macos-15
+    permissions:
+      contents: write
```
milanshen: **[P2] Separate building from publishing permissions**

This job holds contents: write while checking out code, installing dependencies, testing, and building. Split the read-only build and smoke work from the publishing job so release credentials are available only to the minimal publication steps.

.github/workflows/release.yml
```diff
@@ -0,0 +1,181 @@
...
+    steps:
+      - uses: actions/checkout@v5
```
milanshen: **[P2] Pin privileged actions to SHAs**

These actions execute inside a release workflow with repository credentials, and checkout is also used later with the tap credential. Pin every action to a verified full commit SHA, retain a version comment, and let Dependabot maintain the pins.

.github/workflows/release.yml
```diff
@@ -0,0 +1,181 @@
...
+        with:
+          fetch-depth: 0
```
milanshen: **[P3] Use the default shallow checkout**

Nothing in this workflow uses Git history: VERSION is supplied explicitly and the build does not call git describe. Remove fetch-depth: 0 to keep the checkout simpler and faster.

milanshen (commented):
Before the first public tag:

- Scope HOMEBREW_TAP_TOKEN to the homebrew-tap repository only.
- Enable immutable GitHub releases.
- Protect creation, modification, and deletion of v* tags.

.github/workflows/release.yml
```diff
@@ -0,0 +1,181 @@
...
+            http://127.0.0.1:8787/ -o "$smoke_dir/index.html"
+          grep -q '<div id="root"' "$smoke_dir/index.html"
+          curl -sf http://127.0.0.1:8787/api/sessions | python3 -m json.tool >/dev/null
+          kill "$server_pid"
+          wait "$server_pid" 2>/dev/null || true
+          server_pid=
+
+          intel_stem="coslash_${VERSION}_darwin_amd64"
+          tar -xzf "collector/dist/${intel_stem}.tar.gz" -C "$smoke_dir"
+          file "$smoke_dir/$intel_stem/coslash" | grep -q 'Mach-O 64-bit executable x86_64'
```
milanshen: **[P2] Execute the Intel artifact natively**

The release advertises Intel support but only inspects the binary header. Run the packaged amd64 binary, including --version and the server smoke test, on a macos-15-intel job—or defer publishing the Intel artifact.

.github/workflows/release.yml
```diff
@@ -0,0 +1,181 @@
...
+          (cd collector/dist && shasum -a 256 -c checksums.txt)
+      - name: Publish
+        if: startsWith(github.ref, 'refs/tags/')
```
milanshen: **[P2] Keep manual dispatch non-publishing**

The workflow promises that workflow_dispatch only builds and smoke-tests, but a dispatch targeting a tag satisfies this condition and publishes. Require github.event_name == 'push' here and in bump-homebrew.

.github/workflows/release.yml
```diff
@@ -0,0 +1,181 @@
...
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          TAG: ${{ github.ref_name }}
+        run: |
+          set -euo pipefail
+          assets=(collector/dist/*.tar.gz collector/dist/checksums.txt)
+          if gh release view "$TAG" >/dev/null 2>&1; then
+            gh release upload "$TAG" "${assets[@]}" --clobber
```
milanshen: **[P1] Preserve published release assets**

A full rerun deletes and replaces existing assets, and identical clean builds currently produce different archive hashes. Create the release as a draft, upload and verify every asset, then publish it. A rerun may resume a draft, but it should fail if the release is already public rather than using --clobber.

.github/workflows/release.yml
```diff
@@ -0,0 +1,181 @@
...
+        with:
+          repository: centauri-ai/homebrew-tap
+          token: ${{ secrets.HOMEBREW_TAP_TOKEN }}
+          path: tap
+      - name: Download and verify release assets
+        env:
+          TAG: ${{ github.ref_name }}
+        run: |
+          set -euo pipefail
+          if [[ ! "$TAG" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
```
milanshen: **[P3] Remove duplicate version validation**

The release job already validates the tag, and this job only runs after that job succeeds for tags without a hyphen. This second semver check is redundant and can be removed.

.github/workflows/release.yml
```diff
@@ -0,0 +1,181 @@
...
+          {
+            echo "BREW_VERSION=${TAG#v}"
+            echo "RELEASE_TAG=$TAG"
+            echo "ARM_SHA=$(shasum -a 256 "downloads/$arm_asset" | awk '{print $1}')"
+            echo "INTEL_SHA=$(shasum -a 256 "downloads/$intel_asset" | awk '{print $1}')"
+          } >> "$GITHUB_ENV"
+      - name: Render formula
+        run: |
+          mkdir -p tap/Formula
+          ruby -rerb -e 'File.write(ARGV[1], ERB.new(File.read(ARGV[0]), trim_mode: "-").result)' coslash/.github/homebrew/coslash.rb.erb tap/Formula/coslash.rb
```
milanshen, line 269: **[P2] Validate the formula before pushing**

The generated formula is committed directly to the public tap without automated validation. Run brew style and brew audit after rendering and before committing; ideally the tap should also install the formula and run brew test in CI.

.github/workflows/release.yml
```diff
@@ -0,0 +1,181 @@
...
+          set -euo pipefail
+          git add Formula/coslash.rb
+          if git diff --cached --quiet; then
+            echo "Formula already matches $TAG"
+            exit 0
+          fi
+          git config user.name github-actions[bot]
+          git config user.email 41898282+github-actions[bot]@users.noreply.github.com
+          git commit -m "coslash $TAG"
+          git push
```
milanshen: **[P2] Reject stale Homebrew bumps**

A delayed older tag can commit its formula after a newer release and downgrade the tap. Serialize tap updates and compare the existing formula version before committing, refusing to replace it with an older version.

.github/workflows/release.yml
luhe19001: Split into read-only build and native smoke jobs; only the artifact-only publish job now has `contents: write`.

.github/workflows/release.yml
luhe19001: Pinned every action to a full commit SHA with version comments and added weekly GitHub Actions Dependabot updates.

.github/workflows/release.yml
luhe19001: Removed `fetch-depth`; checkout now uses the default shallow clone.

.github/workflows/release.yml
luhe19001: Added a `macos-15-intel` job that runs `--version` and the server smoke test against the packaged amd64 binary.

.github/workflows/release.yml
luhe19001: Both publication jobs now require `github.event_name == 'push'`; manual dispatch only builds and smoke-tests.

.github/workflows/release.yml
luhe19001: Publication now creates or resumes a draft, verifies every remote asset and checksum, refuses public releases, and only then publishes—without `--clobber`. Archives are reproducible across clean builds too.

.github/workflows/release.yml
luhe19001: Removed the duplicate semver check; this job only follows the validated stable-tag publication path.

.github/workflows/release.yml
luhe19001: The rendered formula now runs formula-specific `brew style` and `brew audit --strict` checks before commit.

.github/workflows/release.yml
luhe19001: Tap updates now share one concurrency group and compare the existing formula with `Gem::Version`, rejecting downgrades before rendering.

luhe19001:
Also enabled immutable releases and an active `v*` tag ruleset restricting creation, updates, and deletion to repository admins. The workflow now exposes `HOMEBREW_TAP_TOKEN` only to the final push step; GitHub does not expose the PAT’s repository selection, so please confirm it selects only `centauri-ai/homebrew-tap` before the first tag.

README.md
```diff
@@ -2,6 +2,33 @@
...
+
+```sh
+VERSION=v0.0.1
+ARCH=arm64  # or amd64 on Intel
+curl -LO "https://github.com/centauri-ai/coslash/releases/download/${VERSION}/coslash_${VERSION}_darwin_${ARCH}.tar.gz"
+tar -xzf "coslash_${VERSION}_darwin_${ARCH}.tar.gz"
+"coslash_${VERSION}_darwin_${ARCH}/coslash"
+```
+
+Every release publishes a `checksums.txt`; verify with
```
milanshen: **[P2] Make checksum verification match the one-archive flow**

The instructions download and execute one architecture archive without downloading `checksums.txt`. Even if it is downloaded, that file contains both architectures, so `shasum -a 256 -c checksums.txt` fails when the other archive is absent. Download the checksum file, filter it to the selected archive, and verify before executing; please mirror the corrected command in the generated release notes.

luhe19001: README and generated release notes now download `checksums.txt`, filter it to the selected archive, and verify before extraction. Locally reproduced the one-archive flow: checksum passed and the packaged binary reported `v1.2.3-test`.

milanshen (approved):

PR #17, opened 2026-08-04
ci: add pull-request CI for the release codepath

luhe19001:
## Context

The repository had no CI. Nothing checked a pull request, and nothing verified that the release build — the one that embeds the built frontend into the Go binary via `go:embed` — still produced a working executable. A broken release was only discoverable by hand.

This adds both, with the release path exercised the same way a person would run it rather than approximated by a separate set of steps.

## Changes

**`.github/workflows/ci.yml` — three jobs**, on every pull request, on pushes to `main`, and on manual dispatch. In-progress runs are cancelled for pull requests but not for `main`.

| Job | Runner | What it does |
| --- | --- | --- |
| `Frontend` | ubuntu | `npm ci`, lint, format check, tests, production build |
| `Collector` | ubuntu | `make check` (gofmt + `go vet`), `make test` |
| `Release` | macOS | `make release`, worktree-clean check, smoke test, dev build, dev smoke test, artifact upload |

The `Release` job is the point of the change. It runs the same `make release` a human release uses, then boots the resulting binary and asserts its behavior over HTTP. Step order matters: the embedded binary is copied aside *before* `make build` overwrites `bin/coslash`, so the uploaded artifact is the embedded one and not the bare development build.

**`collector/scripts/smoke.sh`** boots a built binary under an empty `HOME` and checks its HTTP contract in two modes.

- *Embedded* — `/` serves the application document; a `/assets/*.js` file referenced by that document resolves; `/coslash` survives a direct hit rather than 404ing; a missing asset returns **404 rather than falling back to the SPA document**; an unrouted `/api` path returns 404 rather than the application document; `/api/sessions` returns JSON; `--version` is not `dev`.
- *Bare* — `/` returns 503 and `/api/sessions` still answers.

The 404 and 503 assertions lock in the behavior from `0cb5405`, which was previously only verifiable by hand.

**`make check`, `make smoke`, `make smoke-dev`** give CI stable entry points instead of shell inlined into YAML, and stay usable locally.

**The collector now builds on non-darwin platforms.** The new Linux job caught this immediately: `Birthtimespec` exists only on darwin's `syscall.Stat_t`, so `internal/vendors/claude/fork.go` had never compiled anywhere else — nothing had ever tried. `fileCreationTime` already fell back to `ModTime()` when the `Stat_t` type assertion failed, so the intent was present; it simply could not compile where the field is absent. The darwin-only read now sits behind a build-tagged `birthtime` helper. No behavior change on darwin — the expression is unchanged — and fork ordering only uses birthtime to break ties between transcripts.

**`CoslashPage.tsx` reformatted** so the new formatting gate is not red on its first run. Whitespace and Tailwind class ordering only; nothing renders differently.

Plus a CI badge in the README.

## Test

All three jobs pass: `Frontend` 23s, `Collector` 23s, `Release` 48s.

**The uploaded artifact is the embedded binary** — the one failure mode that could ship behind a green check. Downloaded `coslash-macos` (10.1 MB), `--version` printed a real commit SHA rather than `dev`, and booting it under an empty `HOME` gave `/` → 200 with `<title>coSlash` and one `id="root"` div, `/api/sessions` → 200. A 503 there would have meant the wrong binary was uploaded.

**The frontend build step earns its slot.** With a deliberate type error present, `npm run lint`, `npm run format:check`, `vitest run`, and `vite build` all exit 0; only `npm run build` fails, with `TS2322`. `vite build` passing on its own confirms why — esbuild strips types without checking them, so `tsc -b` is doing the work. It uses the same script the release path invokes, so both check the identical build.

**Every CI step also replayed locally**, in the workflow's order, on darwin-arm64 / Node 23.5.0 / Go 1.26.5: `make check`, `make test`, `make release` (worktree clean afterwards), `make smoke`, `make build`, `make smoke-dev`, and the full frontend sequence from a cold `node_modules`. `GOOS=linux go build ./...` and `go vet ./...` pass on amd64 and arm64 after the portability fix.

Not yet done: deliberately breaking each gate to watch it go red — though the gofmt/vet gate proved itself for real via the Linux bug above, and the formatting gate was genuinely red before the `CoslashPage.tsx` fix — and the smoke script's port-fallback and argument-validation edge cases.

## Note for reviewers

`go test ./...` currently has no test files, so that step passes vacuously. The wiring is correct and picks up tests as soon as they exist, and `smoke.sh` is doing the real behavioral verification in the meantime, but a green `Collector` check should not be read as coverage. `internal/web`'s routing rules are pure functions of an `fs.FS` and are the obvious first candidate.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,115 @@
+name: CI
+
+on:
+  pull_request:
+  push:
+    branches: [main]
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  frontend:
+    name: Frontend
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    defaults:
+      run:
+        working-directory: frontend
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v7
+      - name: Set up Node.js
+        uses: actions/setup-node@v7
+        with:
+          node-version-file: frontend/.nvmrc
+          cache: npm
+          cache-dependency-path: frontend/package-lock.json
+      - name: Install frontend dependencies
+        run: npm ci
+      - name: Lint frontend
+        run: npm run lint
+      - name: Check frontend formatting
+        run: npm run format:check
+      - name: Test frontend
+        run: npm test
+
+  collector:
+    name: Collector
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    defaults:
+      run:
+        working-directory: collector
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v7
+      - name: Set up Go
+        uses: actions/setup-go@v7
+        with:
+          go-version-file: collector/go.mod
+          # This module has no dependencies or go.sum to cache.
+          cache: false
+      - name: Check collector
+        run: make check
+      - name: Test collector
+        run: make test
+
+  release:
+    name: Release
+    runs-on: macos-latest
+    timeout-minutes: 20
+    defaults:
+      run:
+        working-directory: collector
+    steps:
+      - name: Check out repository with tags
+        uses: actions/checkout@v7
+        with:
+          fetch-depth: 0
+      - name: Set up Node.js
+        uses: actions/setup-node@v7
+        with:
+          node-version-file: frontend/.nvmrc
+          cache: npm
+          cache-dependency-path: frontend/package-lock.json
+      - name: Set up Go
+        uses: actions/setup-go@v7
+        with:
+          go-version-file: collector/go.mod
+          # This module has no dependencies or go.sum to cache.
+          cache: false
+      - name: Build release binary
+        run: make release
+      - name: Verify release leaves worktree clean
+        run: |
+          status=$(git status --porcelain)
+          if [[ -n "$status" ]]; then
+            echo "Release build dirtied the worktree:"
+            echo "$status"
+            exit 1
+          fi
+      - name: Smoke test embedded frontend
+        run: make smoke
+      - name: Preserve release binary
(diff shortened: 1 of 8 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,117 @@
...
+  push:
+    branches: [main]
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```
milanshen, line 14: **[P2] Main CI runs can be dropped**

All pushes to `main` share one concurrency group. GitHub retains at most one running and one pending run, so a third push cancels the pending run even when `cancel-in-progress` is false. Group PR runs by PR number and use `github.run_id` for non-PR events.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,117 @@
...
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    defaults:
+      run:
+        working-directory: frontend
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v7
+      - name: Set up Node.js
+        uses: actions/setup-node@v7
```
milanshen: **[P2] Pin actions to immutable SHAs**

Replace every mutable `@v7` action reference with its full commit SHA, retaining a version comment or Dependabot configuration for updates. Apply this to checkout, setup-node, setup-go, and upload-artifact.

milanshen (commented):
Two recommended workflow changes before merge.

.github/workflows/ci.yml
luhe19001: Fixed in 7f43e5e: PR runs are grouped by PR number, while non-PR runs use github.run_id, so pushes to main no longer share a concurrency group.

.github/workflows/ci.yml
luhe19001: Fixed in 7f43e5e: checkout, setup-node, setup-go, and upload-artifact are pinned to their current v7 commit SHAs, with v7 comments retained for update visibility.

milanshen (approved):

PR #18, opened 2026-08-04
Add global synthesis and terminal settings

milanshen:
## Context

coSlash currently hard-codes its synthesis backend, model, and macOS terminal. This adds secure machine-wide settings with explicit synthesis consent, configurable runtime choices, and fail-closed handling for invalid configuration.

## Changes

- Add a versioned `~/.coslash/settings.json` schema and restricted atomic persistence.
- Add first-run synthesis consent and a full settings UI with availability and error states.
- Support Claude Code and Codex synthesis backends with selectable models.
- Support Apple Terminal and iTerm2 with availability checks and immediate runtime updates.

## Test

fresh start or delete `~/.coslash/settings.json`

### Screenshots

Screenshot 2026-08-04 at 3 20 29 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,3 +32,18 @@ cd frontend && npm run dev
 ```
 
 The UI is at [http://localhost:5173](http://localhost:5173): Vite serves it and proxies `/api` to the Go server on port `8787`. `make run` embeds no frontend and opens no browser, so port `8787` serves the API alone.
+
+## Global settings
+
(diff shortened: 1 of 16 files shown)


luhe19001 (changes requested):
## Requested change

First-run can save synthesis **On** with an unavailable backend still selected.

The consent draft defaults to `enabled: true` and `claude-cli`. Unavailable backends are not clickable, but the current selection is not cleared and Save is not blocked — so a Codex-only machine can accept the dialog as-is and persist a config that fails every synthesis run.

Please block that path, e.g. auto-select the first available backend, disable Save when synthesis is on and the selected backend is unavailable, or default draft `enabled` to false when no backend is available.

luhe19001 (approved):

PR #19, opened 2026-08-05
Harden loopback server access with per-run tokens

luhe19001:
## Context

Protect private local transcript data from unrelated browser tabs. 

## Changes

- Guard loopback Host, Origin, and fetch metadata; require a per-run API token and add security headers.
- Deliver the token through the URL fragment/session storage and inject it through the Vite proxy.
- Resolve the token per request so a new restart URL recovers an existing tab.
- Add a frontend regression test for token refresh after restart.

## Test

- `go test ./...` — packages compile; no Go test files
- `npm test` — passed
- `npm run lint` — passed with two existing warnings
- `npm run build` — passed
- Manual: verified token mode/rotation, 401/200/403 API guards, headers, and Vite proxy on fallback port 5174.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,11 +42,11 @@ Running `coslash` starts one executable at
 [http://127.0.0.1:8787](http://127.0.0.1:8787) and opens a browser. There is no
 Node, Vite, or second process at runtime.
 
-| Flag | Effect |
-| --- | --- |
-| `--port N` | serve on port N (default `8787`) |
-| `--no-open` | do not open the browser |
-| `--version` | print the version and exit |
+| Flag        | Effect                           |
+| ----------- | -------------------------------- |
+| `--port N`  | serve on port N (default `8787`) |
+| `--no-open` | do not open the browser          |
+| `--version` | print the version and exit       |
 
 ## Build from source
 
@@ -63,6 +63,12 @@ That builds the frontend, embeds it into the binary, and leaves it at
 architectures into `collector/dist/` with a `checksums.txt`, which is what the
 release workflow publishes.
 
+On every start, coSlash prints and opens a URL containing a new access token.
+Use that URL rather than typing the address by hand. The token is delivered in
+the URL fragment, kept in browser session storage, and, when local storage is
+writable, saved to `~/.coslash/token` with owner-only permissions. An old link
+expires when the server restarts.
+
 ## Develop
 
 In two terminals:
@@ -72,7 +78,23 @@ cd collector && make run
 cd frontend && npm run dev
 ```
 
-The UI is at [http://localhost:5173](http://localhost:5173): Vite serves it and proxies `/api` to the Go server on port `8787`. `make run` embeds no frontend and opens no browser, so port `8787` serves the API alone.
+The UI is at [http://127.0.0.1:5173](http://127.0.0.1:5173): Vite serves it and
+proxies `/api` to the Go server on port `8787`, reading the current development
+token automatically. If port `5173` is occupied, use the fallback URL Vite
+prints. `make run` embeds no frontend and opens no browser, so port `8787`
+serves the API alone.
+
+The production server binds only to IPv4 loopback and rejects unexpected Host,
+Origin, and cross-site browser requests. Direct API calls must include the
+current token. This form keeps the token out of `curl`'s process arguments:
+
+```sh
+sed 's/^/header = "X-Coslash-Token: /; s/$/"/' ~/.coslash/token \
+  | curl --config - http://127.0.0.1:8787/api/sessions
+```
+
+coSlash does not enable CORS. New HTTP routes must remain behind the same
+loopback request guard.
 
 ## Global settings
 
@@ -91,7 +113,7 @@ Use the top-right Settings dialog to apply synthesis and terminal changes immedi
 
 After installing coSlash, run `coslash doctor` to check detected session sources, CLIs, and local storage.
 
-| Command | Effect |
-| --- | --- |
-| `coslash doctor` | Print checks and local diagnostic facts; exits non-zero when a check fails. |
-| `coslash doctor --json` | Print the same diagnostic snapshot as JSON. |
+| Command                 | Effect                                                                      |
+| ----------------------- | --------------------------------------------------------------------------- |
+| `coslash doctor`        | Print checks and local diagnostic facts; exits non-zero when a check fails. |
+| `coslash doctor --json` | Print the same diagnostic snapshot as JSON.                                 |
--- a/collector/CLAUDE.md
+++ b/collector/CLAUDE.md
@@ -1,0 +1,11 @@
+# Collector HTTP security
+
+All routes, including the embedded frontend, must be served through
+`internal/httpsec.Guard`. New routes must not bypass the guard or emit CORS
+headers. API routes require the per-run token and must use method-qualified
+`http.ServeMux` patterns. Frontend requests to `/api/` must use `apiFetch` so
+the per-run token is attached. Keep internal errors in server logs and return
+fixed, non-sensitive messages to clients.
+
+The token prevents requests from other browser origins; it does not protect
+against local processes running as the same user, which can read the token file.
--- a/collector/cmd/coslash/main.go
+++ b/collector/cmd/coslash/main.go
@@ -3,6 +3,8 @@ package main
 
 import (
 	"context"
+	"crypto/rand"
+	"encoding/base64"
 	"errors"
 	"flag"
 	"fmt"
@@ -10,11 +12,13 @@ import (
 	"net"
 	"net/http"
 	"os"
+	"path/filepath"
 	"syscall"
 	"time"
 
 	"github.com/centauri-ai/coslash/collector/internal/collector"
 	"github.com/centauri-ai/coslash/collector/internal/diagnostics"
+	"github.com/centauri-ai/coslash/collector/internal/httpsec"
 	"github.com/centauri-ai/coslash/collector/internal/session"
 	"github.com/centauri-ai/coslash/collector/internal/settings"
 	"github.com/centauri-ai/coslash/collector/internal/synthesis"
@@ -94,22 +98,44 @@ func main() {
 	if err != nil {
 		log.Fatalf("coslash: %v", err)
 	}
-	url := "http://" + listener.Addr().String()
-	log.Printf("listening on %s", url)
+	token, err := newToken()
+	if err != nil {
+		log.Fatalf("coslash: generate API token: %v", err)
+	}
+	if err := writeToken(token); err != nil {
+		log.Printf("write API token: %v", err)
+	}
+	baseURL := "http://" + listener.Addr().String()
+	accessURL := baseURL + "/#t=" + token
+	log.Printf("listening on %s", baseURL)
+	log.Printf("open %s", accessURL)
 	if !opts.noOpen {
-		if err := openBrowser(url); err != nil {
-			log.Printf("could not open a browser (%v); open %s yourself", err, url)
+		if err := openBrowser(accessURL); err != nil {
+			log.Printf("could not open a browser (%v); use the URL above", err)
 		}
 	}
-	if err := http.Serve(listener, routes(mgr, settingsStore)); err != nil {
+	guard := httpsec.Guard{Addr: listener.Addr().String(), Token: token}
+	server := newServer(guard, mgr, settingsStore)
+	if err := server.Serve(listener); err != nil {
 		log.Fatalf("coslash: %v", err)
+	}
+}
+
+func newServer(guard httpsec.Guard, mgr *synthesis.Manager, settingsStore *settings.Store) *http.Server {
+	return &http.Server{
+		Handler:           guard.Wrap(routes(mgr, settingsStore)),
+		ReadHeaderTimeout: 10 * time.Second,
+		ReadTimeout:       30 * time.Second,
+		WriteTimeout:      3 * time.Minute,
+		IdleTimeout:       60 * time.Second,
+		MaxHeaderBytes:    1 << 16,
 	}
 }
 
 func routes(mgr *synthesis.Manager, settingsStore *settings.Store) *http.ServeMux {
 	mux := http.NewServeMux()
 	api := http.NewServeMux()
-	api.HandleFunc("/api/sessions", func(w http.ResponseWriter, r *http.Request) {
+	api.HandleFunc("GET /api/sessions", func(w http.ResponseWriter, r *http.Request) {
 		handleList(w, r, mgr)
 	})
 	api.HandleFunc("GET /api/synthesis", func(w http.ResponseWriter, r *http.Request) {
@@ -127,30 +153,18 @@ func routes(mgr *synthesis.Manager, settingsStore *settings.Store) *http.ServeMu
 	api.HandleFunc("GET /api/diagnostics", func(w http.ResponseWriter, r *http.Request) {
 		writeJSON(w, diagnostics.Collect(r.Context(), version, false))
 	})
-	apiHandler := sameOriginAPI(api)
-	mux.Handle("/api", apiHandler)
-	mux.Handle("/api/", apiHandler)
+	mux.Handle("/api", api)
+	mux.Handle("/api/", api)
 
 	frontend, err := web.Handler()
 	if err != nil {
 		// A `make build` binary has no staged assets; keep its API usable for
 		// `npm run dev` instead of failing to start.
-		log.Printf("coslash: %v", err)
-		frontend = unavailable(err)
+		log.Printf("coslash: frontend unavailable: %v", err)
+		frontend = unavailable()
 	}
 	mux.Handle("/", frontend)
 	return mux
-}
-
-func sameOriginAPI(next http.Handler) http.Handler {
-	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		site := r.Header.Get("Sec-Fetch-Site")
-		if site != "" && site != "same-origin" && site != "none" {
-			http.Error(w, "cross-origin API requests are not allowed", http.StatusForbidden)
-			return
-		}
-		next.ServeHTTP(w, r)
-	})
 }
 
 func listen(port int) (net.Listener, error) {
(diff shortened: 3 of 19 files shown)


collector/cmd/coslash/main.go
```diff
@@ -79,23 +83,49 @@ func main() {
...
 		log.Fatalf("coslash: %v", err)
 	}
-	url := "http://" + listener.Addr().String()
-	log.Printf("listening on %s", url)
+	token, err := newToken()
+	if err != nil {
+		log.Fatalf("coslash: generate API token: %v", err)
+	}
+	if err := writeToken(token); err != nil {
+		log.Fatalf("coslash: write API token: %v", err)
```
calvintvu: **Blocking.** A failed token write stops the server.

The browser gets the token from the URL fragment. Only the Vite proxy and `curl` read the file. A read-only or odd-permission `~/.coslash` now prevents startup.

`main` already accepts a failure from `synthesis.EnsureDirs` and continues. Use `log.Printf` here and continue.

collector/internal/httpsec/httpsec.go
```diff
@@ -0,0 +1,91 @@
...
+				http.Error(w, "request rejected", http.StatusForbidden)
+				return
+			}
+		} else if origin := r.Header.Get("Origin"); origin != "" && origin != "http://"+g.Addr {
+			log.Printf("http security: rejected unexpected origin %q", origin)
+			http.Error(w, "request rejected", http.StatusForbidden)
+			return
+		}
+
+		if strings.HasPrefix(r.URL.Path, "/api/") && g.Token != "" && !g.allowedToken(r) {
```
calvintvu: **Blocking.** `g.Token != ""` disables API authentication for a zero-valued `Guard`. No caller uses the empty token. Line 17 documents the empty token as a feature.

Delete the condition. If a no-auth mode becomes necessary, add an explicit field.

collector/internal/httpsec/httpsec.go
```diff
@@ -0,0 +1,91 @@
...
+
+// Guard restricts requests to the loopback listener and its browser origin.
+// Token may be left empty to disable API authentication.
+type Guard struct {
+	Addr  string
+	Token string
+}
+
+// Wrap applies loopback request validation and security response headers.
+func (g Guard) Wrap(next http.Handler) http.Handler {
```
calvintvu, line 23: **Optional.** This guard is the security boundary of the whole change. It has no Go test.

Add one table test with `httptest`:

- a bad Host returns 403
- `Sec-Fetch-Site: cross-site` returns 403
- `Sec-Fetch-Site: same-site` returns 403
- a missing token returns 401
- a wrong token returns 401
- the correct token returns 200
- a non-API path without a token returns 200
- Host `localhost:8787` returns 200

The PR text says "no Go test files". `internal/session` has a test now.

calvintvu (commented):
## Summary

The design is correct. Loopback binding is not an authorization boundary. The four controls match the real threats:

- The Host allowlist stops DNS rebinding.
- The guard rejects `Sec-Fetch-Site: same-site`, not only `cross-site`. That closes the local case, a page on another 127.0.0.1 port.
- The token gates `/api/`.
- The URL fragment keeps the token out of server logs and the Referer header.

Per-request token resolution is the best part. A hash-only navigation does not reload the document, so a live tab recovers on its next poll. The error sanitization in `api.go` also removes internal detail from client responses.

## What I verified

I ran the mux in isolation and read the frontend of the branch.

- Method handling is inconsistent. `POST /api/sessions` returns 405. `POST /api/synthesis` and `GET /api/launch` return 404, because the `/api/` catch-all matches every method.
- `hasTraversal` is reachable. `/assets/%2e%2e/%2e%2e/etc/passwd` arrives at the frontend handler, because the mux cleans the escaped path only.
- The CSP does not break the release build. The fonts come from `@fontsource-variable/geist` in the bundle, and `src/` holds no external URL. `style-src 'unsafe-inline'` is necessary for the inline style attributes of Radix.
- Every `/api/` caller of the branch uses `apiFetch`. No raw `fetch` remains.

## Blocking

1. A failed token write stops the server (`main.go:91`).
2. An empty `Token` disables API authentication (`httpsec.go:46`).

The other comments are minor.

collector/cmd/coslash/main.go
```diff
@@ -79,23 +83,49 @@ func main() {
...
+}
+
 func routes(mgr *synthesis.Manager) *http.ServeMux {
 	mux := http.NewServeMux()
-	mux.HandleFunc("/api/sessions", func(w http.ResponseWriter, r *http.Request) {
+	mux.HandleFunc("GET /api/sessions", func(w http.ResponseWriter, r *http.Request) {
 		handleList(w, r, mgr)
 	})
+	mux.HandleFunc("/api/sessions", func(w http.ResponseWriter, _ *http.Request) {
+		w.Header().Set("Allow", "GET, HEAD")
```
calvintvu: I ran this mux in isolation. `POST /api/sessions` returns 405. `POST /api/synthesis` and `GET /api/launch` return 404. The `/api/` catch-all matches every method and suppresses the automatic 405 of Go.

So this handler is not redundant, but it covers one route of three. Delete it, or add one for each route. `collector/CLAUDE.md` tells the next contributor to use method-qualified patterns, and 404 is the result they get.

collector/internal/web/web.go
```diff
@@ -36,6 +36,10 @@ func handler(assets fs.FS) (http.Handler, error) {
 	}
 	files := http.FileServerFS(assets)
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if hasTraversal(r.URL.Path) {
```
calvintvu, line 39: The guard is reachable. `/assets/%2e%2e/%2e%2e/etc/passwd` arrives here with `r.URL.Path` equal to `/assets/../../etc/passwd`. The mux cleans the escaped path only, so it sends no redirect.

Nothing was exploitable before this guard. `fs.ValidPath` rejects every path that contains `..`, so `isFile` returned false and the handler served `index.html`. The guard changes that result to 404.

Keep the guard, but add a test. Without a test, a later reader deletes the guard as dead code.

frontend/vite.config.ts
```diff
@@ -6,7 +8,34 @@ import { defineConfig } from 'vite';
 export default defineConfig({
   plugins: [react(), tailwindcss()],
   server: {
-    proxy: { '/api': 'http://127.0.0.1:8787' },
+    host: '127.0.0.1',
+    port: 5173,
+    strictPort: true,
```
calvintvu: `strictPort: true` removes the fallback port. The PR text reports a manual test on fallback port 5174. With `strictPort`, Vite stops instead.

If you want the fallback port, delete `strictPort`. If you want the strict port, correct the PR text.

frontend/vite.config.ts
```diff
@@ -6,7 +8,34 @@ import { defineConfig } from 'vite';
...
+    proxy: {
+      '/api': {
+        target: 'http://127.0.0.1:8787',
+        changeOrigin: true,
+        configure: (proxy) => {
+          let warnedAboutMissingToken = false;
+          proxy.on('proxyReq', (proxyRequest) => {
+            if (proxyRequest.getHeader('origin') === 'http://127.0.0.1:5173') {
```
calvintvu: This comparison hardcodes port 5173. A developer who opens `http://localhost:5173` sends a different Origin. Then this branch skips the rewrite.

The rewrite works only because the guard reads Origin when `Sec-Fetch-Site` is absent. A browser always sends `Sec-Fetch-Site`.

Replace the rewrite with `proxyRequest.removeHeader('origin')`. The result is one line and no port in the code. `changeOrigin: true` still corrects the Host header, which `allowedHost` needs.

README.md
```diff
@@ -29,4 +37,19 @@ cd collector && make run
...
+proxies `/api` to the Go server on port `8787`, reading the current development
+token automatically. `make run` embeds no frontend and opens no browser, so port
+`8787` serves the API alone.
+
+The production server binds only to IPv4 loopback and rejects unexpected Host,
+Origin, and cross-site browser requests. Direct API calls must include the
+current token:
+
+```sh
+curl -H "X-Coslash-Token: $(cat ~/.coslash/token)" \
```
calvintvu: This command puts the token in the argv of `curl`. On a shared host, `ps` shows argv to every user. The shell history also keeps the token.

A variable does not remove this risk, because the shell expands the variable into argv. `curl -K file` reads the header from a file instead. A note about the risk is also enough.

collector/CLAUDE.md
```diff
@@ -0,0 +1,7 @@
+# Collector HTTP security
+
+All routes, including the embedded frontend, must be served through
+`internal/httpsec.Guard`. New routes must not bypass the guard or emit CORS
+headers. API routes require the per-run token and should use method-qualified
```
calvintvu: Agents and humans read `should` as optional. Write `must` for a rule.

Add one line about the boundary of the token. The token stops other browser origins. The token does not stop local processes, because any process of the user reads `~/.coslash/token`. This limit matches the stated threat model. Without the line, a later contributor assumes the wrong boundary.

collector/cmd/coslash/main.go
```diff
@@ -128,6 +158,44 @@ func listen(port int) (net.Listener, error) {
 
 func unavailable(err error) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
-		http.Error(w, err.Error(), http.StatusServiceUnavailable)
+		log.Printf("frontend unavailable: %v", err)
```
calvintvu: This handler logs one line for every request. A `make build` binary plus one browser request for `/favicon.ico` fills the log. Log the error once at startup.

collector/cmd/coslash/main.go
```diff
@@ -128,6 +158,44 @@ func listen(port int) (net.Listener, error) {
...
+		return err
+	}
+	path := filepath.Join(home, "token")
+	temporary, err := os.CreateTemp(home, ".token-*")
+	if err != nil {
+		return err
+	}
+	temporaryPath := temporary.Name()
+	defer os.Remove(temporaryPath)
+	if err := temporary.Chmod(0o600); err != nil {
```
calvintvu: `os.CreateTemp` creates the file with mode 0600 already. Delete this call.

frontend/src/pages/coslash/hooks/use-sessions.ts
```diff
@@ -59,9 +64,16 @@ export function useSessions(timeWindow: TimeWindow) {
...
+    setLoadError(null);
     setRetryCount((key) => key + 1);
   };
 
-  return { sessions, isLoading, loadFailed, sessionsVersion, retrySessions };
+  return {
+    sessions,
+    isLoading,
+    loadError,
+    loadFailed: loadError != null,
```
calvintvu: The hook returns two shapes of one state. `SessionsStats` accepts `loadError != null` at the call site instead. Then delete `loadFailed`.

frontend/src/pages/coslash/lib/api.test.ts
```diff
@@ -0,0 +1,36 @@
...
+    vi.stubGlobal('fetch', fetchMock);
+
+    await apiFetch('/api/sessions');
+    location.hash = '#t=new-token';
+    await apiFetch('/api/sessions');
+
+    const secondRequest = fetchMock.mock.calls[1]?.[1] as RequestInit;
+    expect(new Headers(secondRequest.headers).get('X-Coslash-Token')).toBe('new-token');
+    expect(stored.get('coslash-api-token')).toBe('new-token');
+    expect(replaceState).toHaveBeenCalledTimes(2);
```
calvintvu, line 34: This test covers the token refresh well. The test does not cover the 401 mapping to `ApiAuthenticationError`. That mapping is the security-visible branch. Add one case for status 401.

calvintvu:
also make sure to rebase

frontend/src/pages/coslash/lib/api.test.ts
luhe19001: Added a frontend regression test confirming that a 401 response maps to ApiAuthenticationError.

README.md
luhe19001: Updated the README example to pipe a generated curl config through stdin, keeping the token out of curl's process arguments and shell history.

collector/CLAUDE.md
luhe19001: Updated the guidance: API routes must use method-qualified patterns. It also now states that the token protects against browser origins, not same-user local processes.

frontend/vite.config.ts
luhe19001: Removed Vite strictPort to preserve fallback-port behavior. The README now directs developers to the fallback URL Vite prints.

collector/cmd/coslash/main.go
luhe19001: Fixed: the frontend initialization error is logged once when routes are created; the unavailable handler now returns only the fixed 503 response.

collector/cmd/coslash/main.go
luhe19001: Fixed in ecd88e8: token-file persistence is now best-effort. A write failure is logged and startup continues with the browser fragment token.

frontend/vite.config.ts
luhe19001: The proxy now removes the Origin header unconditionally, while changeOrigin supplies the backend Host. This avoids hardcoded frontend ports and hostnames.

collector/cmd/coslash/main.go
luhe19001: Removed the redundant chmod call; os.CreateTemp already creates the token file with mode 0600.

collector/internal/httpsec/httpsec.go
luhe19001: Fixed in ecd88e8: API authentication now fails closed. An empty configured token is explicitly rejected, with a regression test for the zero-value guard.

collector/cmd/coslash/main.go
luhe19001: Fixed: all API routes now use method-qualified patterns on the API sub-mux. Regression tests confirm unsupported methods consistently return 405.

collector/internal/httpsec/httpsec.go
luhe19001: Added table-driven httptest coverage for Host validation, cross-site and same-site requests, missing/wrong/correct tokens, non-API paths, and localhost.

frontend/src/pages/coslash/hooks/use-sessions.ts
luhe19001: useSessions now returns only loadError; callers derive the boolean with a null check where needed.

collector/internal/web/web.go
luhe19001: Kept the traversal guard and added a regression test for an assets path containing parent-directory traversal, plus missing assets returning 404.

frontend/src/pages/coslash/lib/api.ts
```diff
@@ -0,0 +1,34 @@
...
+      window.history.state,
+      '',
+      `${window.location.pathname}${window.location.search}`,
+    );
+    return suppliedToken;
+  }
+  return window.sessionStorage.getItem(TOKEN_STORAGE_KEY);
+}
+
+export async function apiFetch(path: string, init: RequestInit = {}): Promise<Response> {
```
calvintvu, line 27: **Blocking.** Three API callers do not use `apiFetch`. The rebase added them from main:

- `use-settings.ts:18` calls `GET /api/settings`
- `use-settings.ts:40` calls `PUT /api/settings`
- `use-diagnostics.ts:15` calls `GET /api/diagnostics`

The guard requires a token for every `/api/` path. So all three requests get 401.

I ran the release binary. Both endpoints return 401 without a token and 200 with a token. I loaded the app once through `open`. The server log then got one `rejected unauthenticated API request` line.

The result for the user: the Settings dialog does not load and does not save. `readError` returns the response body, so the dialog shows the guard text `authentication required`. The Diagnostics dialog also fails.

Use `apiFetch` in both hooks. Then add the frontend rule to `collector/CLAUDE.md`. Every `/api/` call must use `apiFetch`. Without that rule, the next new route repeats this error.

collector/internal/httpsec/httpsec.go
```diff
@@ -0,0 +1,93 @@
...
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		setSecurityHeaders(w.Header(), r.URL.Path)
+
+		if !g.allowedHost(r.Host) {
+			log.Printf("http security: rejected unexpected host %q", r.Host)
+			http.Error(w, "request rejected", http.StatusForbidden)
+			return
+		}
+
+		if site := r.Header.Get("Sec-Fetch-Site"); site != "" {
```
calvintvu, line 33: **Blocking.** This condition rejects a cross-site top-level navigation to the document. When the user clicks a link on another page, Chrome sends `Sec-Fetch-Site: cross-site`. The user then sees the plain text `request rejected`.

I did tests against the release binary:

- `open`, a typed address, and a bookmark send `none`. The document returns 200. The default startup flow works.
- A link click from a web page, a chat app, or the docs sends `cross-site`. The document returns 403.

The document navigation is not the threat. `/api/` still requires the token, so a cross-site page that navigates the user to the SPA gets no data.

Accept navigations:

```go
if site := r.Header.Get("Sec-Fetch-Site"); site != "" {
	navigation := strings.EqualFold(r.Header.Get("Sec-Fetch-Mode"), "navigate") &&
		strings.EqualFold(r.Header.Get("Sec-Fetch-Dest"), "document")
	if !navigation && !strings.EqualFold(site, "same-origin") && !strings.EqualFold(site, "none") {
		// reject
	}
}
```

A subresource stays rejected. I did a test with `cross-site`, `Sec-Fetch-Mode: cors`, and `Sec-Fetch-Dest: empty`. The guard rejects that request, which is correct.

If you keep the strict rejection, serve an HTML page that explains the cause.

calvintvu (commented):
## Re-review of `ecd88e8`

All 13 earlier comments are correct now. I did these tests:

- `go vet ./...` is clean. `go test ./...` passes.
- `vitest` passes 13 tests in 5 files. `oxlint` gives 2 pre-existing warnings. `npm run build` passes.
- I built the release binary and ran it. I sent 15 requests through the guard. Every result matches the documentation: 401 without a token, 200 with a header token or a Bearer token, 403 for a bad Host, a bad Origin, `cross-site`, and `same-site`, 405 for `POST /api/sessions` and `GET /api/launch`, 404 for `/api/nope` and for encoded traversal. The token file has mode 0600. No response carries a CORS header.
- The built `index.html` holds no inline script. So `script-src 'self'` does not break the bundle.
- The `curl --config -` form in the README works. The token stays out of the process arguments.

Two blockers remain. The rebase caused both of them. The original design is sound.

Minor items for a later pass:

- `httpsec_test.go` expects 200 for Host `[::1]:8787`. The server binds IPv4 only, so that Host cannot reach the listener.
- `httpsec_test.go` sets absent headers with `Header.Set(k, "")`. `Header.Del` states the intent.
- `writeToken` calls `os.Chmod(home, 0o700)` on every start. This call overrides the permissions of the user, and it follows a symlink.
- The history contains `chore: remove Go test files`. The next commit adds the same three files again. Squash before merge.

frontend/src/pages/coslash/lib/api.ts
calvintvu: https://github.com/centauri-ai/coslash/actions/runs/31050926489/job/92457654782?pr=19

luhe19001: Fixed in `35c74b8`. Settings GET/PUT and Diagnostics GET now use `apiFetch`, and `collector/CLAUDE.md` documents that every frontend `/api/` request must use it. The smoke script also authenticates API probes so it exercises the guarded routes correctly.

collector/internal/httpsec/httpsec.go
luhe19001: Fixed in `35c74b8`. The guard now accepts top-level `navigate`/`document` requests while continuing to reject cross-site subresources. The table test covers both cases.

calvintvu (approved):

PR #40, opened 2026-08-05
Read model prices and context windows from LiteLLM instead of hand-kept tables

calvintvu:
## Context

we dont want to manually sync model information, we'll follow `ccusage` pattern and pull info from litellm

## Changes

- The collector reads prices and context windows from one embedded table, generated from LiteLLM (MIT). A new model release needs no code change.
- The collector computes each cost and sends it with the session. The frontend holds no pricing table.
- `make models` rewrites the table. The Models workflow runs it daily and opens a pull request, so a person reads every price change before it merges.
- The workflow compares the model list before and after each refresh. The pull request names every model added and dropped, and a dropped model turns the run red.
- Three prices now match LiteLLM: gpt-5.6-terra input is $2.00 per MTok, not $2.50. gpt-5.6-luna input is $0.20, not $1.00. The gpt-5.5-pro cache read is $3.00, not $30.00.
- The context window of claude-sonnet-5 is 1,000,000, not 967,000. The context gauge shows a smaller percentage for those sessions.
- LiteLLM carries no entry for claude-mythos-5 or gpt-5.5-cyber. Sessions on those models show the unpriced-model warning and no cost.
- A Codex session takes its context window from the table when the transcript reports none.

## Test

- `make check` - passed. The collector carries no test files, so `make test` runs nothing.
- `npm run format:check`, `npx tsc -b`, `npm run lint`, `npm test` (8 tests), `npm run build` - all passed
- Manual: ran the dev binary and read `/api/sessions`. 191 sessions, total cost $1166.56, no unpriced model, subagent costs present, claude-sonnet-5 window 1,000,000.
- Manual: ran the workflow comparison locally. A real refresh reported no change. A table with `claude-opus-5` deleted reported the drop and exited 1.
- The last commit deletes `o4-mini-2025-04-16` from the table on purpose, to watch the workflow report a model as added. Revert it before merge.

### Screenshots

Screenshot 2026-08-10 at 13 23 03

Screenshot 2026-08-10 at 13 23 23


calvintvu:
6e190b8057aee433d46974e63512bc2e55e4d640 removes model o4-mini-2025-04-16 from models json, upon merge we can test workflow to see if we updated the models json correctly

luhe19001 (commented):
This backend migration regresses cost and context-window handling for model IDs that were supported by the previous pricing table but are absent from the generated LiteLLM subset. For sessions using previously supported IDs such as `claude-mythos-5`, `gpt-5.5-pro`, or `gpt-5.5-cyber`, the lookup now fails after only removing a date suffix, so usage becomes unpriced and contributes $0. `claude-mythos-5` also loses its context window. Please preserve local aliases or fallback entries for these IDs when moving pricing to the backend.

collector/internal/session/models.go
```diff
@@ -1,42 +1,114 @@
...
-func ContextWindowFor(model string) *int {
-	if model == "" {
-		return nil
+func modelInfoFor(model string) (modelInfo, bool) {
+	key := strings.TrimSuffix(model, "[1m]")
+	if info, ok := models[key]; ok {
+		return info, true
 	}
+	info, ok := models[releaseDateSuffix.ReplaceAllString(key, "")]
+	return info, ok
```
luhe19001, line 63: [P2] Retain pricing for model aliases absent from LiteLLM\n\nFor sessions using previously supported IDs that LiteLLM does not list, such as `claude-mythos-5`, `gpt-5.5-pro`, or `gpt-5.5-cyber`, this lookup now fails after only removing a date suffix. Those tokens become unpriced and contribute $0, whereas the removed frontend table explicitly priced them; `claude-mythos-5` also loses its context window. Preserve local aliases/fallback entries for these IDs when moving pricing to the backend.

calvintvu: lets deliberately drop our handwritten support for these 2 models: `claude-mythos-5`, `gpt-5.5-cyber`
we can pull `gpt-5.5-pro` from litellm 
likelihood of our anticipated user base using these models are low

collector/internal/session/models.go
```diff
@@ -1,42 +1,114 @@
...
+		}
+		if info.CacheRead == 0 {
+			info.CacheRead = info.Input
+		}
+		loaded[model] = info
+	}
+	return loaded
 }
 
 var releaseDateSuffix = regexp.MustCompile(`-\d{8}$`)
```
luhe19001: **[P2]** ISO-formatted dated model IDs cannot use the intended base-model fallback. When a transcript reports an ID such as `o4-mini-2025-04-16` that is not explicitly in the snapshot, this regex cannot fall back to the existing `o4-mini` entry because it only recognizes `-YYYYMMDD`. The session is therefore marked unpriced and its cost is reported as $0 even though a base-model price is available. Please support `-YYYY-MM-DD` suffixes as well.

calvintvu: Fixed in 91097c1. The regex now reads `-\d{4}-?\d{2}-?\d{2}$`, so it strips Anthropic's `-20251001` and OpenAI's `-2025-04-16`. Both `o4-mini-2026-07-01` and `gpt-5.4-2026-09-09` resolve to their base model now.

One correction to the example: `o4-mini-2025-04-16` is an exact key in the table already, so it was priced before this change. LiteLLM publishes the base and the dated variant together. The gap you found is real for a dated id that ships before LiteLLM keys it, which the daily refresh shortens but does not close.

luhe19001 (approved):

PR #41, opened 2026-08-05
dark mode

calvintvu:
## Context

shadcn/tailwind makes adding dark mode simple enough 

## Changes

- add dark mode toggle (light is default)
- ~~preference saved to browser local storage~~ preference saved to app settings file 

## Test

no tests

### Screenshots

- [x] Screen or component — state and visible result
Screenshot 2026-08-05 at 13 48 35
Screenshot 2026-08-05 at 13 48 42
Screenshot 2026-08-05 at 13 48 47
Screenshot 2026-08-05 at 13 53 38


frontend/src/pages/coslash/CoslashPage.tsx
```diff
@@ -57,7 +57,10 @@ function CoslashPageHeader({
   return (
     <div className="flex items-center justify-between gap-4 px-4">
       <div className="flex items-center gap-2">
-        <img src="/brand/coslash-logo.svg" alt="coSlash" className="h-12" />
+        <span aria-label="coSlash">
+          <img src="/brand/coslash-logo.svg" alt="" className="h-12 dark:hidden" />
+          <img src="/brand/coslash-logo-reverse.svg" alt="" className="hidden h-12 dark:block" />
```
luhe19001, line 63: [P2] Restore an accessible name for the header logo\n\nBoth logo images now have empty alternative text, and `aria-label` on this generic `span` is not a reliable accessible name. Screen-reader users therefore lose the previously announced “coSlash” logo; give the visible image a non-empty `alt` or make the wrapper an element with an appropriate named role.

calvintvu: don't need accessibility features as of now

luhe19001 (approved):

milanshen (changes requested):
"Save Changes" button is not clickable after toggle on the dark mode (looks like it saves auto, and takes effect when user clicks outside the dialog)

milanshen (approved):

PR #44, opened 2026-08-06
Read commit messages from the -F and -am commit forms

calvintvu:
## Context

The session inspector showed `(commit)` for every commit in session 5088f0d6. `CommitMessage` read only `-m "…"`. Every commit in that session used `git commit -F -`, which reads the message from stdin.

## Changes

- When `-F` comes before any `-m`, read the message from the here-document.
- Accept `--message=`, `-am`, and `-m` with no space before the quote.
- List the subject line only, so no body and no trailer reaches the inspector, the handoff, or the synthesis prompt.
- Do not count `git commit --amend` with no message as a new commit.
- Add `commits_test.go` with the 14 command forms found in real transcripts.

Note: a `grep` pattern that contains `|git commit -m` still registers as a commit. One occurrence in 194 sessions. A regex cannot separate that case.

## Test

- `make check` — passed
- `make test` — passed
- Manual: 194 local sessions through `/api/sessions`. Placeholder rows fell from 22 to 1 of 60. Session 5088f0d6 lists all 11 subjects.

### Screenshots

before
Screenshot 2026-08-05 at 17 57 30
after
Screenshot 2026-08-05 at 17 57 49

### DIFF
--- a/collector/internal/session/commits.go
+++ b/collector/internal/session/commits.go
@@ -11,8 +11,8 @@ var gitCommitCommand = regexp.MustCompile(`(?:^|[\n;&|])\s*git\s+commit(?:$|[\s;
 // -m, --message, or a combined short flag such as -am, single or double-quoted
 var commitMessage = regexp.MustCompile(`(?:--message|-[a-zA-Z]*m)[=\s]*(?:"([^"]+)"|'([^']+)')`)
 
-// -F or --file, which read the message from a file instead of the command line
-var commitFileFlag = regexp.MustCompile(`(?:--file|-[a-zA-Z]*F)[=\s]`)
+// -F or --file with its operand, which name the file the message comes from
+var commitFileFlag = regexp.MustCompile(`(?:--file|-[a-zA-Z]*F)[=\s]+(\S+)`)
 
 var commitAmend = regexp.MustCompile(`--amend\b`)
 
@@ -37,9 +37,12 @@ func CommitMessage(command string) (string, bool) {
 
 func commitSubject(command string) string {
 	flag := commitMessage.FindStringIndex(command)
-	file := commitFileFlag.FindStringIndex(command)
+	file := commitFileFlag.FindStringSubmatchIndex(command)
 	if file != nil && (flag == nil || file[0] < flag[0]) {
-		body, _ := unwrapMultilineMessage(command) // `-F -` takes stdin
+		if command[file[2]:file[3]] != "-" {
+			return "" // a named file, whose contents the transcript does not hold
+		}
+		body, _ := unwrapMultilineMessage(command) // `-F -` reads stdin
 		return firstLine(body)
 	}
 	if flag == nil {
--- a/collector/internal/session/commits.go
+++ b/collector/internal/session/commits.go
@@ -8,8 +8,13 @@ import (
 // git commit as a complete command word, excludes plumbing commands (i.e.  commit-tree)
 var gitCommitCommand = regexp.MustCompile(`(?:^|[\n;&|])\s*git\s+commit(?:$|[\s;&|])`)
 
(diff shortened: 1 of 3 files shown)


collector/internal/session/commits.go
```diff
@@ -20,19 +25,40 @@ func CommitMessage(command string) (string, bool) {
...
+		return "", false
+	}
+	return "(commit)", true
+}
+
+func commitSubject(command string) string {
+	flag := commitMessage.FindStringIndex(command)
+	file := commitFileFlag.FindStringIndex(command)
+	if file != nil && (flag == nil || file[0] < flag[0]) {
+		body, _ := unwrapMultilineMessage(command) // `-F -` takes stdin
```
luhe19001: **[P2] Only read a heredoc when the file argument is stdin**

When `-F` names an actual file, this still unwraps the first heredoc anywhere in the shell command and records its first line as the commit subject. For example, `git commit -F .gitmessage <<'EOF' ... EOF` commits the contents of `.gitmessage`, but this reports the heredoc content instead; inspect the `-F` operand and only unwrap when it is `-`.

collector/internal/session/commits.go
```diff
@@ -20,19 +25,40 @@ func CommitMessage(command string) (string, bool) {
...
 	// only match after git commit command, discard everything before
-	matchStart := matchIndices[0]
-	match := commitMessage.FindStringSubmatch(command[matchStart:])
-	if match == nil {
-		return "(commit)", true
+	tail := command[matchIndices[0]:]
+	if subject := commitSubject(tail); subject != "" {
+		return subject, true
+	}
+	if commitAmend.MatchString(tail) {
```
luhe19001: **[P2] Restrict amend detection to the matched commit invocation**

This searches the entire remaining shell script, so a later `--amend` suppresses an earlier non-amend commit that has no extractable subject. For example, `git commit --no-edit && git commit --amend --no-edit` finds the first commit command but returns `false` because the second invocation contains `--amend`, causing the parsers to omit the created commit.

luhe19001 (commented):
Two edge cases in the new commit source/amend handling:

collector/internal/session/commits.go
calvintvu: Fixed in 31f8b82. `commitFileFlag` now captures the operand, and the unwrap runs only when that operand is `-`. A named file returns `(commit)`, because the transcript does not hold the contents of the file.

The likelier trigger than the example is `git commit -F .gitmessage && gh pr create --body "$(cat <<'EOF' ... EOF)"`. That command reported the first line of the PR body as the commit subject. Two test cases cover it: a named file with an unrelated block, and `--file=.gitmessage`.

collector/internal/session/commits.go
calvintvu: Fixed in 4390f05. The amend test now reads only the flags of the matched invocation, cut at the first `;`, `&`, `|`, or newline.

One trap here: `gitCommitCommand` consumes the leading boundary character. A slice from the match start therefore makes `& git commit --amend` cut at index 0, and amend detection stops completely. The slice starts at `matchIndices[1]-1`, which keeps the one byte that the match consumed after `commit`. A space continues into the flags, a `;&|` means the invocation carries no flags, and the end of the string lands inside `commit`.

Two test cases: your `git commit --no-edit && git commit --amend --no-edit` chain, and `make unstage & git commit -q --amend --no-edit && echo done` for the leading-separator shape. Across 194 local sessions the counts stay at 60 commits with 1 placeholder.

Known limit: `git commit \` with `--amend` on the next line. The newline ends the scan, so that amend registers as `(commit)`. The failure mode is a phantom row, not a lost commit.

luhe19001 (approved):

PR #46, opened 2026-08-06
Preserve Codex question and answer history

milanshen:
## Context

Codex input requests can contain multiple questions, with responses recorded separately. coSlash previously retained only the first question and omitted its answer, leaving the session timeline incomplete.

## Changes

- Parse every question in Codex input requests and correlate responses by call and question ID
- Preserve unanswered questions and combine multi-select responses
- Display each answer beneath its question in the session timeline

## Test

Before: if a session asked 3 questions in a row, only displayed the last question
After: all questions displayed

### DIFF
--- a/collector/internal/vendors/codex/parse.go
+++ b/collector/internal/vendors/codex/parse.go
@@ -240,8 +240,10 @@ func analyzeCodexSession(file string) (*codexSessionAnalysis, error) {
 						analysis.pullRequests++
 					}
 				}
-				if question, ok := questionFrom(row.Payload); ok {
-					analysis.digest.Push(analysis.prompts, session.DigestQuestion, question)
+				if questions, ok := questionsFrom(row.Payload); ok {
+					for _, question := range questions {
+						analysis.digest.Push(analysis.prompts, session.DigestQuestion, question)
+					}
 				}
 				analysis.notePlan(row.Payload)
 			case "function_call_output":
@@ -472,7 +474,6 @@ func unifiedDiffStat(diff string) (additions, deletions int) {
 var execCmdPattern = regexp.MustCompile(
 	`(?:^|[{,]\s*)(?:"cmd"|cmd)\s*:\s*("(?:[^"\\]|\\.)*")`,
 )
-var questionPattern = regexp.MustCompile(`"question"\s*:\s*("(?:[^"\\]|\\.)*")`)
 var planStepPattern = regexp.MustCompile(
 	`"?step"?\s*:\s*("(?:[^"\\]|\\.)*")\s*,\s*"?status"?\s*:\s*"(\w+)"`,
 )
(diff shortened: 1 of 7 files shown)


collector/internal/vendors/codex/parse.go
```diff
@@ -508,23 +514,71 @@ func (analysis *codexSessionAnalysis) notePlan(payload codexPayload) {
...
 
-func questionFrom(payload codexPayload) (string, bool) {
-	text := payload.Input
+type codexQuestion struct {
+	id   string
+	text string
+}
+
+func questionsFrom(payload codexPayload) ([]codexQuestion, bool) {
+	if payload.Name != "request_user_input" {
```
luhe19001: **[P2] Preserve wrapped user-input calls**

When Codex invokes request_user_input through an exec custom-tool wrapper, the payload name is exec and the request appears in payload.Input. The previous parser explicitly detected this form, but this guard now drops it entirely, so those sessions lose their question digest entries. Continue recognizing wrapped requests while adding support for direct calls.

collector/internal/vendors/codex/parse.go
```diff
@@ -508,23 +514,71 @@ func (analysis *codexSessionAnalysis) notePlan(payload codexPayload) {
...
 	}
-	if payload.Name != "request_user_input" && !strings.Contains(text, "request_user_input") {
-		return "", false
+	return questions, true
+}
+
+func questionAnswersByCall(rows []codexRow) map[string]map[string][]string {
+	answersByCall := make(map[string]map[string][]string)
+	for _, row := range rows {
+		if row.Type != "response_item" || row.Payload.Type != "function_call_output" || row.Payload.CallID == "" {
```
luhe19001: **[P2] Preserve answers from wrapped user-input calls**

An `exec`-wrapped `request_user_input` is recorded as a `custom_tool_call`, and its matching response uses `custom_tool_call_output` with the same call ID. This pre-pass accepts only `function_call_output`, so even after wrapped questions are recognized, their answers never enter `questionAnswers` and the digest incorrectly shows them as unanswered. Include wrapped custom-tool outputs when correlating answers.

calvintvu:
## Review

This review covers complexity only. It does not cover correctness, security, or performance.

### 1. The answer travels inside a display string

`collector/internal/vendors/codex/parse.go:247` and `frontend/src/pages/coslash/components/SessionInspector.tsx:495-497`

The Go code joins the question and the answer with a `" ↳ "` separator. The React code then splits them apart with `indexOf` and `slice`. Add `Answer string` with the tag `json:"answer,omitempty"` to `session.DigestEntry`. Set the field on the entry, then render `entry.answer`. This removes the concatenation and the 3-line split. A question that contains the separator glyph currently splits at the wrong position.

### 2. A unicode glyph acts as a UI icon

`frontend/src/pages/coslash/components/SessionInspector.tsx:506`

The line `↳ {answer}` uses a glyph as an icon. `frontend/CLAUDE.md` forbids this. Use a lucide icon (`CornerDownRight`) or plain indentation.

### 3. The payload text idiom is duplicated

`collector/internal/vendors/codex/parse.go:522-527`

`questionsFrom` reads `payload.Arguments` first, then falls back to `payload.Input`. This is the fourth copy of that block. It also reverses the priority that `notePlan` and `commandFrom` use. Extract one function, `payloadText(p codexPayload) string`. Then call it in all three places.

### 4. The bool return value is derivable

`collector/internal/vendors/codex/parse.go:522`

`questionsFrom` returns `([]codexQuestion, bool)`. Every path that returns a non-empty slice also returns `true`. Return the slice alone. Then test `len(questions) > 0` at the call site.

### 5. Two decode paths for one format

`collector/internal/vendors/codex/parse.go:563-581`

`questionAnswersFrom` decodes `Output` as a raw value and as a JSON string. `spawnedAgentID` in the same file already fixes the format: `Output` is always a JSON-encoded string. Remove the raw path and match `spawnedAgentID`.

### 6. A loop rebuilds one map into another

`collector/internal/vendors/codex/parse.go:575-580`

The loop copies `map[string]struct{Answers []string}` into `map[string][]string`. Name the inner struct once, then return `result.Answers` directly. Index it as `answers[id].Answers` at the use site.

### Total

`net: -22 lines possible.`

### Two notes outside the complexity scope

- The `codex` package holds no test file. `questionsFrom` and `questionAnswersFrom` each hold four branches. Add one `parse_test.go` with two cases: one multi-question row with answers, and one malformed payload.
- `CoslashPage.tsx` and `SessionCard.tsx` hold format-only changes. Both are correct. Both are unrelated to this PR.

collector/internal/vendors/codex/parse.go
```diff
@@ -508,30 +511,99 @@ func (analysis *codexSessionAnalysis) notePlan(payload codexPayload) {
...
-	m := questionPattern.FindStringSubmatch(text)
-	if m == nil {
-		return "requested user input", true
+	if len(questions) == 0 {
+		return []codexQuestion{{text: "requested user input"}}
+	}
+	return questions
+}
+
+type codexQuestionAnswer struct {
```
calvintvu, line 565: **[P2] Preserve legacy Codex answers**

Codex rollouts from the initial `request_user_input` format store each answer as `{"selected":[...],"other":"..."}`. This decode succeeds but leaves `Answers` empty because `codexQuestionAnswer` only has `answers`.

As a result, questions from those sessions appear unanswered. Decode both formats. Then combine the `selected` values and the nonempty `other` text.

collector/internal/vendors/codex/parse.go
```diff
@@ -240,8 +241,11 @@ func analyzeCodexSession(file string) (*codexSessionAnalysis, error) {
 						analysis.pullRequests++
 					}
 				}
```
calvintvu, line 243: **Optional: Remove the `user_note:` prefix**

Codex stores free-form answers as `user_note: <text>`. The current join sends this internal prefix to the session timeline.

Remove the prefix before you store the digest answer.

calvintvu (approved):

PR #48, opened 2026-08-07
Settle forked-skill subagents from their own transcript

calvintvu:
## Context

A subagent kept the `running` badge after its session closed. The subagent came from a forked skill, for example `/code-review`. No tool call spawns a forked skill, so its `meta.json` carries no `toolUseId`. `subagentStatus` matched that empty key against the completed tool results of the parent, and the match never succeeded.

## Changes

- When its own transcript is complete, a forked-skill subagent now reads `returned`. Before this change it read `running` forever.
- The fallback needs an empty spawn key. A normal Task subagent keeps the tool-result path, which stays authoritative. That path settles a child only after the parent writes the tool result, so a child mid-handoff never flickers to `returned`.
- `InTurn` is the signal, and it already clears on a terminal `stop_reason` (`claude/parse.go:275`). The fix reads a value the parser produces today and adds no new parsing.
- Note: the digest gap remains. `linkSpawnDigest` still logs `has no spawn row in the parent transcript` for a forked skill, so the subagent shows in the rail with no turn anchor.

## Test

- `go vet ./...` — passed
- `go test ./...` — passed
- Manual: `List(0)` over the local transcripts. Subagent `agent-a5101e8ad9c72bdc6` in session `cebc97ec` reads `returned`. It read `running` before the fix.

### Screenshots

with session closed
before
Screenshot 2026-08-06 at 18 03 50

after
Screenshot 2026-08-06 at 18 04 05

### DIFF
--- a/collector/internal/collector/helpers.go
+++ b/collector/internal/collector/helpers.go
@@ -106,9 +106,17 @@ func subagentStatus(
 	if _, ok := parent.Completed[child.SpawnKey]; ok {
 		return session.SubagentReturned
 	}
-	// A forked skill is not spawned by a tool call, fall back to child transcript
-	if child.SpawnKey == "" && !child.InTurn {
-		return session.SubagentReturned
+	// A forked skill is not spawned by a tool call, so its meta.json carries no
+	// toolUseId and it never reaches parent.Completed. Settle it from its own
+	// transcript, which clears InTurn on a terminal stop_reason.
+	if child.SpawnKey == "" {
+		if !child.InTurn {
+			return session.SubagentReturned
+		}
+		// The transcript stops mid-turn when the run dies with its parent.
+		if _, live := metadata.Live[parent.Session.ID]; !live {
+			return session.SubagentAborted
+		}
 	}
 	return session.SubagentRunning
 }
--- a/collector/internal/collector/helpers.go
+++ b/collector/internal/collector/helpers.go
@@ -106,5 +106,9 @@ func subagentStatus(
 	if _, ok := parent.Completed[child.SpawnKey]; ok {
 		return session.SubagentReturned
 	}
+	// A forked skill is not spawned by a tool call, fall back to child transcript
+	if child.SpawnKey == "" && !child.InTurn {
+		return session.SubagentReturned
+	}
 	return session.SubagentRunning
 }
--- a/collector/internal/vendors/claude/parse.go
+++ b/collector/internal/vendors/claude/parse.go
@@ -95,6 +95,10 @@ func analyzeClaudeSession(file string) (*claudeSessionAnalysis, error) {
 		return nil, err
 	}
 	analysis := &claudeSessionAnalysis{
+		// A subagent transcript opens mid-turn: the spawn already happened, and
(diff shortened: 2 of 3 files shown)


collector/internal/collector/helpers.go
```diff
@@ -106,5 +106,9 @@ func subagentStatus(
 	if _, ok := parent.Completed[child.SpawnKey]; ok {
 		return session.SubagentReturned
 	}
+	// A forked skill is not spawned by a tool call, fall back to child transcript
+	if child.SpawnKey == "" && !child.InTurn {
```
luhe19001: **[P1] Track forked-skill activity before marking it returned**

For forked skills, Claude writes the initial user row with `isMeta: true`; `claude/parse.go` excludes meta rows from `promptText`, so `InTurn` remains false even while the skill is actively issuing tool calls. Consequently, every active forked skill with an empty `SpawnKey` is immediately reported as returned. Derive activity/completion from transcript events that include forked-skill meta prompts rather than relying on the current `InTurn` value.

calvintvu: Confirmed, and fixed in a9d3321. Thank you.

I reproduced the fault against the transcript of a real forked skill. I parsed truncated prefixes of the file and read `InTurn` at each one:

| lines | state | `InTurn` |
| --- | --- | --- |
| 1 | prompt row only | false |
| 20 | mid-run | false |
| 50 | mid-run, 18 tool calls | false |
| 99 | final `end_turn` | false |

`InTurn` was false at every point, exactly as you described. The transcript holds one prompt row with `isMeta: true`, and 35 tool-result rows. `emitPrompt` never fired, so nothing raised `inTurn`. My guard reported `returned` from the first record of the agent. The finished transcript gave the correct answer only by accident.

## The fix

The cause was in the parser, so the fix is there. A subagent transcript opens mid-turn, because the spawn already happened:

```go
// claude/parse.go
inTurn: ParentIDFromPath(file) != "",
```

The terminal `stop_reason` handler at `parse.go:281` then clears it. `InTurn` now tracks the forked skill correctly, and `subagentStatus` reads a value that means what its name says.

`subagentStatus` also gains a second branch. A run that dies with its parent stops mid-turn and leaves `InTurn` true, so the transcript alone cannot settle it. The liveness of the parent settles it instead:

```go
if _, live := metadata.Live[parent.Session.ID]; !live {
    return session.SubagentAborted
}
```

## Test

Same truncated prefixes, after the fix:

| lines | live parent | closed parent |
| --- | --- | --- |
| 1, 20, 50, 98 | `running` | `aborted` |
| 99 | `returned` | `returned` |

`go vet ./...` and `go test ./...` pass. Over the local transcripts, 297 subagents in 202 sessions settle the same before and after the change: 259 `returned` and 38 `aborted`. The fix changes no verdict for a finished subagent. It only stops the wrong verdict for a live one.

`InTurn` has two readers. `resolveStatus` reads it for roots, and a child is never a root. `subagentStatus` reads it for children. The comment on the field in `transcript.go` now gives both meanings.

luhe19001 (approved):

PR #49, opened 2026-08-10
feat(canvas): port Session, DaGama, and Atlas Canvas as a coSlash plugin

luhe19001:
Ports the three Canvas products — Session Canvas, DaGama Canvas, and Atlas Canvas — from the legacy Fleetlog frontend into coSlash as a compile-time plugin.

64 commits · 311 files · +71,333 · 482 Go tests in `canvas/` · 374 frontend tests

## ⚠️ Read before reviewing depth-first

**This is inert on `main` until Task 19.** The destination is deliberately not registered with the plugin shell, so no user can open Canvas after this merges. From `frontend/src/plugins/canvas/atlas/index.ts`:

> The destination is not registered with the plugin shell here: shared lazy registration and destination readiness are Task 19 integration work, and an incomplete destination must stay hidden until then.

If you would rather trunk receive a feature someone can actually open, **hold this until Task 19 lands** and I will rebase. Merging now is safe but ships code nothing reaches.

## What moved, and why it is not a port

The legacy products ran privileged logic in Vite middleware. Everything that touches the filesystem, Git, processes, terminals, agents, artifacts, verification, or publication now lives behind authenticated Go APIs in `collector/`. The frontend keeps the legacy designs and workflows and uses coSlash's session model, auth, theme, and embedded-frontend build.

Deliberately **not** carried across: ttyd, and privileged Vite middleware. Terminals are native PTY over a guarded WebSocket.

## Architecture worth knowing before reading the diff

- **Runs are event-sourced.** `events.jsonl` *is* the run; `run.json` is a rebuildable view. Every append is fsynced before the effect it authorizes, so a crash between intent and effect leaves the intent recorded and reconciliation can see it.
- **Session identity is composite `{agent, id}`.** A bare id is not unique across agents — the reference fixtures deliberately give a Claude session and a Codex session the same one.
- **Board writes are revision-checked.** Two tabs editing one board is the ordinary case; the stale writer is refused, never silently wins.
- **Atlas stages are committees.** N workers draft into isolated directories, then one main seat consolidates. Controls address an *attempt*, not a component.

## Verification

```
cd collector && go test -race ./...     # clean, whole module
cd collector && make check              # clean
cd frontend  && npm test                # 374 passed
cd frontend  && npm run build           # clean
```

`collector/internal/plugins/canvas/hardening/` (50 tests) asserts the acceptance criteria against the assembled plugin behind the real `httpsec.Guard` — token, host, origin, method, size, traversal, symlink, command injection, WebSocket, file-render, goroutine/PTY/timer leak counts, plus every Atlas and migration acceptance row.

## What is NOT verified — please do not read this as a product sign-off

- **No browser, visual, or viewport matrix.** This repository has no jsdom and no browser environment. Frontend behavior is covered by `renderToStaticMarkup`, which asserts markup and not layout.
- **No live provider run.** Claude, Codex, tmux, and publication are exercised through fakes. What is verified is that the collector drives them correctly — not that a real provider behaves as expected.

The release gate this PR passes is the collector's, not the product's.

## Known gaps, recorded rather than glossed

- Task 19: destination registration, readiness, and the migration HTTP route.
- Legacy import copies boards and runs but not artifact blobs; nothing walks a legacy directory (the legacy backend is not in the reference repo, so its on-disk root cannot be inferred).
- Atlas restart reconciliation is unimplemented.
- Only the `plan → build → review` starter chain is runnable; a custom graph is editable and the editor names why it cannot run.

## Suggested review path

Reading 71k lines linearly is not useful. In dependency order:

1. `collector/internal/plugins/canvas/contracts/` — the frozen surface everything else agrees on
2. `runfs/` — scoped IO and the event log every store is built from
3. `dagama/runstore.go` + `reducer.go` — the state machine, then Atlas's variant of it
4. `hardening/` — what the assembled plugin is actually asserted to do
5. Frontend: `plugins/canvas/shared/`, then one product

## Real defects found and fixed along the way

- DaGama accepted **any** run-closing status and wrote it through; `isTerminal` then read it back as live, leaving a finished run every control offered to advance.
- Approve-without-publish published anyway.
- After a takeover, the original turn and the handback path both advanced the run, appending a gate onto a run that already had one open.
- `JSONObjectValidator` accepted a bare `null` — `json.Unmarshal` succeeds on it and leaves the map nil.
- Publication did not re-check that the staged tree still matched the reviewed tree; an agent editing the worktree mid-publication could ship a commit that was never approved.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


luhe19001:
Closing: this work is staying on the fork (luhe19001/coslash-exp) for now rather than targeting upstream main. Nothing here was merged, and main is untouched at 1bfe2e2. Not a comment on the changes — a decision about where they live.

PR #50, opened 2026-08-10
Support non-file session sources in the collector

calvintvu:
## Context

Session collection currently assumes each session is backed by its own transcript file. This refactor moves file-specific behavior behind an adapter and uses per-session activity for synthesis cache revisions, allowing non-file sources (i.e. opencode) without changing Claude or Codex behavior.

## Changes

- Add source-level collection, lookup, and health callbacks while retaining Claude and Codex through a file adapter.
- Invalidate synthesis per session using last activity while preserving compatibility with existing cache records.

## Test

- `go test ./...` - passed
- `go vet ./...` - passed

### DIFF
--- a/collector/internal/collector/collector.go
+++ b/collector/internal/collector/collector.go
@@ -23,6 +23,14 @@ const (
 )
 
 type vendorSource struct {
+	name    string
+	collect func(since int64) ([]*vendors.ParsedTranscript, *vendors.SessionMetadata, error)
+	get     func(id string) (*vendors.ParsedTranscript, error)
+	health  func() SourceHealth
+	fork    func(parsed []*vendors.ParsedTranscript)
+}
+
+type fileSource struct {
 	name     string
 	root     func() (string, error)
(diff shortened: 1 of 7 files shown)


collector/internal/collector/collector.go
```diff
@@ -23,6 +23,14 @@ const (
 )
 
 type vendorSource struct {
+	name    string
+	collect func(since int64) ([]*vendors.ParsedTranscript, *vendors.SessionMetadata, error)
+	get     func(id string) (*vendors.ParsedTranscript, error)
+	health  func() SourceHealth
```
milanshen, line 29: Non-blocking: should `SourceHealth` be generalized along with these callbacks? Diagnostics still derive health from `SourceScan.Files`, and `sessionFileCount` only recognizes Claude and Codex. A genuinely non-file-backed source could therefore report misleading counts unless it fabricates file data. Fine as a follow-up, but worth tracking with the first non-file source.

milanshen (approved):

PR #51, opened 2026-08-11
Make session source diagnostics storage-neutral

calvintvu:
## Context

follow up to https://github.com/centauri-ai/coslash/pull/50

Diagnostics currently inspect transcript files and derive vendor-specific session counts. This refactor moves source-specific health reporting behind vendor adapters so file- and database-backed sources can use the same diagnostics contract.

## Changes

- Report normalized source entries, root sessions, missing state, skipped entries, and errors without exposing file scans to diagnostics.
- Update diagnostic checks, CLI output, and the diagnostics dialog to use source-neutral counts and wording.
- Count only root Codex rollouts as sessions.

## Test

- `make test` - passed
- `make check` - passed
- `npm test` - passed
- `npm run lint` - passed
- `npm run build` - passed
- Manual: `go run ./cmd/coslash doctor` reported source entries and root sessions for Claude and Codex.

### Screenshots

- [x] Diagnostics dialog - source facts show session counts without file-specific wording
Screenshot 2026-08-11 at 09 31 55

### DIFF
--- a/collector/internal/collector/helpers.go
+++ b/collector/internal/collector/helpers.go
@@ -33,10 +33,26 @@ func adaptFileVendor(source fileSource) vendorSource {
 		health: func() SourceHealth {
 			root, err := source.root()
 			if err != nil {
-				return SourceHealth{Agent: source.name, ScanErr: err}
+				return SourceHealth{Agent: source.name, Err: err}
 			}
 			scan, err := source.scan()
-			return SourceHealth{Agent: source.name, Root: root, Scan: scan, ScanErr: err}
+			if err != nil {
+				return SourceHealth{Agent: source.name, Root: root, Err: err}
+			}
+			health := SourceHealth{
+				Agent:        source.name,
+				Root:         root,
+				Entries:      len(scan.Files),
+				Missing:      scan.RootMissing,
+				Skipped:      scan.Skipped,
+				SkippedTotal: max(scan.SkippedTotal, len(scan.Skipped)),
+			}
+			for _, file := range scan.Files {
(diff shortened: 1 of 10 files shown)


collector/internal/collector/helpers.go
```diff
@@ -33,10 +33,26 @@ func adaptFileVendor(source fileSource) vendorSource {
...
+			health := SourceHealth{
+				Agent:        source.name,
+				Root:         root,
+				Entries:      len(scan.Files),
+				Missing:      scan.RootMissing,
+				Skipped:      scan.Skipped,
+				SkippedTotal: max(scan.SkippedTotal, len(scan.Skipped)),
+			}
+			for _, file := range scan.Files {
+				if source.isRoot(file) {
```
milanshen: [P3 · non-blocking] The root classifier is boolean-only, while Codex header inspection maps open, decode, and header-validation failures to false. With both valid and malformed or unreadable roots, diagnostics can report state=ok and skippedTotal=0 even though collection skips the bad entry. Consider returning a tri-state/error and including classification failures in Skipped/SkippedTotal.

milanshen (approved):

collector/internal/collector/helpers.go
calvintvu: Fixed in `b2ef606`. The root classifier now returns `(bool, error)`. Codex now returns open, decode, and header-validation errors. Source health adds these errors to `Skipped` and `SkippedTotal`. A regression test includes valid entries and a classification error in the same scan.

PR #60, opened 2026-08-11
Enable releases from GitHub Actions

milanshen:
## Context

The manual release workflow only builds and smoke-tests artifacts. Publishing still requires creating protected tags outside GitHub Actions.

Add guarded Actions UI modes for publishing an RC from `main` and promoting an accepted RC commit to stable, while retaining human-pushed tags as a fallback.

## Changes

- Add `dry-run`, `publish-rc`, and `promote-stable` operations with input and release-state validation.
- Build and smoke-test the exact release commit before creating a protected tag.
- Publish verified RC and stable assets through Release Drafter, updating Homebrew only for stable releases.
- Gate manual publication through the `release` environment and a narrowly scoped GitHub App.
- Prevent duplicate runs from App-created tags while preserving human tag-push releases.
- Support safe retries for matching tags, drafts, and assets.

## Repository setup

- [x] Created the `release` environment and restricted it to protected branches.
- [x] Configured `RELEASE_APP_CLIENT_ID` and `RELEASE_APP_PRIVATE_KEY`.
- [x] Installed the release GitHub App and added it to the protected-tag ruleset bypass list.

## Test

- `go run github.com/rhysd/actionlint/cmd/actionlint@v1.7.7 -ignore 'label "macos-15-intel" is unknown' .github/workflows/release.yml` — passed
- All 15 embedded shell blocks passed `bash -n`.
- Manual: verified the workflow's `gh release view` command reads the current rolling draft outside a Git checkout.
- Live publication not run because it would create public tags and releases.


.github/workflows/release.yml
```diff
@@ -111,57 +339,132 @@ jobs:
...
+              echo "error: release drafter would overwrite the existing $draft_tag draft" >&2
+              exit 1
+            fi
+          done <<<"$draft_tags"
+      - name: Create release App token
+        if: needs.plan.outputs.create_tag == 'true'
+        id: release-app
+        uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0
+        with:
+          client-id: ${{ vars.RELEASE_APP_CLIENT_ID }}
```
luhe19001, line 375: **[P1] Pass `app-id` to create the release token**

Please replace `client-id` with the action’s required `app-id` input. `actions/create-github-app-token` does not consume `client-id`, so both `publish-rc` and `promote-stable` will fail before creating their protected tags.

milanshen: The pinned v3.2.0 action declares client-id as the preferred input and deprecates app-id. Its runtime reads client-id first and falls back to app-id, so the existing configuration is intentional.

.github/workflows/release.yml
```diff
@@ -1,22 +1,252 @@
...
+          INPUT_OPERATION: ${{ inputs.operation }}
+          INPUT_SOURCE_RC: ${{ inputs.source_rc }}
+          INPUT_VERSION: ${{ inputs.version }}
+          REPOSITORY: ${{ github.repository }}
+        run: |
+          set -euo pipefail
+
+          version_pattern='^v[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z][0-9A-Za-z.-]*)?$'
+          stable_pattern='^v[0-9]+\.[0-9]+\.[0-9]+$'
+          rc_pattern='^v[0-9]+\.[0-9]+\.[0-9]+-rc\.[1-9][0-9]*$'
```
calvintvu: **[P2] Reject noncanonical SemVer before the workflow creates the tag**

`publish-rc` accepts `v01.2.3-rc.1`, and `promote-stable` accepts `v01.2.3`. These values have leading zeros and are not canonical SemVer.

Release Drafter first uses strict `parse()`. If that function fails, it uses [`coerce()`](https://github.com/release-drafter/release-drafter/blob/34d80673e067bdc0c24568d3af899c216adcfaa9/src/actions/drafter/lib/build-release-payload/version-descriptor.ts#L99-L106), which changes `01.2.3-rc.1` to `1.2.3`.

The workflow creates the protected tag before Release Drafter runs. Release Drafter then writes `VERSION=v1.2.3` in the release notes. The later check expects `VERSION=v01.2.3-rc.1` and fails. This leaves the protected tag and draft release behind.

Require each numeric component to match `0|[1-9][0-9]*`.

.github/workflows/release.yml
luhe19001: Confirmed against the exact pinned v3.2.0 action metadata: `client-id` is supported and preferred, while `app-id` is deprecated. Thanks for the correction—resolving.

.github/workflows/release.yml
milanshen: Good catch—fixed locally in a follow-up commit. The workflow now rejects leading-zero major, minor, and patch components before creating a tag.

calvintvu (approved):

PR #61, opened 2026-08-11
Add OpenCode subagent support

calvintvu:
## Context

OpenCode stores delegated tasks as child sessions in its SQLite database. Root-session support currently omits this delegated work from dashboard cards and the session inspector. This adds direct child-session parity with the existing vendors.

## Changes

- Load non-archived direct child sessions and retain roots when child activity is newer.
- Link task metadata to child names, results, models, durations, commands, tokens, exact costs, and digest entries.
- Normalize running, returned, and aborted states while deduplicating repeated task links.

## Test

- `make test` - passed
- `make check` - passed
- `npm test` - passed
- `./node_modules/.bin/oxlint` - passed with two existing warnings
- `npm run build` - passed
- `CGO_ENABLED=0 GOOS=darwin GOARCH=arm64 go build ./...` - passed
- `CGO_ENABLED=0 GOOS=darwin GOARCH=amd64 go build ./...` - passed
- Manual: Loaded an active OpenCode database and verified linked child sessions, digest entries, and normalized statuses.

### Screenshots

- [x] Dashboard - OpenCode root card with its subagent rail
Screenshot 2026-08-11 at 15 31 25

- [x] Session inspector - linked subagent digest row and detail dialog
Screenshot 2026-08-11 at 15 31 41

### DIFF
--- a/collector/internal/vendors/opencode/source.go
+++ b/collector/internal/vendors/opencode/source.go
@@ -15,10 +15,48 @@ import (
 	"github.com/centauri-ai/coslash/collector/internal/vendors"
 )
 
-const activeRootsQuery = `
-SELECT id, directory, title, summary_files, summary_diffs, model, cost, time_updated
-FROM session
-WHERE parent_id IS NULL AND time_archived IS NULL`
+const activeFamiliesQuery = `
+WITH selected_roots AS (
+	SELECT root.id,
+		MAX(root.time_updated,
+			COALESCE(MAX(child.time_updated), root.time_updated),
+			COALESCE(MAX(child.time_archived), root.time_updated)) AS family_updated
+	FROM session AS root
+	LEFT JOIN session AS child ON child.parent_id = root.id
+	WHERE root.parent_id IS NULL AND root.time_archived IS NULL
+	GROUP BY root.id
+)
+SELECT member.id, member.parent_id, member.directory, member.title, member.summary_files,
+	member.summary_diffs, member.model, member.cost,
+	CASE WHEN member.id = selected_roots.id THEN selected_roots.family_updated ELSE member.time_updated END
+FROM selected_roots
+JOIN session AS member ON member.id = selected_roots.id OR member.parent_id = selected_roots.id
+WHERE member.time_archived IS NULL`
+
+const activeRootQuery = `
+SELECT root.id, root.parent_id, root.directory, root.title, root.summary_files,
+	root.summary_diffs, root.model, root.cost,
+	MAX(root.time_updated, COALESCE((
+		SELECT MAX(child.time_updated)
+		FROM session AS child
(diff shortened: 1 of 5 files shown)


collector/internal/vendors/opencode/source.go
```diff
@@ -115,45 +154,69 @@ func load(db *sql.DB, query string, args ...any) ([]*vendors.ParsedTranscript, e
 		return nil, err
 	}
 
-	parsed := make([]*vendors.ParsedTranscript, 0, len(stored))
+	parsed := make([]parsedSession, 0, len(stored))
 	for _, row := range stored {
 		item, err := parse(tx, row)
 		if err != nil {
 			return nil, fmt.Errorf("parse OpenCode session %q: %w", row.id, err)
```
luhe19001, line 161: **[P2] Isolate malformed child sessions**

When any newly loaded child session contains malformed model, message, part, or summary JSON, this return aborts the entire OpenCode collection and hides every otherwise valid OpenCode family. Previously such child rows were never parsed; please skip the affected family (while reporting it) rather than failing the whole source.

calvintvu: Fixed in #63. The collector now parses OpenCode sessions by family. If a session contains malformed JSON, the collector logs and skips only that family. Other valid families remain available.

collector/internal/vendors/opencode/source.go
```diff
@@ -240,9 +303,33 @@ func parse(tx *sql.Tx, row storedSession) (*vendors.ParsedTranscript, error) {
...
+							digest.PushSubagent(turns, childID)
+						}
+						if part.State.Input.Description != "" {
+							link.name = part.State.Input.Description
+						} else if link.name == "" {
+							link.name = part.State.Input.SubagentType
+						}
+						link.status = part.State.Status
+						tasks[childID] = link
+						if part.State.Status == "completed" {
```
luhe19001, line 326: **[P2] Keep background task subagents running**

When OpenCode experimental background subagents are enabled, the task tool part becomes `completed` as soon as the task is handed off and carries `metadata.background: true`, while the child session continues running. Treating every completed task part as child completion makes `subagentStatus` report the active child as returned. Could you please parse the background flag and derive completion from the child/job state, or comment here to explicitly acknowledge and accept this risk?

luhe19001 (approved):
author to decide on the P2 warning treatment

collector/internal/vendors/opencode/source.go
calvintvu: Fixed in PR #63 with commit [23f4e812](https://github.com/centauri-ai/coslash/commit/23f4e81247111a81e5c227a1a300e375b85e6073). The collector now checks the active OpenCode child state before the completed parent task marker. A background child remains `running` until its session is no longer active.

PR #69, opened 2026-08-12
Use canonical Git remote identities for sessions

calvintvu:
## Context

Session repository grouping currently relies on local directory names, which can collide and vary across checkout layouts. Normalize origin remotes so repositories have stable identities across machines, while preserving local-only fallbacks.

## Changes

- Normalize HTTPS, SSH, and SCP-style origin URLs to `host/owner/repository`.
- Mark basename fallbacks as local-only and preserve linked worktree grouping.

## Test

- `go test ./...` - passed
- `go vet ./...` - passed
- `npm test` - passed
- `npm run build` - passed

### Screenshots

- [x] Session board - canonical remote repository label and local-only fallback
Screenshot 2026-08-12 at 11 10 46

### DIFF
--- a/collector/internal/session/probes.go
+++ b/collector/internal/session/probes.go
@@ -1,6 +1,7 @@
 package session
 
 import (
+	"net/url"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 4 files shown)


collector/internal/session/probes.go
```diff
@@ -31,49 +32,83 @@ func FileModificationTime(filePath string) int64 {
...
+	if len(parts) < 2 {
+		return ""
+	}
+	for _, part := range parts {
+		if part == "" || part == "." || part == ".." {
+			return ""
+		}
 	}
-	return filepath.Base(root)
+	return strings.ToLower(parsed.Hostname()) + "/" + path
```
luhe19001: **[P2] Preserve explicit ports in canonical repository names**

For origins hosted on non-default ports, `Hostname()` drops the port, so `https://git.example.com:8443/org/repo.git` and `https://git.example.com/org/repo.git` both become `git.example.com/org/repo`. These can be distinct servers, causing unrelated sessions to be grouped together. Please include the explicit port in the canonical identity.

collector/internal/session/probes.go
```diff
@@ -31,49 +32,83 @@ func FileModificationTime(filePath string) int64 {
...
+	}
+	parsed, err := url.Parse(remote)
+	if err != nil || parsed.Hostname() == "" {
+		return ""
+	}
+	path := strings.Trim(parsed.Path, "/")
+	if strings.HasSuffix(strings.ToLower(path), ".git") {
+		path = path[:len(path)-len(".git")]
+	}
+	parts := strings.Split(path, "/")
```
luhe19001: **[P2] Accept single-segment remote repository paths**

Valid remotes can be hosted directly at the server root, such as `ssh://git@git.example.com/repo.git`. Rejecting a single path segment falls back to the local directory name with `repoLocalOnly=true`, so differently named checkouts of the same remote are not grouped. Please reject only an empty path.

luhe19001 (commented):
Please address the two inline findings before merging.

collector/internal/session/probes.go
calvintvu: Fixed in 155bda7: canonical identity now keeps the full host including the explicit port, so git.example.com:8443 and git.example.com no longer collapse together.

collector/internal/session/probes.go
calvintvu: Fixed in 155bda7: root-hosted remotes are now accepted (single-segment path), since Git supports cloning a repository directly from the host root.

luhe19001 (approved):

PR #71, opened 2026-08-13
Refactor session collection around vendor-owned sources

calvintvu:
## Context

Session discovery, parsing, fork attribution, and enrichment were split inconsistently between the collector and vendor packages. This refactor gives Claude, Codex, and OpenCode the same source boundary while keeping shared composition and enrichment in one pipeline. Optional metadata failures no longer hide persisted sessions.

## Changes

- Move vendor-specific collection, parsing, fork normalization, targeted lookup, and health checks behind vendor-owned sources.
- Scope parent-child relationships by vendor and centralize activity, pricing, status, subagent, and Git enrichment.
- Continue targeted lookup after vendor errors and child matches, returning accumulated errors only when no root is found.
- Remove internal `logPath` and redundant `commandCount` API fields; command counts now derive from `commands`.

## Test

- `cd collector && make test` — passed
- `cd frontend && npm test` — passed
- `cd frontend && npm run lint` — passed with two pre-existing warnings
- `cd frontend && npm run format:check` — passed
- `cd frontend && npm run build` — passed
- Manual: Not run; no visible UI behavior changed


luhe19001 (approved):
reviewed and all good

PR #76, opened 2026-08-13
Show session-specific file diffs in the inspector

calvintvu:
## Context

The session inspector lists edited files but does not show the changes that produced them. This adds a per-file view backed by recorded session edits instead of the current working tree. Changes remain in session order, while file writes show their full content.

## Changes

- Make edited filenames clickable and open a dedicated file-changes panel over the inspector
- Show sequential edits with colored lines and addition/deletion statistics
- Preserve recorded changes from Claude, Codex, and OpenCode, including accurate OpenCode replacement patches
- Add a session-scoped `/api/diff` endpoint with loading, empty, and error states

## Test

- `go test ./cmd/coslash -count=1` — passed
- `go build ./...` — passed
- `npm test` — passed
- `npm run build` — passed
- Manual: not run after restarting the collector

### Screenshots

- [x] Session inspector — clickable file list and overlaid sequential diff panel
Screenshot 2026-08-13 at 16 29 01

### DIFF
--- a/collector/internal/session/file_edits.go
+++ b/collector/internal/session/file_edits.go
@@ -1,5 +1,15 @@
 package session
 
+import "strings"
+
+type FileChange struct {
+	Kind      string `json:"kind"`
+	Text      string `json:"text"`
+	Operation string `json:"operation"`
+	Additions int    `json:"additions"`
+	Deletions int    `json:"deletions"`
+}
+
 type FileEditSet struct {
 	Edits []FileEdit
 	index map[string]int
@@ -10,6 +20,11 @@ func NewFileEditSet() *FileEditSet {
 }
 
 func (s *FileEditSet) Add(path string, additions, deletions int, isNew bool) {
+	operation := "Edit"
+	if isNew {
+		operation = "Add"
+	}
+	change := FileChange{Operation: operation, Additions: additions, Deletions: deletions}
 	if i, ok := s.index[path]; ok {
 		edit := &s.Edits[i]
 		edit.Additions += additions
@@ -18,6 +33,7 @@ func (s *FileEditSet) Add(path string, additions, deletions int, isNew bool) {
 		if isNew {
 			edit.IsNew = true
 		}
+		edit.changes = append(edit.changes, change)
 		return
 	}
 	s.index[path] = len(s.Edits)
@@ -27,5 +43,72 @@ func (s *FileEditSet) Add(path string, additions, deletions int, isNew bool) {
 		Deletions: deletions,
 		Edits:     1,
 		IsNew:     isNew,
+		changes:   []FileChange{change},
 	})
+}
+
+func (s *FileEditSet) Change(path, before, after string) {
+	change := s.pendingChange(path)
+	if before == after && change.Operation != "Add" {
+		return
+	}
+	change.Kind = "diff"
+	change.Text = renderChange(path, before, after)
+}
+
+func (s *FileEditSet) Write(path, content string) {
+	change := s.pendingChange(path)
+	change.Kind = "content"
+	change.Text = content
+	change.Operation = "Write"
+}
+
+func (s *FileEditSet) Patch(path, patch string) {
+	change := s.pendingChange(path)
+	change.Kind = "diff"
+	change.Text = patch
+	if change.Operation != "Add" {
+		change.Operation = "Patch"
+	}
+}
+
+func (e FileEdit) Changes() []FileChange {
+	changes := make([]FileChange, 0, len(e.changes))
+	for _, change := range e.changes {
+		if change.Kind != "" {
+			changes = append(changes, change)
+		}
+	}
+	return changes
+}
+
+func (s *FileEditSet) pendingChange(path string) *FileChange {
+	edit := &s.Edits[s.index[path]]
+	if len(edit.changes) == 0 || edit.changes[len(edit.changes)-1].Kind != "" {
+		edit.changes = append(edit.changes, FileChange{Operation: "Edit"})
+	}
+	return &edit.changes[len(edit.changes)-1]
+}
+
+func renderChange(path, before, after string) string {
+	var output strings.Builder
+	output.WriteString("--- " + path + "\n+++ " + path + "\n@@\n")
+	writeSnippetDiff(&output, before, after)
+	return output.String()
+}
+
+func writeSnippetDiff(output *strings.Builder, before, after string) {
+	for _, line := range splitLines(before) {
+		output.WriteString("-" + line + "\n")
+	}
+	for _, line := range splitLines(after) {
+		output.WriteString("+" + line + "\n")
+	}
+}
+
+func splitLines(text string) []string {
+	if text == "" {
+		return nil
+	}
+	return strings.Split(strings.TrimSuffix(text, "\n"), "\n")
 }
--- a/collector/internal/session/file_edits.go
+++ b/collector/internal/session/file_edits.go
@@ -53,7 +53,7 @@ func (s *FileEditSet) Change(path, before, after string) {
 		return
 	}
 	change.Kind = "diff"
-	change.Text = renderChange(path, before, after)
+	change.Text = renderChange(before, after)
 }
 
 func (s *FileEditSet) Write(path, content string) {
@@ -90,9 +90,9 @@ func (s *FileEditSet) pendingChange(path string) *FileChange {
 	return &edit.changes[len(edit.changes)-1]
 }
 
-func renderChange(path, before, after string) string {
+func renderChange(before, after string) string {
 	var output strings.Builder
-	output.WriteString("--- " + path + "\n+++ " + path + "\n@@\n")
+	output.WriteString("@@\n")
 	writeSnippetDiff(&output, before, after)
 	return output.String()
 }
--- a/collector/internal/vendors/claude/parse.go
+++ b/collector/internal/vendors/claude/parse.go
@@ -311,6 +311,10 @@ func analyzeClaudeSession(file string) (*claudeSessionAnalysis, error) {
 				var diffLines []string
 				createdContent := ""
 				for _, patch := range result.StructuredPatch {
+					diffLines = append(diffLines, fmt.Sprintf(
+						"@@ -%d,%d +%d,%d @@",
+						patch.OldStart, patch.OldLines, patch.NewStart, patch.NewLines,
+					))
 					diffLines = append(diffLines, patch.Lines...)
 				}
 				lineAdditions, lineDeletions := session.DiffStat(diffLines)
--- a/collector/internal/vendors/claude/parse.go
+++ b/collector/internal/vendors/claude/parse.go
@@ -309,13 +309,13 @@ func analyzeClaudeSession(file string) (*claudeSessionAnalysis, error) {
 			}
 			if result != nil && result.FilePath != "" {
 				var diffLines []string
+				createdContent := ""
 				for _, patch := range result.StructuredPatch {
 					diffLines = append(diffLines, patch.Lines...)
 				}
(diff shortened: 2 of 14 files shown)


collector/internal/vendors/opencode/source.go
```diff
@@ -599,6 +558,104 @@ func parse(tx *sql.Tx, row storedSession) (parsedSession, error) {
...
+	if part.Tool == "edit" && len(part.State.Metadata.Files) == 0 {
+		file := part.State.Metadata.FileDiff
+		path := file.File
+		if path == "" {
+			path = part.State.Input.FilePath
+		}
+		if path != "" {
+			path = normalizeFilePath(cwd, path)
+			edits.Add(path, file.Additions, file.Deletions, false)
+			edits.Change(path, part.State.Input.OldString, part.State.Input.NewString)
```
luhe19001: **[P2] Use the recorded OpenCode patch for fallback edits**

`metadata.filediff.patch` contains the complete unified diff, including every `replaceAll` hunk. Rebuilding from `OldString` and `NewString` drops other hunks and context while retaining aggregate stats. Use `file.Patch` when present and synthesize only as a fallback.

frontend/src/pages/coslash/components/DiffList.tsx
```diff
@@ -0,0 +1,95 @@
...
+              {change.text.split('\n').map((line, lineIndex) => {
+                const kind = diffLineKind(line);
+                return (
+                  <span
+                    key={lineIndex}
+                    className={cn('block min-w-max px-3', {
+                      'text-muted-foreground': kind === 'meta',
+                      'text-brand': kind === 'hunk',
+                      'bg-success-bg text-success-fg': kind === 'addition',
+                      'bg-destructive/10 text-destructive': kind === 'deletion',
```
luhe19001: **[P3] Replace the prohibited slash-opacity utility**

`bg-destructive/10` violates `frontend/CLAUDE.md`’s solid color-scale rule. Please use the corresponding solid color-scale class.

luhe19001 (commented):
@calvintvu, please review these two findings.

collector/internal/vendors/opencode/source.go
```diff
@@ -599,6 +558,104 @@ func parse(tx *sql.Tx, row storedSession) (parsedSession, error) {
...
+		}
+		if path != "" {
+			path = normalizeFilePath(cwd, path)
+			edits.Add(
+				path,
+				file.Additions,
+				file.Deletions,
+				file.Type == "add" || file.Type == "added",
+			)
+			patch := file.Patch
```
milanshen: **P1 — fix before merge.** Defeats the feature outright for this path, and the PR description specifically claims accurate OpenCode replacement patches. Worth confirming against a real OpenCode DB how often `edit` populates `metadata.files` without a per-file `patch`.

**OpenCode `edit` with `metadata.files` records no diff.** When an `edit` part carries `metadata.files` entries that have no per-file `patch`, no change text is recorded at all: `state.input.patchText` is empty for `edit` (it passes oldString/newString), so `applyPatchForFile("", ...)` returns `""`, `edits.Patch` is never called, the pending `FileChange` keeps `Kind == ""`, and `FileEdit.Changes()` filters it out. The oldString/newString fallback only runs in the `len(Files) == 0` branch below, so it is unreachable here. Result: the inspector lists the file with +3/−1 but the panel shows "No recorded changes from this session."

collector/internal/vendors/claude/parse.go
```diff
@@ -294,26 +294,31 @@ func analyzeClaudeSession(file string) (*claudeSessionAnalysis, error) {
...
 					} else {
-						lineAdditions = session.CountLines(content)
+						lineAdditions = session.CountLines(createdContent)
 					}
 				}
 				analysis.fileEdits.Add(
 					result.FilePath, lineAdditions, lineDeletions, result.Type == "create",
 				)
+				if len(diffLines) > 0 {
+					analysis.fileEdits.Patch(result.FilePath, strings.Join(diffLines, "\n"))
```
milanshen, line 337: **P1 — fix before merge.** Claude is the most-used vendor here and multi-hunk edits are the common case, so this misrenders the majority of diffs the panel will show.

**Multi-hunk Claude patches lose their hunk boundaries.** `diffLines` flattens every `StructuredPatch[i].Lines` into one slice, and the struct only carries `Lines` (no `oldStart`/`newStart`). For an Edit whose patch has one hunk at line 10 and another at line 200, the joined text is hunk 1 immediately followed by hunk 2, with no `@@` separator for `diffLineKind` to pick up — the panel renders the two edits as if they were adjacent.

collector/internal/vendors/opencode/source.go
milanshen: **P2 — worth fixing in this PR, cheap.** Confusing rather than wrong-wrong, and the fix is small since `FileDiff.Patch` is already parsed here.

**Diff body contradicts its own header stats.** `Change` renders *every* line of `oldString` as a deletion and every line of `newString` as an addition, while the header `+n −m` comes from `FileDiff.Additions`/`Deletions`. A 10-line replacement that changes one line reports "+1 −1" in the header but shows 10 red lines followed by 10 green lines. `storedToolFile.Patch` is added in this PR — using `part.State.Metadata.FileDiff.Patch` when non-empty would render the real hunk.

frontend/src/pages/coslash/components/DiffList.tsx
```diff
@@ -0,0 +1,95 @@
+import { LoaderCircleIcon } from 'lucide-react';
+import { cn } from '@/lib/utils';
+import type { FileChange } from '@/pages/coslash/hooks/use-sessions';
+
+type DiffLineKind = 'meta' | 'hunk' | 'addition' | 'deletion' | 'context';
+
+/* oxlint-disable react/only-export-components -- exported for focused rendering tests */
+export function diffLineKind(line: string): DiffLineKind {
+  if (line.startsWith('+++') || line.startsWith('---')) return 'meta';
```
milanshen: **P3 — follow-up is fine.** Cosmetic; wrong color, right content. Note it interacts with the `parse.go:318` finding: with no real `---`/`+++` header in Claude patches, this check only ever produces false positives there.

**`+++`/`---` also match ordinary source lines.** Those prefixes are treated as file-header meta, but in a unified diff they equally occur on added/deleted lines whose own content starts with `++`/`--`. Adding the line `++i;` produces `+++i;` (rendered muted gray instead of green); deleting `-- reset counter` produces `--- reset counter` (gray instead of red). Claude patches carry no real `---`/`+++` header at all, so every such line in them is misclassified.

collector/internal/vendors/opencode/source.go
```diff
@@ -599,6 +558,104 @@ func parse(tx *sql.Tx, row storedSession) (parsedSession, error) {
 	}, nil
 }
 
+func addCompletedToolEdit(cwd string, part *storedPart, edits *session.FileEditSet) bool {
+	if part.Tool == "write" {
+		file := storedToolFile{}
+		if len(part.State.Metadata.Files) > 0 {
+			file = part.State.Metadata.Files[0]
```
milanshen: **P3 — follow-up is fine.** A genuine regression from the refactor, but `write` writes one file, so it likely never triggers. Restore the loop when convenient.

**`write` now drops `metadata.files` beyond the first.** The replaced code iterated *every* entry of `Metadata.Files` for all tools including `write`; this branch reads only `Files[0]`. If a `write` part records two entries, the second file no longer reaches `fileEdits.Add` and disappears from the edited-file list along with its additions/deletions.

collector/internal/vendors/opencode/source.go
calvintvu: Commit `f833466` fixes this error. OpenCode edits now prefer an available `metadata.filediff.patch` value. If the recorded patch is empty, the code uses `oldString` and `newString`.

frontend/src/pages/coslash/components/DiffList.tsx
calvintvu: Commit `ad2fd73` fixes this error. The deletion background now uses the solid `bg-red-50` and `dark:bg-red-950` classes.

collector/internal/vendors/opencode/source.go
calvintvu: Commit `f833466` fixes this error. An edit with `metadata.files` now uses `metadata.filediff.patch`. If the recorded patch is empty, the code uses `oldString` and `newString`.

collector/internal/vendors/claude/parse.go
calvintvu: Commit `c3b7184` fixes this error. The parser now keeps the four recorded hunk coordinates. It adds an `@@ ... @@` header before each hunk.

collector/internal/vendors/opencode/source.go
calvintvu: Commit `f833466` fixes this error. The panel now uses the recorded OpenCode patch. As a result, the displayed body agrees with the recorded line counts.

frontend/src/pages/coslash/components/DiffList.tsx
calvintvu: Commit `ad2fd73` fixes this error. The renderer no longer treats all `+++` and `---` prefixes as metadata. These lines now use addition or deletion colors.

collector/internal/vendors/opencode/source.go
calvintvu: We did not make this change. The native OpenCode `write` tool accepts one file path and one content value. A multi-file record requires `apply_patch` or a custom tool.

collector/internal/session/file_edits.go
```diff
@@ -27,5 +43,72 @@ func (s *FileEditSet) Add(path string, additions, deletions int, isNew bool) {
...
+	edit := &s.Edits[s.index[path]]
+	if len(edit.changes) == 0 || edit.changes[len(edit.changes)-1].Kind != "" {
+		edit.changes = append(edit.changes, FileChange{Operation: "Edit"})
+	}
+	return &edit.changes[len(edit.changes)-1]
+}
+
+func renderChange(path, before, after string) string {
+	var output strings.Builder
+	output.WriteString("--- " + path + "\n+++ " + path + "\n@@\n")
```
milanshen: **P2 — new regression from the `+++`/`---` fix in `ad2fd73`.** Dropping the `meta` kind from `diffLineKind` was the right call for source lines, but `renderChange` still emits a `--- path` / `+++ path` header, and nothing classifies those any more.

Verified output for a Claude-created file:

```
kind="diff" op="Add" text="--- a.ts\n+++ a.ts\n@@\n+hello\n"
```

`diffLineKind("--- a.ts")` now returns `deletion` and `diffLineKind("+++ a.ts")` returns `addition`, so every `Change`-rendered block opens with a fake red "deleted" line and a fake green "added" line, both showing the filename that is already in the sheet header two rows above. This hits Claude file creations (`parse.go` → `Change(path, "", createdContent)`), Codex new files without a `unified_diff`, and the OpenCode `edit` oldString/newString fallback.

Simplest fix is to stop emitting the header entirely and write only `"@@\n"`, since the panel already displays the path. That leaves `path` unused in `renderChange`, so drop the parameter and update the two call sites in `Change`.

Separately: `diffLineKind` and `fileChangeLabel` carry the comment `exported for focused rendering tests`, but there is no test file for `DiffList.tsx` (frontend tests exist only under `src/pages/coslash/lib/`). A case for `--- a.ts` would have caught this one.

calvintvu: Fixed in bffc815. `renderChange` now emits only the hunk marker and changed lines, so generated snippets no longer add fake filename rows. I also removed the unused path parameter and tightened the API test to assert the exact header-free output. Verified with `go test ./cmd/coslash -count=1`.

milanshen (approved):

PR #78, opened 2026-08-14
Fix Codex subagent filtering and inspector entries

calvintvu:
## Context

Codex transcripts can include internal permission-review child sessions and newer nested subagent activity events. The collector exposed internal checks as delegated work while omitting some genuine subagents from the session inspector.

## Changes

- Exclude internal permission-review sessions using structured source metadata.
- Record nested subagent start events so genuine subagents appear consistently in the rail and inspector.
- Preserve root sessions and genuine thread-spawned subagents.

## Test

- `go test internal/vendors/codex/discovery.go internal/vendors/codex/fork.go internal/vendors/codex/metadata.go internal/vendors/codex/parse.go internal/vendors/codex/types.go internal/vendors/codex/parse_test.go -count=1` — passed
- `go build ./...` — passed
- `git diff --check` — passed

### Screenshots

- [x] Session views — genuine subagent appears in both the rail and inspector
Screenshot 2026-08-14 at 13 43 21

- [x] Permission review — internal check is absent from both views
before
Screenshot 2026-08-14 at 13 42 23

after
Screenshot 2026-08-14 at 13 42 07

### DIFF
--- a/collector/internal/vendors/codex/parse.go
+++ b/collector/internal/vendors/codex/parse.go
@@ -71,6 +71,17 @@ func (analysis *codexSessionAnalysis) noteFileChanges(changes codexPatchChanges)
 	}
 }
 
+func (analysis *codexSessionAnalysis) noteSubagentStarted(id string) {
+	if id == "" {
+		return
+	}
+	if _, exists := analysis.spawnTurns[id]; exists {
+		return
+	}
+	analysis.spawnTurns[id] = max(analysis.prompts, 1)
+	analysis.digest.PushSubagent(analysis.prompts, id)
+}
+
 func completedItemText(item codexItem) string {
 	text := make([]string, 0, len(item.Content))
 	for _, content := range item.Content {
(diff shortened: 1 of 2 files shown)


collector/internal/vendors/codex/parse.go
```diff
@@ -181,6 +186,11 @@ func analyzeCodexSession(file string) (*codexSessionAnalysis, error) {
 					}
 				case "FileChange":
 					analysis.noteFileChanges(item.Changes)
+				case "SubAgentActivity":
+					if item.Kind == "started" && item.AgentThreadID != "" {
+						analysis.spawnTurns[item.AgentThreadID] = max(analysis.prompts, 1)
```
milanshen: Duplicate spawn recording: this mirrors the existing `event_msg` / `sub_agent_activity` handler a few lines below (parse.go:215). If a Codex build emits both the legacy event and its structured `item_completed` replay for the same agent thread, `spawnTurns[agentThreadID]` is written twice and the later write wins. The duplicate digest row is harmless (`linkSpawnDigest` fills the first, `unresolvedSpawn` prunes the second), but `SpawnedAtTurn` (helpers.go:130) reads the map, so the subagent can be attributed to a later turn than the digest row it links to. Worth extracting one helper called from both cases, or skipping a key already present in `spawnTurns`.

calvintvu: Fixed in 779875c4a62f486d23f450fd6f77c555b0b4c854. Both event shapes now route through an idempotent recorder, so the first sighting owns SpawnTurns and the digest row. Added a regression test with duplicate events on different turns.

milanshen (approved):

PR #79, opened 2026-08-14
Show only surviving commits in session inspector

calvintvu:
## Context

Session commit lists were derived from observed commands, so amendments, resets, and rebases could leave obsolete rows while compound commands could omit valid commits. This change reconciles successful commit observations with the recorded branch’s surviving Git history, with a fallback when the repository is unavailable.

## Changes

- Capture commit operations and output hashes across OpenCode, Codex, and Claude.
- Remove superseded commits without collapsing distinct commits that share a subject.
- Exclude failed operations and preserve amend behavior when Git history is unavailable.

## Test

- `go test ./...` — passed
- Manual: verified amended OpenCode and Claude sessions report only surviving commits; exercised Codex success and failure records with native fixtures.

### Screenshots

- [x] Session Inspector — rewritten session showing only surviving commits

**before**
Screenshot 2026-08-14 at 15 48 49

**after**
Screenshot 2026-08-14 at 15 49 13

### DIFF
--- a/collector/internal/collector/collector.go
+++ b/collector/internal/collector/collector.go
@@ -297,6 +297,9 @@ func probeGitEnvironment(roots []*vendors.ParsedSession) {
 	wg.Wait()
 	for _, p := range roots {
 		s := p.Session
+		s.Commits = session.ReconcileCommits(
+			s.CommitLog, s.WorkingDirectory, s.Branch,
+		)
 		s.GitProbed = true // synthesis's lazy probe must not redo this
 		if s.WorkingDirectory == "" {
 			continue
--- a/collector/internal/collector/collector.go
+++ b/collector/internal/collector/collector.go
@@ -295,9 +295,10 @@ func probeGitEnvironment(roots []*vendors.ParsedSession) {
 		}()
 	}
 	wg.Wait()
+	reconcileCommits := session.NewCommitReconciler()
 	for _, p := range roots {
 		s := p.Session
-		s.Commits = session.ReconcileCommits(
+		s.Commits = reconcileCommits(
 			s.CommitLog, s.WorkingDirectory, s.Branch,
 		)
 		s.GitProbed = true // synthesis's lazy probe must not redo this
--- a/collector/internal/session/commits.go
+++ b/collector/internal/session/commits.go
@@ -103,11 +103,41 @@ type repositoryCommit struct {
 	subject string
 }
 
+// NewCommitReconciler caches repository history for one reconciliation pass.
+func NewCommitReconciler() func([]CommitObservation, string, *string) []string {
+	type cacheKey struct{ cwd, branch string }
+	type cachedHistory struct {
+		commits []repositoryCommit
+		ok      bool
+	}
+	histories := map[cacheKey]cachedHistory{}
+	return func(observations []CommitObservation, cwd string, branch *string) []string {
+		if len(observations) == 0 {
+			return []string{}
+		}
+		branchName := ""
+		if branch != nil {
+			branchName = strings.TrimSpace(*branch)
+		}
+		key := cacheKey{cwd: cwd, branch: branchName}
+		history, found := histories[key]
+		if !found {
+			history.commits, history.ok = repositoryHistory(cwd, branch)
+			histories[key] = history
+		}
+		return reconcileCommits(observations, history.commits, history.ok)
+	}
+}
+
 func ReconcileCommits(observations []CommitObservation, cwd string, branch *string) []string {
 	if len(observations) == 0 {
 		return []string{}
 	}
 	history, ok := repositoryHistory(cwd, branch)
+	return reconcileCommits(observations, history, ok)
+}
+
+func reconcileCommits(observations []CommitObservation, history []repositoryCommit, ok bool) []string {
 	if !ok {
 		return fallbackCommitMessages(observations)
 	}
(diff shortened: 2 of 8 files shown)


milanshen (commented):
I found four issues in the commit reconciliation path. The existing tests and CI pass, but these cases are not covered by the current suite.

collector/internal/session/commits.go
```diff
@@ -16,30 +17,145 @@ var commitFileFlag = regexp.MustCompile(`(?:--file|-[a-zA-Z]*F)[=\s]+(\S+)`)
...
+func repositoryHistory(cwd string, branch *string) ([]repositoryCommit, bool) {
+	if cwd == "" {
+		return nil, false
+	}
+	ref := "HEAD"
+	if branch != nil && strings.TrimSpace(*branch) != "" {
+		ref = strings.TrimSpace(*branch)
+	}
+	output, err := exec.Command(
+		"git", "-C", cwd, "log", "--format=%H%x00%s", ref, "--",
```
milanshen, line 190: **[P1] Reconcile against more than the recorded branch.** This ref can be stale: Codex records the branch in session metadata before commands run, so a session that runs `git switch -c feature` and commits will still query `main` here and drop a commit that remains reachable on `feature`. If the recorded branch was deleted, this command instead errors and the fallback resurrects every observation. Please reconcile reachability across relevant refs (or record the branch per observation) and distinguish a missing ref from an unavailable repository.

collector/internal/session/commits.go
```diff
@@ -16,30 +17,145 @@ var commitFileFlag = regexp.MustCompile(`(?:--file|-[a-zA-Z]*F)[=\s]+(\S+)`)
...
+		resolved[i] = -1
+	}
+	used := make([]bool, len(history))
+	for i, observation := range observations {
+		if observation.Hash == "" {
+			continue
+		}
+		for j, commit := range history {
+			if !used[j] && strings.HasPrefix(commit.hash, observation.Hash) &&
+				(observation.Subject == "(commit)" || commit.subject == observation.Subject) {
```
milanshen, line 155: **[P2] Do not reject an exact hash because subject parsing differs.** Shell expansion makes this happen for ordinary commands such as `SUBJECT="expanded value"; git commit -m "$SUBJECT"`: Git reports the correct hash and subject, but the command parser records `$SUBJECT`, so this condition rejects the surviving commit and the subject-only pass cannot recover it. Please trust a validated output hash, or parse the Git commit-output line narrowly enough that the hash is authoritative.

collector/internal/session/commits.go
```diff
@@ -1,12 +1,13 @@
...
 import (
+	"os/exec"
 	"regexp"
 	"strings"
 )
 
-// git commit as a complete command word, excludes plumbing commands (i.e.  commit-tree)
-var gitCommitCommand = regexp.MustCompile(`(?:^|[\n;&|])\s*(?:rtk\s+)?git\s+commit(?:$|[\s;&|])`)
+// git commit as complete command words anywhere in a shell segment; excludes commit-tree.
+var gitCommitCommand = regexp.MustCompile(`(?:^|[\n;&|])[^\n;&|]*?(\bgit\s+commit)(?:$|[\s;&|])`)
```
milanshen: **[P2] Avoid matching `git commit` inside quoted or data text.** This regex treats commands such as `echo "git commit -m example"` or searches through scripts as actual commit invocations. A successful search can therefore create an observation and surface an unrelated historical commit with the same subject. Please use shell-aware tokenization or constrain the allowed prefix without scanning arbitrary arguments and quoted text.

collector/internal/collector/collector.go
```diff
@@ -287,6 +287,9 @@ func probeGitEnvironment(roots []*vendors.ParsedSession) {
 	wg.Wait()
 	for _, p := range roots {
 		s := p.Session
+		s.Commits = session.ReconcileCommits(
```
milanshen: **[P2] Cache repository history per `(cwd, branch)`.** Every root reaches `repositoryHistory`, which runs a full `git log`; sessions sharing the same repository and branch therefore repeat the identical history scan serially on every list refresh. Please reuse history by the same key already used for the adjacent drift probe, or pass cached history into reconciliation.

collector/internal/session/commits.go
calvintvu: Fixed in a2301d1. If the command subject contains shell expansion, the parser records an unknown subject. Reconciliation then accepts the exact output hash.

collector/internal/collector/collector.go
calvintvu: Fixed in e96849c. Each list refresh now creates one in-memory reconciler. It reuses repository history for sessions with the same (cwd, branch) key. The cache ends after that refresh.

collector/internal/session/commits.go
calvintvu: Fixed in 5b89534. The parser masks quoted and escaped text before it matches commands. It now matches git commit only at a command position.

collector/internal/session/commits.go
calvintvu: Partially fixed in 4a10dce. A missing branch now returns an empty authoritative history. An unavailable repository still uses the transcript fallback. I did not add multi-ref reachability because reconciliation remains limited to the recorded branch.

milanshen (approved):

PR #80, opened 2026-08-17
Add OpenCode as a synthesis backend

milanshen:
## Context

Session debriefs could only run through the Claude Code or Codex CLIs. With
neither installed, synthesis could not be enabled at all — the Settings dialog
left it off because no backend reported as available. OpenCode was already a
first-class read source, so the gap was only on the invoke side.

## Changes

- **OpenCode as a third synthesis backend** — detected on PATH like the others.
  - Offers the free OpenCode Zen models, plus *Whichever model OpenCode is set to use*.
  - That list comes from `opencode models opencode --verbose`, cached for ten minutes
    and filtered to zero-cost models that OpenCode has not deprecated.
  - The probe runs from an empty directory with `--pure` and project config off.
    OpenCode loads project plugins from the working directory, so a probe inheriting
    the collector's cwd would run repository code on opening Settings. Global config
    is left intact, since enumerating models needs its providers.
  - It stays a probe rather than a read of the embedded price table: only 7 of the 27
    free models OpenCode has published are still live, that table carries no
    retirement signal, and it freezes at release while the free tier turns over in
    weeks.
  - Defaults to `opencode/deepseek-v4-flash-free`, run at its `high` reasoning variant.
  - That last option passes no `--model`, leaving OpenCode to resolve one itself:
    its configured model key, else the model last selected in the CLI.
  - Paid providers are reached that way rather than listed — OpenCode resolves them
    from ambient credentials, and the full list runs to hundreds of entries.
- **Compensating for flags OpenCode does not have**
  - No structured output → schema described in the prompt, reply parsed leniently,
    then checked for the fields the schema marks required. A partial object now
    fails into cooldown rather than being cached as a debrief.
  - No system-prompt flag → instructions ride along with the message.
  - No tool flags → tools denied through an injected config.
  - No ephemeral mode → each run gets its own scratch database under
    `~/.coslash/synthesis`, removed once the run ends, so runs never enter the
    user's own OpenCode history.
  - Per run rather than shared: OpenCode instances that overlap in time deadlock
    during init on a shared database, and the manager runs up to four at once.
  - A crash skips the cleanup, so directories older than an hour are swept at
    startup. The cutoff sits well past the 90s run timeout, so a second collector
    on `--port` cannot lose a live run's database to the sweep.
  - `--dir` passed explicitly, since OpenCode reads `PWD` before the process
    working directory.
- **Model validation relaxed to a shape check**, across all three backends.
  - The dropdown still offers the same curated options.
  - `settings.json` now accepts any model the chosen CLI can reach, including one
    behind an API proxy such as `ANTHROPIC_BASE_URL`.
  - Ids starting with `-` are rejected, so a value cannot be read as a CLI flag.
- **Each backend's default model runs at high reasoning effort.**
  - Codex: `-c model_reasoning_effort=high`; OpenCode: `--variant high`; Claude: `--effort high`.
  - Scoped to the default model on each backend, so an explicitly chosen model
    keeps the CLI's own default.
  - Note `--effort` works on `claude-haiku-4-5` through the CLI even though the
    Messages API rejects `output_config.effort` for that model.
- **Debrief outcomes are no longer truncated**, for every backend.
  - Previously capped at 200 characters with a trailing ellipsis; now only
    whitespace-collapsed. Goals, decisions, and next step keep their caps.

## Test

- Automated
  - `go build ./... && go vet ./... && go test ./...` — passed
  - Unit tests cover the OpenCode parse path: incomplete objects rejected
    (including `null` goals), bare, fenced, and prose-wrapped objects accepted.
  - Unit tests cover JSON framing (nesting, braces and escapes inside strings) and
    the scratch sweep (abandoned removed, in-flight and unrelated left alone).
  - `npm run build && npm test && npm run lint` — passed
- Manual
  - Real debriefs generated through all three backends: Claude, Codex, and OpenCode.
  - Reasoning-effort flags confirmed with logging shims on PATH — present for each
    backend's default model, absent for a non-default one.
  - Scratch database verified: after a full run, zero synthesis sessions in the
    user's OpenCode database and one in the scratch file. Verified before the
    database became per-run; concurrent runs not yet exercised by hand.
  - Probe isolation verified against OpenCode 1.18.18: a repo carrying a marker
    plugin in `.opencode/plugin` and `.opencode/plugins` executed it on the old
    invocation, and did not on the new one, which still returned the same 7 models.
  - Outcome verified untruncated end to end (512 and 685 characters, no ellipsis).
  - Validation matrix: curated and proxy-style ids accepted; flag-shaped and
    whitespace ids rejected.
  - Free-model list cross-checked against OpenCode's registry: the 7 models the CLI
    reports match filtering `models.opencode.ai/api.json` on zero cost and
    non-deprecated status, exactly.

### Screenshots

- [X] Settings dialog — synthesis on, three backend cards, OpenCode selected with the model dropdown open
Screenshot 2026-08-17 at 5 10 21 PM

- [X] Session inspector — a DEBRIEF card generated by the OpenCode backend
Screenshot 2026-08-17 at 5 11 57 PM

### DIFF
--- a/collector/internal/settings/settings.go
+++ b/collector/internal/settings/settings.go
@@ -31,6 +31,9 @@ const (
 	// The default Codex model, which the runner runs at high reasoning effort.
 	CodexSynthesisModel = "gpt-5.6-luna"
 
+	// The default Claude model, likewise run at high reasoning effort.
+	ClaudeSynthesisModel = "claude-haiku-4-5"
+
 	TerminalApple = "terminal"
 	TerminalITerm = "iterm2"
 )
--- a/collector/internal/settings/settings.go
+++ b/collector/internal/settings/settings.go
@@ -21,14 +21,11 @@ const (
 	BackendCodex    = "codex_exec"
 	BackendOpenCode = "opencode"
 
-	// OpenCodeDefaultModel passes no model, leaving OpenCode to resolve one:
-	// its configured model key, or failing that the model last selected in
-	// the CLI. That second case can change between runs.
+	// Passes no model, leaving OpenCode to resolve one: its configured model
+	// key, else the model last selected in the CLI, which varies between runs.
 	OpenCodeDefaultModel = "default"
 
-	// OpenCodeSynthesisModel is the free model coSlash prefers for OpenCode
-	// synthesis. It is the one free model with a reasoning variant worth
-	// turning up, which the runner does.
+	// The one free model with a reasoning variant, which the runner turns up.
 	OpenCodeSynthesisModel = "opencode/deepseek-v4-flash-free"
 
 	TerminalApple = "terminal"
@@ -137,8 +134,7 @@ func BackendOptions() []BackendOption {
 			},
 		},
 		{
-			// OpenCode models come from the user's own providers, so the
-			// caller fills these in from the CLI rather than a fixed list.
+			// Filled in from the CLI; OpenCode models depend on the user's providers.
 			ID:     BackendOpenCode,
 			Label:  "OpenCode CLI",
 			Models: []ModelOption{},
@@ -146,17 +142,11 @@ func BackendOptions() []BackendOption {
 	}
 }
 
-// modelPattern accepts bare ids like claude-haiku-4-5 and namespaced ones like
-// openrouter/anthropic/claude-haiku-4-5. The models a CLI can reach depend on
-// the user's own account, provider config, and any API proxy such as
-// ANTHROPIC_BASE_URL, so settings.json may name a model the picker does not
-// list. The leading character must not be "-", or the CLI would read the value
-// as another flag.
+// A leading "-" is excluded, or the CLI would read the value as another flag.
 var modelPattern = regexp.MustCompile(`^[A-Za-z0-9][A-Za-z0-9._:/-]*$`)
 
-// ValidSynthesisModel checks the shape of a model id, not membership of a
-// fixed list. BackendOptions drives the picker; this guards what the runner
-// will hand to the CLI.
+// ValidSynthesisModel checks shape, not membership of a fixed list: an API
+// proxy such as ANTHROPIC_BASE_URL can serve models the picker never lists.
 func ValidSynthesisModel(backend, model string) bool {
 	if backend == BackendOpenCode && model == OpenCodeDefaultModel {
 		return true
--- a/collector/internal/settings/settings.go
+++ b/collector/internal/settings/settings.go
@@ -150,10 +150,7 @@ var modelPattern = regexp.MustCompile(`^[A-Za-z0-9][A-Za-z0-9._:/-]*$`)
 
 // ValidSynthesisModel checks shape, not membership of a fixed list: an API
 // proxy such as ANTHROPIC_BASE_URL can serve models the picker never lists.
-func ValidSynthesisModel(backend, model string) bool {
-	if backend == BackendOpenCode && model == OpenCodeDefaultModel {
-		return true
-	}
+func ValidSynthesisModel(model string) bool {
 	return len(model) <= 200 && modelPattern.MatchString(model)
 }
 
@@ -194,7 +191,7 @@ func Validate(config Config) error {
 	if !known {
 		return fmt.Errorf("unsupported synthesis backend %q", config.Synthesis.Backend)
 	}
-	if !ValidSynthesisModel(config.Synthesis.Backend, config.Synthesis.Model) {
+	if !ValidSynthesisModel(config.Synthesis.Model) {
 		return fmt.Errorf("model %q is not a valid model id", config.Synthesis.Model)
 	}
 	if config.Appearance.Theme != "light" && config.Appearance.Theme != "dark" {
--- a/collector/internal/settings/settings.go
+++ b/collector/internal/settings/settings.go
@@ -9,6 +9,7 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
+	"regexp"
 	"sync"
 )
 
@@ -16,8 +17,19 @@ const (
 	SchemaURL = "https://raw.githubusercontent.com/centauri-ai/coslash/main/settings.schema.json"
 	Version   = 1
 
-	BackendClaude = "claude-cli"
-	BackendCodex  = "codex_exec"
+	BackendClaude   = "claude-cli"
+	BackendCodex    = "codex_exec"
+	BackendOpenCode = "opencode"
+
+	// OpenCodeDefaultModel passes no model, leaving OpenCode to resolve one:
+	// its configured model key, or failing that the model last selected in
+	// the CLI. That second case can change between runs.
+	OpenCodeDefaultModel = "default"
+
+	// OpenCodeSynthesisModel is the free model coSlash prefers for OpenCode
+	// synthesis. It is the one free model with a reasoning variant worth
+	// turning up, which the runner does.
+	OpenCodeSynthesisModel = "opencode/deepseek-v4-flash-free"
 
 	TerminalApple = "terminal"
 	TerminalITerm = "iterm2"
@@ -124,7 +136,26 @@ func BackendOptions() []BackendOption {
 				{ID: "gpt-5.6-sol", Label: "GPT-5.6 Sol"},
 			},
 		},
+		{
+			// OpenCode models come from the user's own providers, so the
+			// caller fills these in from the CLI rather than a fixed list.
+			ID:     BackendOpenCode,
+			Label:  "OpenCode CLI",
+			Models: []ModelOption{},
+		},
 	}
+}
+
+// openCodeModelPattern matches "provider/model". Routed providers add
+// segments, as in openrouter/anthropic/claude-haiku-4-5. It stays permissive
+// so any provider the user types remains valid.
+var openCodeModelPattern = regexp.MustCompile(`^[A-Za-z0-9._:-]+(/[A-Za-z0-9._:-]+)+$`)
+
+func ValidOpenCodeModel(model string) bool {
+	if model == OpenCodeDefaultModel {
+		return true
+	}
+	return len(model) <= 200 && openCodeModelPattern.MatchString(model)
 }
 
 func TerminalOptions() []TerminalOption {
@@ -140,6 +171,8 @@ func BackendExecutable(backend string) string {
 		return "claude"
 	case BackendCodex:
 		return "codex"
+	case BackendOpenCode:
+		return "opencode"
 	default:
 		return ""
 	}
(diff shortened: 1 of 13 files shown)


collector/internal/synthesis/opencode.go
```diff
@@ -0,0 +1,135 @@
...
+)
+
+const openCodeConfigContent = `{"permission":"deny","autoupdate":false}`
+
+func openCodeEnv() []string {
+	return []string{
+		"OPENCODE_CONFIG_CONTENT=" + openCodeConfigContent,
+		// OpenCode has no ephemeral mode, so its runs go to a scratch database
+		// rather than the one holding the user's own sessions.
+		"OPENCODE_DB=" + filepath.Join(SynthesisCwd(), "opencode.db"),
```
calvintvu: [P1] Avoid sharing this database across concurrent synthesis runs.

The manager permits four Runner.Run calls at once, and the initial sweep can schedule many sessions, so every OpenCode process opens this same SQLite file. OpenCode versions accepted by the current LookPath-only probe have a documented concurrent-initialization deadlock when they share OPENCODE_DB; affected runs wait for the 90-second timeout and then enter cooldown. Please serialize OpenCode runs or use and clean up a per-invocation database.

https://github.com/anomalyco/opencode/issues/29395

collector/internal/synthesis/opencode.go
```diff
@@ -0,0 +1,135 @@
...
+		if text := strings.TrimSpace(texts[key]); text != "" {
+			parts = append(parts, text)
+		}
+	}
+	return strings.Join(parts, "\n"), nil
+}
+
+// OpenCode cannot enforce an output schema, so prose around the object is tolerated.
+func parseOpenCodeSynthesis(text string) (session.SessionSynthesis, error) {
+	synthesis, err := parseSynthesis([]byte(text))
```
calvintvu: [P2] Validate the required synthesis fields after parsing.

OpenCode only receives the schema as prompt text. parseSynthesis unmarshals into a zero-valued struct, and normalize rejects only when both goals and outcome are empty, so a response such as {"outcome":"done"} succeeds even though goals, keyDecisions, and nextStep are absent. That incomplete result is then cached instead of falling back. Please validate required-field presence and the goals minimum on this unenforced-schema path.

collector/internal/settings/settings.go
```diff
@@ -9,15 +9,30 @@ import (
...
+	OpenCodeDefaultModel = "default"
+
+	// The one free model with a reasoning variant, which the runner turns up.
+	OpenCodeSynthesisModel = "opencode/deepseek-v4-flash-free"
+
+	// The default Codex model, which the runner runs at high reasoning effort.
+	CodexSynthesisModel = "gpt-5.6-luna"
+
+	// The default Claude model, likewise run at high reasoning effort.
+	ClaudeSynthesisModel = "claude-haiku-4-5"
```
calvintvu: may become outdated upon new model releases, but only might become an issue if model is deprecated

collector/internal/synthesis/opencode.go
milanshen: Fixed in 49b88ca — each run now gets its own database, so no two OpenCode processes share one.

Went with the per-invocation database rather than serializing: a sweep can schedule 20 sessions, and at up to 90s each that would stretch toward the 30-minute sweep interval. `openCodeScratchDir` mints a temp dir under `~/.coslash/synthesis`, `OPENCODE_DB` points inside it, and `defer os.RemoveAll` clears it — including on the timeout path.

Worth noting the linked issue is worse than version-gating would fix: it was auto-closed as `not planned`, and the fix PR (#31461) was closed unmerged. So there is no good version to probe for.

`--dir` and `PWD` still point at the synthesis cwd, so the `servableRoots` filter that hides synthesis sessions is unaffected.

collector/internal/synthesis/opencode.go
milanshen: Fixed in 49b88ca. Confirmed the case you named: `{"outcome":"done"}` did parse and get cached, since `normalize` only rejects when goals and outcome are both empty.

Added `requireSynthesisFields` on the OpenCode path — presence of all four schema-required keys, plus the `minItems: 1` goals minimum. An incomplete object now falls through to cooldown instead of caching. Claude and Codex are untouched and keep relying on their enforced schemas.

Added `opencode_test.go` covering it: incomplete objects rejected (including `null` goals), and bare, fenced, and prose-wrapped objects still accepted.

collector/internal/settings/settings.go
milanshen: Agreed, and it applies to all three constants equally. Leaving them pinned: each one also gates the reasoning-effort flag, so it has to name a specific model rather than track whatever the CLI currently defaults to. A deprecation shows up as a run failure with the CLI-auth-or-model message, and since validation is now a shape check, the fix is a one-line constant bump or a user-set model in `settings.json`.

collector/internal/vendors/opencode/models.go
```diff
@@ -0,0 +1,103 @@
...
+			Text bool `json:"text"`
+		} `json:"output"`
+	} `json:"capabilities"`
+	Status string `json:"status"`
+}
+
+func loadModels() []string {
+	ctx, cancel := context.WithTimeout(context.Background(), modelTimeout)
+	defer cancel()
+	output, err := exec.CommandContext(ctx, "opencode", "models", "opencode", "--verbose").Output()
```
calvintvu: [P1] Isolate the model probe from project plugins

writeSettings reaches this probe whenever Settings is opened. Unlike the synthesis run, it inherits the collector cwd and environment and does not pass --pure or OPENCODE_DISABLE_PROJECT_CONFIG. OpenCode automatically loads project .opencode/plugins at startup, so starting coSlash in an untrusted checkout and merely opening Settings can execute repository plugin code. Please run this probe from a controlled directory with external plugins and project config disabled while retaining the global provider config needed to enumerate models.

collector/internal/synthesis/runner.go
```diff
@@ -104,8 +115,40 @@ func (r *CLIRunner) Run(ctx context.Context, input string) (session.SessionSynth
...
+			if err != nil {
+				return session.SessionSynthesis{}, err
+			}
+			return parseOpenCodeSynthesis(text)
+		}
+		scratchDir, err := openCodeScratchDir()
+		if err != nil {
+			return session.SessionSynthesis{}, err
+		}
+		defer os.RemoveAll(scratchDir)
```
calvintvu, line 134: [P2] Sweep abandoned OpenCode scratch databases

This defer cleans up normal returns and timeout paths, but a collector crash or SIGKILL bypasses it. The per-run OpenCode database contains the synthesis request and response, and there is no later sweep, so an interrupted run leaves that data under synthesis indefinitely. Please remove stale .opencode-* directories on startup or before a later run. The 0700 parent limits cross-user exposure but does not address retention.

collector/internal/synthesis/opencode.go
```diff
@@ -0,0 +1,174 @@
...
+		}
+	}
+	var goals []string
+	if err := json.Unmarshal(fields["goals"], &goals); err != nil || len(goals) == 0 {
+		return errors.New("synthesis result has no goals")
+	}
+	return nil
+}
+
+func extractJSONObject(value string) string {
```
calvintvu, line 199: [Optional][complexity] Replace the hand-written JSON framer with encoding/json

After finding the first opening brace, json.Decoder.Decode into json.RawMessage already handles nested objects, strings, and escapes and stops after one JSON value. That can replace most of extractJSONObject. This is an optional simplification; the current state machine is not a merge blocker.

collector/internal/synthesis/runner.go
```diff
@@ -56,6 +61,8 @@ func NewRunner(config settings.SynthesisSettings) (Runner, error) {
 		runner.Bin = "claude"
 	case settings.BackendCodex:
 		runner.Bin = "codex"
+	case settings.BackendOpenCode:
```
calvintvu: [Optional][complexity] Reuse settings.BackendExecutable here

NewRunner repeats the backend-to-binary mapping already owned by settings.BackendExecutable. Reading that helper once and rejecting an empty result removes this switch and prevents the two mappings from drifting. Optional cleanup only.

collector/internal/settings/settings.go
```diff
@@ -124,9 +136,27 @@ func BackendOptions() []BackendOption {
...
+		},
 	}
 }
 
+// A leading "-" is excluded, or the CLI would read the value as another flag.
+var modelPattern = regexp.MustCompile(`^[A-Za-z0-9][A-Za-z0-9._:/-]*$`)
+
+// ValidSynthesisModel checks shape, not membership of a fixed list: an API
+// proxy such as ANTHROPIC_BASE_URL can serve models the picker never lists.
+func ValidSynthesisModel(backend, model string) bool {
```
calvintvu: [Optional][complexity] Drop the redundant backend special case

OpenCodeDefaultModel is default, which already passes modelPattern, so the branch does not change the result. ValidSynthesisModel can take only model and return the length and pattern check; update the two call sites accordingly. Optional cleanup only.

collector/internal/vendors/opencode/models.go
milanshen: Fixed in 8a19fa2. Confirmed against OpenCode 1.18.18 before changing anything: a repo with a marker plugin in `.opencode/plugin` and `.opencode/plugins` executed both on the exact invocation the probe used, while still listing the 7 models — so this was live, not theoretical.

The probe now runs from an empty `os.MkdirTemp` directory with `--pure`, `OPENCODE_DISABLE_PROJECT_CONFIG=1`, and `PWD` pointed at that directory, since OpenCode reads `PWD` before cwd and `exec.Cmd` does not update it.

Global config is deliberately untouched — no `OPENCODE_CONFIG_CONTENT` — so provider resolution still works, per your point about needing it to enumerate. Re-ran the fixed `loadModels` with cwd set to that repo: no marker written, same 7 models returned.

collector/internal/synthesis/runner.go
milanshen: Fair — that retention gap came in with the per-run change. Fixed in 98198d2.

`CleanupOpenCodeScratch` sweeps `.opencode-*` directories at startup, mirroring the existing `launch.CleanupHandoffs` age-based sweep. Cutoff is an hour, well past the 90s run timeout: there is no single-instance guard, so a second collector started on `--port` can have runs in flight, and a blanket sweep would pull the database out from under one.

Tested: abandoned directory removed, in-flight and unrelated directories left alone.

collector/internal/synthesis/opencode.go
milanshen: Taken in 98198d2 — `extractJSONObject` is now the first-brace scan plus `json.Decoder.Decode` into a `json.RawMessage`, down from 28 lines to 9.

Added tests first to pin the equivalence, since the state machine handled cases worth not regressing: nested objects, braces inside strings, escaped quotes before a brace, stopping after the first value, and unterminated input.

collector/internal/synthesis/runner.go
milanshen: Taken in d519e49 — `NewRunner` now reads `settings.BackendExecutable` and rejects an empty result, so the mapping lives in one place.

collector/internal/settings/settings.go
milanshen: Confirmed and taken in d519e49. `default` does pass `modelPattern` — `d` matches the leading class and the rest are word characters — so the branch was dead. `ValidSynthesisModel(model string)` now does only the length and pattern check, with both call sites updated.

collector/internal/synthesis/opencode.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+	return parseSynthesis([]byte(text))
+}
+
+func requireSynthesisFields(data []byte) error {
+	var fields map[string]json.RawMessage
+	if err := json.Unmarshal(data, &fields); err != nil {
+		return fmt.Errorf("decode synthesis result: %w", err)
+	}
+	for _, name := range []string{"goals", "outcome", "keyDecisions", "nextStep"} {
```
calvintvu: **[P2]** Validate required values, not only key presence

This loop proves only that the keys exist, and the later check validates only the pre-normalized goals length. encoding/json accepts null for string and slice fields and accepts null array elements as empty strings. As a result, `{"goals":["ship"],"outcome":null,"keyDecisions":null,"nextStep":null}` succeeds and is cached; keyDecisions then serializes as null while SessionInspector and handoffBrief call `.length` on it. `{"goals":[null],"outcome":"done","keyDecisions":[],"nextStep":"ship"}` also succeeds before the null goal normalizes away. Please reject null required values and null array elements before parsing and caching.

calvintvu (approved):

collector/internal/synthesis/opencode.go
milanshen: Fixed in ad60fd2 — OpenCode synthesis validation now rejects null required values and null array elements before parsing and caching.

PR #84, opened 2026-08-17
feat(snapshot): define session snapshot v1 contract

luhe19001:
## Summary

Define the public `session-snapshot/v1` wire contract, canonical JSON encoding, validation, hashing, and JSON Schema.

## Why this is separate

Reviewers can evaluate the protocol without also auditing private collector mapping, redaction, or payload fitting. The package has no dependency on collector internals.

## Changes

- add explicit v1 envelope and session wire types;
- freeze canonical JSON member order and SHA-256 content hashing;
- enforce the 256 KiB aggregate ceiling and per-field byte/item limits;
- reject unknown fields, non-canonical bytes, unsafe relative paths, invalid metadata pointers, negative counters, and inverted timestamps;
- publish a matching JSON Schema and boundary tests.

## Review guide

Read `snapshot.go` and `README.md` first, compare `schema.json` with the Go validator, then review the exact-limit tests. This PR does not map or export local sessions.

## Verification

- `go test ./...`
- `go vet ./...`
- `jq empty collector/snapshot/v1/schema.json`

## Stack

This is PR 2 of 5 for ENG-1340. Its base is PR 1.

| Order | PR | Scope |
| ---: | --- | --- |

## Rollout

This only publishes a new versioned contract package; no caller uses it until the next PR.

### DIFF
--- a/collector/snapshot/v1/snapshot.go
+++ b/collector/snapshot/v1/snapshot.go
@@ -618,8 +618,16 @@ func jsonPointerResolves(value any, pointer string) bool {
 				return false
 			}
 		case []any:
+			if segment == "" || len(segment) > 1 && segment[0] == '0' {
+				return false
+			}
+			for i := range len(segment) {
+				if segment[i] < '0' || segment[i] > '9' {
+					return false
+				}
+			}
 			index, err := strconv.Atoi(segment)
-			if err != nil || index < 0 || index >= len(current) {
+			if err != nil || index >= len(current) {
 				return false
 			}
 			value = current[index]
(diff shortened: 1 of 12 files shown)


collector/snapshot/v1/snapshot.go
```diff
@@ -0,0 +1,662 @@
...
+			if err != nil || index < 0 || index >= len(current) {
+				return false
+			}
+			value = current[index]
+		default:
+			return false
+		}
+	}
+	return true
+}
```
calvintvu, line 639: jsonPointerResolves accepts array indices with leading zeros because `strconv.Atoi("01") == 1`. Thus `/session/digest/01` passes validation, although [RFC 6901 forbids leading zeros](https://www.rfc-editor.org/rfc/rfc6901.html#section-4)

luhe19001: Addressed in de41cfc. Array reference tokens now require canonical unsigned decimal syntax before conversion, so leading-zero indices such as `01` (and signed forms such as `+1`) are rejected while `0` and `1` remain valid. Added regression coverage and verified with `go test ./...`.

calvintvu (approved):

PR #85, opened 2026-08-17
feat(collector): add explicit session export boundary

luhe19001:
## Summary

Add the explicit private-to-wire allow-list for session snapshots, including path proof, credential redaction, text/item budgets, and frozen usage/cost mapping.

## Why this is separate

This is the privacy boundary. Keeping it apart from the public schema and aggregate fitting lets reviewers focus on exactly which local fields may leave the machine and how each one is sanitized.

## Changes

- map approved session, usage, digest, todo, edit, commit, Git, and subagent facts;
- exclude raw commands, transcripts, tool output, diffs, reasoning, log paths, and parser state;
- redact quoted and unquoted credential patterns without leaking quoted suffixes;
- export file paths only after proving they are repository-relative;
- normalize and safely truncate UTF-8 text and paths;
- transform, drop, merge, sort, and deduplicate model names before validation;
- add a nested allow-list census so new local fields remain excluded by default.

## Review guide

Review `Build` field by field, then the credential/path helpers, followed by `allowlist_test.go`. `Marshal` still delegates directly to the public contract in this slice; aggregate degradation arrives next.

## Verification

- `go test ./...`
- `go vet ./...`

## Stack

This is PR 3 of 5 for ENG-1340. Its base is PR 2.

| Order | PR | Scope |
| ---: | --- | --- |

## Rollout

The mapper is internal and has no preview/upload caller in this stack, so it remains inert until integration is added.

### DIFF
--- a/collector/internal/session/probes.go
+++ b/collector/internal/session/probes.go
@@ -48,7 +48,7 @@ func CanonicalRepositoryName(cwd string) (string, bool) {
 	if err != nil {
 		return fallback, true
 	}
-	root := repositoryRoot(resolved)
+	root := RepositoryRoot(resolved)
 	if root == "" {
 		return fallback, true
 	}
@@ -114,7 +114,8 @@ func canonicalRemoteName(remote string) string {
 	return strings.ToLower(parsed.Host) + "/" + path
 }
 
-func repositoryRoot(cwd string) string {
+// RepositoryRoot returns the enclosing Git worktree root when one exists.
+func RepositoryRoot(cwd string) string {
 	info, err := os.Stat(cwd)
 	if err != nil || !info.IsDir() {
 		return ""
--- a/collector/internal/sessionexport/allowlist_test.go
+++ b/collector/internal/sessionexport/allowlist_test.go
@@ -1,0 +1,163 @@
+package sessionexport
+
+import (
+	"reflect"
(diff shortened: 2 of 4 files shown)


calvintvu (commented):
up to author, rename `sessionexport` to just `export` for readability

calvintvu:
**[P1]** – Incomplete credentials leak secrets. The [regexes](https://github.com/centauri-ai/coslash/blob/30d7b62d844e8efed30ad2a8974b9b2aa6c24fe2/collector/internal/sessionexport/snapshot.go#L144-L154) require closing quotes/PEM markers. Truncated password="first-secret second-secret leaks second-secret; an incomplete private-key block leaks its body. Redact through end-of-string when delimiters are missing.

luhe19001:
Addressed the P1 incomplete-credential leak in 3bc080b. Unterminated JSON and shell-style quoted credentials, bearer values, and private-key blocks now redact through end-of-text, including dangling escape cases. Added end-to-end export regression coverage and verified with `go test ./...`.

I kept the `sessionexport` package name: this package is specifically the trust boundary from the private session model to the public snapshot contract, and `export` would be less descriptive and easier to confuse with future export surfaces.

calvintvu (approved):

PR #86, opened 2026-08-17
feat(collector): fit snapshots to aggregate budget

luhe19001:
## Summary

Fit oversized snapshots to the 256 KiB contract ceiling by deterministically reducing optional evidence while preserving the mandatory core.

## Why this is separate

Aggregate fitting is an algorithmic layer with its own ordering and metadata invariants. Reviewers can validate those rules without re-reading the schema or privacy mapper.

## Changes

- reduce command labels, subagent prose, digest answers, older digest entries, todos, older commits, and older edits in a fixed order;
- retain newest digest, commit, and edit evidence;
- merge per-field and aggregate budget records instead of emitting contradictory duplicates;
- delete or reindex metadata pointers whenever evidence is removed;
- use exact-length single-marshal probes during binary search, with authoritative validation and hashing once at the end;
- preserve UTF-8 boundaries without allocating a rune-index table;
- fail clearly only when the mandatory snapshot core cannot fit.

## Review guide

Start with the reducer order in `fitAggregate`, then inspect the shared prefix/newest reducers and metadata filtering. The tests exercise every degradation stage and pointer invariant.

## Verification

- `go test ./...`
- `go vet ./...`

## Stack

This is PR 4 of 5 for ENG-1340. Its base is PR 3.

| Order | PR | Scope |
| ---: | --- | --- |

## Rollout

Fitting is deterministic and only runs when a built snapshot exceeds the public limit. The original monolithic branch remains available as a rollback reference.

### DIFF
--- a/collector/internal/sessionexport/aggregate.go
+++ b/collector/internal/sessionexport/aggregate.go
@@ -27,6 +27,7 @@ func fitAggregate(snapshot snapshotv1.Snapshot) (snapshotv1.Snapshot, error) {
 		reduceTodos,
 		reduceCommits,
 		reduceFileEdits,
+		reduceSessionMetadata,
 	}
 	for _, reduce := range reducers {
 		if fits, err := reduce(&snapshot); err != nil {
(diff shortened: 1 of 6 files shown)


collector/internal/sessionexport/aggregate.go
```diff
@@ -0,0 +1,426 @@
...
+
+	reducers := []func(*snapshotv1.Snapshot) (bool, error){
+		reduceCommandLabels,
+		reduceSubagentProse,
+		reduceDigestAnswers,
+		reduceOlderDigest,
+		reduceTodos,
+		reduceCommits,
+		reduceFileEdits,
+	}
```
calvintvu, line 31: Optional session metadata can make this return ErrOversized even when the documented mandatory core fits. I reproduced a 272,019-byte built snapshot where clearing only summary, declaredGoal, and firstPrompt reduced it to 235,107 bytes, but none of those fields has a reducer, so fitAggregate reports that mandatory facts exceed the limit. Please add a final reduction stage for non-core metadata (and record it) before treating the remainder as mandatory.

collector/internal/sessionexport/aggregate.go
```diff
@@ -0,0 +1,426 @@
...
+func reduceTodos(snapshot *snapshotv1.Snapshot) (bool, error) {
+	return reduceTail(snapshot, "/session/todos", &snapshot.Session.Todos)
+}
+
+func reduceCommits(snapshot *snapshotv1.Snapshot) (bool, error) {
+	return reduceNewest(snapshot, "/session/commits", &snapshot.Session.Commits)
+}
+
+func reduceFileEdits(snapshot *snapshotv1.Snapshot) (bool, error) {
+	return reduceNewest(snapshot, "/session/fileEdits", &snapshot.Session.FileEdits)
```
calvintvu, line 139: reduceNewest treats slice position as edit recency, but FileEditSet.Add updates an existing file in place at its first-seen index. In a 301-file repro, hot.go was created first and edited again last; fitting removed it as old. Please preserve actual last-edit order (or track recency) before taking a suffix.

calvintvu (commented):
Bug-only review: two reproducible failure cases.

collector/internal/sessionexport/aggregate.go
luhe19001: Addressed in 590c0d5. Added a final deterministic reducer for every optional session metadata field before mandatory-core failure. Removed properties also drop stale child metadata, and the reduction is recorded at the still-resolving `/session` pointer with original/exported field counts. Added a near-limit regression where the mandatory usage core fits but summary, declaredGoal, and firstPrompt push the snapshot over the ceiling.

collector/internal/sessionexport/aggregate.go
luhe19001: Addressed in 590c0d5. FileEditSet now maintains oldest-to-newest last-touch order by moving an updated file to the tail, so aggregate suffix retention preserves genuinely recent files. The earlier 2,000-item export cap now also takes the newest valid suffix. Added unit coverage for repeated index updates and an end-to-end 301-file fitting repro that retains hot.go after it is edited last.

calvintvu (approved):

PR #91, opened 2026-08-17
Add session sorting by status

milanshen:
## Context

Session lists can be sorted by recency, estimated cost, token usage, and duration, but not by current activity. A shared status priority makes active and attention-needed work easier to find while keeping list sorting and board columns consistent.

## Changes

- Add Status to the session sort menu, ordered Active → Waiting → Idle → Inactive when descending
- Apply the same status order to board columns
- Preserve the direction control so ascending reverses the list order
- Treat sessions without a live status as inactive

## Test

- `npm test` — passed (11 tests)
- `npm run build` — passed
- Manual: Not run

### Screenshots

Screenshot 2026-08-17 at 4 01 31 PM

### DIFF
--- a/frontend/src/pages/coslash/lib/session.ts
+++ b/frontend/src/pages/coslash/lib/session.ts
@@ -183,6 +183,9 @@ export const STATUSES = {
 
 export type StatusKey = keyof typeof STATUSES;
 
+// Board columns render left to right in this order; status sorting uses the same priority.
+export const STATUS_ORDER: readonly StatusKey[] = ['busy', 'waiting', 'idle', 'inactive'];
+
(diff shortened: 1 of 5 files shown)


calvintvu (approved):

frontend/src/pages/coslash/lib/session.ts
```diff
@@ -182,6 +182,8 @@ export const STATUSES = {
 
 export type StatusKey = keyof typeof STATUSES;
 
+export const STATUS_ORDER: readonly StatusKey[] = ['busy', 'waiting', 'idle', 'inactive'];
```
calvintvu, line 187: up to author, clarify status order goes left to right

PR #93, opened 2026-08-17
Hide undetected vendors from filters and synthesis settings

milanshen:
## Context

The filter bar currently lists every supported vendor even when the loaded session set has no matching sessions. Synthesis settings similarly show backends whose CLI is unavailable. Limit both controls to relevant, usable choices.

## Changes

- Show vendor filter tabs only for vendors represented in loaded sessions, resetting a stale selection to All vendors
- Hide unavailable synthesis backends and their model choices

## Test

- `cd frontend && npm test` — passed
- `cd frontend && npm run build` — passed
- `cd frontend && npm run lint` — passed with two existing Fast Refresh warnings
- Manual UI verification not run; a clean-machine state was not available locally

### Screenshots

- [ ] Vendor filter — only vendors with loaded sessions are shown
- [ ] Synthesis settings — unavailable CLI backends are omitted

### DIFF
--- a/frontend/src/pages/coslash/CoslashPage.tsx
+++ b/frontend/src/pages/coslash/CoslashPage.tsx
@@ -37,7 +37,7 @@ import type { Diagnostics } from '@/pages/coslash/lib/diagnostics';
 import { formatEstimatedCost } from '@/pages/coslash/lib/format';
 import { sessionsEmptyStateCopy } from '@/pages/coslash/lib/page-copy';
 import { sessionMatchesSearchTerm } from '@/pages/coslash/lib/search';
-import { getStatus, type Session } from '@/pages/coslash/lib/session';
+import { getSessionVendors, getStatus, type Session } from '@/pages/coslash/lib/session';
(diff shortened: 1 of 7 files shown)


calvintvu:
**P2 — Handle an empty backend list**

`availableSynthesisBackends()` returns an empty array when no supported CLI is installed.

The current code does not cause a runtime error. It checks the first array item and uses optional access for the selected backend and model.

If synthesis is enabled with an empty array, the dialog shows an empty backend list. The model selector has no valid options. The Save button is disabled.

If the array is empty, show `No supported CLI detected`. In this state, hide the model selector. Add a test with no available backends.

Contract summary:

- `response` and `draft` can be `null` while the UI loads. The code checks both values.
- The `response.options` object and raw backend array always exist in the current server response.
- `availableSynthesisBackends()` always returns an array. This array can be empty.
- The first backend, selected backend, and selected model can be `undefined`. The code checks these values.
- Use `.at(0)` to show the `BackendOption | undefined` type. The project does not enable `noUncheckedIndexedAccess`.

calvintvu (approved):

PR #96, opened 2026-08-18
Clear stale OpenCode waiting status on newer turns

calvintvu:
## Context

OpenCode can retain an abandoned question tool as running after a session continues. Scanning that historical state caused completed resumed sessions to remain shown as waiting.

## Changes

- Clear question-derived waiting state when a newer non-synthetic user prompt begins.
- Keep unanswered questions in the current turn waiting and leave busy status handling unchanged.

## Test

- `go test $(git ls-files 'internal/vendors/opencode/*.go')` — passed

### Screenshots
- [ ] Session board — resumed session with a historical question no longer shown as Waiting

### DIFF
--- a/collector/internal/vendors/opencode/parse.go
+++ b/collector/internal/vendors/opencode/parse.go
@@ -82,6 +82,7 @@ func parse(tx *sql.Tx, row storedSession) (parsedSession, error) {
 			if prompt == "" {
 				continue
 			}
+			waiting = false
 			turns++
 			if firstPrompt == "" {
 				firstPrompt = prompt


milanshen (approved):

collector/internal/vendors/opencode/parse.go
```diff
@@ -82,6 +82,7 @@ func parse(tx *sql.Tx, row storedSession) (parsedSession, error) {
 			if prompt == "" {
 				continue
 			}
+			waiting = false
```
milanshen, line 85: [P2] Exposing the pre-existing stale busy state (non-blocking: probably a follow up PR)

`busy` can already be latched by a historical assistant message without `time.completed`. This PR does not introduce that underlying bug, but clearing only `waiting` here causes affected sessions to fall through to `busy`/Active after a newer prompt, even when no OpenCode TUI is live. I reproduced this with an incomplete assistant/running question followed by a newer user prompt and completed assistant reply. Please either fix the stale-`busy` handling first or clear/recompute it as part of this transition, with a focused regression test.

PR #97, opened 2026-08-18
feat: add exact snapshot preview

luhe19001:
## What this changes

This PR adds an exact, local preview of the session snapshot that a future Team sharing flow will upload. The preview is created before any data leaves the machine and from the selected saved session revision.

| This PR introduces | This prevents |
| --- | --- |
| A pre-upload preview of the exact selected session snapshot | Sending session data to a server before the user can review it |
| One canonical serialization path whose unchanged bytes are displayed by the preview and can be passed to the future uploader | A separate preview payload builder drifting from the data that is actually uploaded |
| A controlled preview dialog with an explicit development-only test entry point | Showing Team sharing controls to users of the local-only application |

## What users see

The normal local-only experience is unchanged: no sharing button is shown.

When the development test entry point is enabled, a session has a **Team sharing preview** action. The dialog is labeled **PREVIEW ONLY · TEAM FEATURE** and shows:

- the canonical JSON snapshot and its byte size and content hash;
- repository identity, frozen cost, and any unpriced models;
- fields that were redacted or truncated, including known credential patterns; and
- data that is structurally excluded, including raw transcripts, assistant reasoning, tool output, file diffs, commands, environment variables, and unresolved local paths.

Other sensitive text can remain in the bounded `firstPrompt`, so the dialog shows its exact value for review rather than claiming that every possible credential format is removed.

The preview blocks approval when the snapshot is invalid, uses an unsupported format, no longer matches the selected revision, or exceeds the 256 KiB payload limit. This PR does not provide an approval or upload action.

## Implementation boundary

The local collector owns this preview because both the session source and the existing `session-snapshot/v1` serializer are local. The collector serializes the snapshot once, and the preview decodes those same bytes for display. A future Team sharing flow must upload those unchanged, already-reviewed bytes.

This PR includes:

- the local `/api/share-preview` endpoint for one session ID and revision;
- the versioned `snapshot-preview/v1` response and its fail-closed states;
- the controlled React preview dialog and byte-parity checks; and
- privacy documentation and public/server fixture conformance.

This PR does not include Team eligibility, authentication, destination selection, multi-session selection, consent submission, upload, retry, or success navigation. Those belong to the production Team sharing flow that will consume this preview.

## Try the preview

1. From `collector/`, run `make run`.
2. In another terminal, from `frontend/`, run `npm run dev`.
3. Open `http://127.0.0.1:5173/?team-preview=1`.
4. Select a session and click **Team sharing preview**.

Remove `?team-preview=1` to confirm that the local-only UI contains no sharing action.

Exact Team sharing preview showing the canonical local session snapshot

## Verification

- `go test ./...`
- `go vet ./...`
- 19 frontend tests
- frontend build, lint, and format checks
- snapshot fixture validation against coSlash Server commit `c0fabc8`

## Follow-up requirement

The production Team sharing flow must open this controlled preview only after the user enters a sharing flow and selects what to share. It must upload the exact canonical bytes the user reviewed, rather than rebuilding the snapshot.

### DIFF
--- a/collector/internal/collector/collector.go
+++ b/collector/internal/collector/collector.go
@@ -45,6 +45,11 @@ var vendorSources = []vendorSource{
 	},
 }
 
+var previewSessionCache = struct {
+	sync.RWMutex
+	byID map[string]session.Session
+}{byID: map[string]session.Session{}}
+
 type SourceHealth = vendors.SourceHealth
 
 func Sources() []SourceHealth {
@@ -83,7 +88,46 @@ func List(since int64) ([]*session.Session, error) {
 	for _, root := range roots {
 		sessions = append(sessions, root.Session)
 	}
+	cachePreviewSessions(sessions)
 	return sessions, nil
+}
+
+// GetSessionForPreview returns the selected fully composed revision.
+func GetSessionForPreview(id string, expectedRevision int64) (*session.Session, error) {
+	if id == "" {
+		return nil, nil
+	}
+	current, err := GetSessionFacts(id)
+	if err != nil || current == nil {
+		return current, err
+	}
+	if current.LastActivityTime != expectedRevision {
+		return current, nil
+	}
+	previewSessionCache.RLock()
+	cached, ok := previewSessionCache.byID[id]
+	previewSessionCache.RUnlock()
+	if ok && cached.LastActivityTime == current.LastActivityTime {
+		return &cached, nil
+	}
+	sessions, err := List(0)
+	if err != nil {
+		return nil, err
+	}
+	for _, candidate := range sessions {
+		if candidate.ID == id {
+			return candidate, nil
+		}
+	}
+	return nil, nil
+}
+
+func cachePreviewSessions(sessions []*session.Session) {
+	previewSessionCache.Lock()
+	defer previewSessionCache.Unlock()
+	for _, value := range sessions {
+		previewSessionCache.byID[value.ID] = *value
+	}
 }
 
 func applyActivityFallbacks(parsed []*vendors.ParsedSession) {
--- a/collector/internal/collector/collector.go
+++ b/collector/internal/collector/collector.go
@@ -45,11 +45,6 @@ var vendorSources = []vendorSource{
 	},
 }
 
-var previewSessionCache = struct {
-	sync.RWMutex
-	byID map[string]session.Session
-}{byID: map[string]session.Session{}}
-
 type SourceHealth = vendors.SourceHealth
 
 func Sources() []SourceHealth {
@@ -88,7 +83,6 @@ func List(since int64) ([]*session.Session, error) {
 	for _, root := range roots {
 		sessions = append(sessions, root.Session)
 	}
-	cachePreviewSessions(sessions)
 	return sessions, nil
 }
 
@@ -104,13 +98,7 @@ func GetSessionForPreview(id string, expectedRevision int64) (*session.Session, 
 	if current.LastActivityTime != expectedRevision {
 		return current, nil
 	}
-	previewSessionCache.RLock()
-	cached, ok := previewSessionCache.byID[id]
-	previewSessionCache.RUnlock()
-	if ok && cached.LastActivityTime == current.LastActivityTime {
-		return &cached, nil
-	}
-	sessions, err := List(0)
+	sessions, err := List(expectedRevision)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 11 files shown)


frontend/src/pages/coslash/lib/preview.ts
```diff
@@ -0,0 +1,140 @@
...
+  const binary = atob(preview.canonicalPayloadBase64);
+  const bytes = Uint8Array.from(binary, (character) => character.charCodeAt(0));
+  if (bytes.byteLength !== preview.payloadBytes) {
+    throw new Error('Canonical preview payload length does not match its metadata.');
+  }
+  const text = new TextDecoder('utf-8', { fatal: true }).decode(bytes);
+  const decoded = JSON.parse(text) as { schemaVersion?: unknown; contentHash?: unknown };
+  if (
+    decoded.schemaVersion !== preview.schemaVersion ||
+    decoded.contentHash !== preview.snapshot.contentHash
```
calvintvu: The byte comparison does not compare the complete snapshot. It compares only `schemaVersion` and `contentHash`. A changed repository or cost can pass this comparison.

Compare the complete decoded object with `preview.snapshot`. Return an error when any field differs.

frontend/src/pages/coslash/lib/preview.ts
```diff
@@ -0,0 +1,140 @@
...
+};
+
+export const PREVIEW_PRIVACY_COPY =
+  'When present, firstPrompt is shared up to 16 KiB after credential-pattern redaction. Prompt data can therefore be included; this preview shows the exact bounded value.';
+
+export const STRUCTURALLY_EXCLUDED = [
+  'Raw transcripts and assistant reasoning',
+  'Tool output and file diffs',
+  'Raw top-level and subagent commands',
+  'Credentials, environment variables, and unresolved local paths',
```
calvintvu: The preview says that credentials are structurally excluded. The serializer only removes credentials that match its regular expressions. Other credential formats can remain in `firstPrompt`.

Replace this statement with the exact limit. State that the serializer removes known credential patterns.

collector/internal/collector/collector.go
```diff
@@ -83,9 +88,48 @@ func List(since int64) ([]*session.Session, error) {
 	for _, root := range roots {
 		sessions = append(sessions, root.Session)
 	}
+	cachePreviewSessions(sessions)
```
calvintvu: Every `List` call adds full sessions to this process-wide cache. The cache never removes sessions. As a result, memory use increases until process exit.

Remove the cache. If the revision matches, call `List(expectedRevision)`. Then select the session with the requested ID.

frontend/src/pages/coslash/components/SnapshotPreviewDialog.tsx
```diff
@@ -0,0 +1,207 @@
...
+    if (!open) {
+      setLoad({ status: 'idle' });
+      return;
+    }
+    const controller = new AbortController();
+    setLoad({ status: 'loading' });
+    apiFetch(previewRequestPath(detail.id, detail.mtime), { signal: controller.signal })
+      .then((response) => {
+        if (!response.ok) throw new Error(`Preview request failed (${response.status})`);
+        return response.json() as Promise<SnapshotPreview>;
```
calvintvu: The JSON response uses a type assertion without a run-time null test. A `200` response with `null` sets `load.preview` to null. React then reads `load.preview.state` and crashes.

Before `setLoad`, make sure that the response is a non-null object. Show the existing error state for an invalid response.

collector/internal/collector/collector.go
calvintvu: also we can avoid having to maintain state

frontend/src/pages/coslash/components/SnapshotPreviewDialog.tsx
luhe19001: Fixed in 550e920. The response is parsed as unknown and validated with isSnapshotPreview before entering the loaded state. Null or malformed top-level responses now use the existing error state; the guard has regression coverage.

frontend/src/pages/coslash/lib/preview.ts
luhe19001: Fixed in 550e920. The UI now says that known credential patterns are redacted and that other sensitive text may remain. Credentials are no longer listed as structurally excluded; the privacy docs and PR description use the same wording.

frontend/src/pages/coslash/lib/preview.ts
luhe19001: Fixed in 550e920. The upload-byte guard now compares the complete decoded snapshot with the displayed snapshot. A regression test changes the repository field and verifies that the bytes are rejected.

collector/internal/collector/collector.go
luhe19001: Fixed in 550e920. The process-wide cache and all cache writes were removed. After confirming the revision, the lookup calls List(expectedRevision) and selects the requested session ID.

calvintvu (approved):

PR #98, opened 2026-08-18
feat: add opt-in Hub share build

luhe19001:
## Summary

This PR adds a fixture-backed **Share to Hub** flow for reviewing the future opt-in sharing experience.

- select sessions by repository and time window
- review the exact canonical payload before approval
- bind consent to the source revision, payload hash and size, and destination workspace
- invalidate stale reviews and ignore late preview responses
- preserve unchanged retries while requiring renewed approval when the source or destination changes
- exclude accepted items from partial retries and expose the canonical Hub route after success

## Try it locally

Append `?team-share=1` to the local coSlash URL.

- `&share-state=signed_out`
- `&share-state=pairing_required`
- `&share-state=credential_dormant`
- `&share-state=credential_revoked`
- `&share-result=partial`

## Scope

This is a clearly labeled **NO UPLOAD** build harness. It does not add real authentication, credential storage, ingestion, persistence, or navigation. Production acceptance remains gated on the end-to-end cloud integration.

## Verification

- 23 frontend tests
- production frontend build
- frontend lint and format checks
- `go test ./...`
- `go vet ./...`
- `git diff --check`
- no Go `*_test.go` files changed

Browser QA remains pending.

image

### DIFF
--- a/frontend/src/pages/coslash/CoslashPage.tsx
+++ b/frontend/src/pages/coslash/CoslashPage.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useState } from 'react';
+import { useEffect, useMemo, useState } from 'react';
 import { Search } from 'lucide-react';
 import { Button } from '@/components/ui/button';
 import { Input } from '@/components/ui/input';
@@ -28,6 +28,8 @@ import {
   type AgentVendor,
   type ViewMode,
 } from '@/pages/coslash/CoslashTabMenus';
+import { HUB_SHARE_VERSION, type DestinationResult } from '@/pages/coslash/features/sharing/model';
+import { ShareToHubDialog } from '@/pages/coslash/features/sharing/ShareToHubDialog';
 import { useDiagnostics } from '@/pages/coslash/hooks/use-diagnostics';
 import { useSessions } from '@/pages/coslash/hooks/use-sessions';
 import { useSettings } from '@/pages/coslash/hooks/use-settings';
@@ -46,6 +48,32 @@ const WINDOW_ACTIVITY_LABELS: Record<TimeWindow, string> = {
   '30d': 'active in the last 30 days',
   'all': 'across all time',
 };
+
+function fixtureDestination(search: string): DestinationResult {
+  const state = new URLSearchParams(search).get('share-state');
+  if (
+    state === 'signed_out' ||
+    state === 'pairing_required' ||
+    state === 'credential_dormant' ||
+    state === 'credential_revoked'
+  ) {
+    return { contractVersion: HUB_SHARE_VERSION, state };
+  }
+  return {
+    contractVersion: HUB_SHARE_VERSION,
+    state: 'ready',
+    destination: {
+      workspaceId: '10000000-0000-4000-8000-000000000001',
+      workspaceName: 'Compiler Team',
+      currentMemberCount: 2,
+      resultingMemberCount: 2,
+      currentApprovedSessionCount: 3,
+      historyDisclosure:
+        "Sharing this revision makes it visible to the workspace's current members. Membership and approved-session counts are current when viewed.",
+      credentialState: 'paired',
+    },
+  };
+}
 
 function CoslashPageHeader({
   onOpenSettings,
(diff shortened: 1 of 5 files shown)


frontend/src/pages/coslash/CoslashPage.tsx
```diff
@@ -253,8 +281,21 @@ export function CoslashPage() {
...
+  const [shareDialogOpen, setShareDialogOpen] = useState(false);
   const settingsState = useSettings();
   const settingsHaveError = settingsState.loadError != null || settingsState.response?.valid === false;
+  const shareParams = new URLSearchParams(window.location.search);
+  const shareBuildEnabled = shareParams.get('team-share') === '1';
+  const shareDestination = fixtureDestination(window.location.search);
+  const shareFixtureOutcome = shareParams.get('share-result') === 'partial' ? 'partial' : 'success';
+  const shareCandidates = useMemo(
+    () =>
+      sessions.map((session, index) => ({
```
calvintvu, line 295: The page loads sessions for the selected page window. The default window is one week. This code passes that limited list to the share dialog. Thus, the 30-day and All time filters cannot show older sessions. The share dialog must receive an unfiltered session list.

frontend/src/pages/coslash/features/sharing/ShareToHubDialog.tsx
```diff
@@ -0,0 +1,559 @@
...
+  const [problem, setProblem] = useState<string | null>(null);
+  const [result, setResult] = useState<ShareResult | null>(null);
+  const [fixtureAttempt, setFixtureAttempt] = useState(0);
+  const previewGeneration = useRef(0);
+
+  const visible = useMemo(
+    () => filterShareCandidates(candidates, search, window),
+    [candidates, search, window],
+  );
+  const selectedCandidates = candidates.filter(({ session }) => selected.has(localSessionId(session)));
```
calvintvu, line 107: The selected set does not change when a background refresh removes a candidate. The current candidate list then becomes empty. However, the button remains enabled because it uses the old selected set. The review function returns without a message. The button state and the selected set must use the current candidates.

frontend/src/pages/coslash/CoslashPage.tsx
luhe19001: Addressed in 24eae7f. Sharing mode now loads the all-time session source while the page still applies its visible window, so the dialog can widen to 30 days or all time.

frontend/src/pages/coslash/features/sharing/ShareToHubDialog.tsx
luhe19001: Addressed in 24eae7f. The dialog now prunes selection after candidate refreshes and derives counts, toggles, and review enablement from current candidates only.

frontend/src/pages/coslash/CoslashPage.tsx
```diff
@@ -238,7 +266,11 @@ function CoslashContent({
 export function CoslashPage() {
   const [vendor, setVendor] = useState<AgentVendor>('all');
   const [timeWindow, setTimeWindow] = useState<TimeWindow>('week');
-  const { sessions, isLoading, loadError, sessionsVersion, retrySessions } = useSessions(timeWindow);
+  const shareParams = new URLSearchParams(window.location.search);
+  const shareBuildEnabled = shareParams.get('team-share') === '1';
+  const { sessions, isLoading, loadError, sessionsVersion, retrySessions } = useSessions(
+    shareBuildEnabled ? 'all' : timeWindow,
```
calvintvu, line 272: Sharing mode now loads the all-time session list, but the page still tests the unfiltered session count for its empty state. If only older sessions exist, the default week view has no visible sessions while hasSessions is true. The page then shows “No sessions match these filters” and does not tell the user to select All. Use `sessionsInWindow.length` for the empty-state value.

luhe19001: Fixed in 26e640b. The empty-state decision now uses sessionsInWindow, so a default week view with only older sessions directs the user to the broader time window.

calvintvu (approved):

PR #99, opened 2026-08-18
Count only confirmed PRs in Codex sessions

calvintvu:
## Context

The Codex session inspector counts every `gh pr create` invocation, including failed retries. The artifact count should reflect only unique, confirmed pull requests.

## Changes

- Count successful PR creation results and assistant success directives.
- Ignore failed attempts, quoted instructions, and commentary.
- Deduplicate repeated evidence for the same PR.

## Test

- Targeted Codex parser regression tests — passed
- `go build ./...` — passed
- `git diff --check` — passed

### DIFF
--- a/collector/internal/vendors/codex/parse.go
+++ b/collector/internal/vendors/codex/parse.go
@@ -166,9 +166,10 @@ func analyzeCodexSession(file string) (*codexSessionAnalysis, error) {
 	ownID := SessionIDFromRollout(file)
 	var metas []codexMeta
 	analysis := &codexSessionAnalysis{
-		spawns:    map[string]vendors.SpawnState{},
-		fileEdits: session.NewFileEditSet(),
-		commitLog: commitObservationsByCall(rows),
+		spawns:       map[string]vendors.SpawnState{},
+		fileEdits:    session.NewFileEditSet(),
+		commitLog:    commitObservationsByCall(rows),
+		pullRequests: len(pullRequestURLs(rows)),
 	}
 	for _, row := range rows {
 		var timestamp *int64
@@ -320,9 +321,6 @@ func analyzeCodexSession(file string) (*codexSessionAnalysis, error) {
 				analysis.toolUseCount++
 				if command, ok := commandFrom(row.Payload); ok {
 					analysis.commands.Note(command, "")
-					if session.IsPullRequestCreate(command) {
-						analysis.pullRequests++
-					}
 				}
 				if questions := questionsFrom(row.Payload); len(questions) > 0 {
 					for _, question := range questions {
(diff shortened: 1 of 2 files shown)


collector/internal/vendors/codex/parse.go
```diff
@@ -699,6 +703,48 @@ type codexCommandResult struct {
...
+		case "function_call", "custom_tool_call":
+			if command, ok := commandFrom(row.Payload); ok &&
+				row.Payload.CallID != "" && session.IsPullRequestCreate(command) {
+				pending[row.Payload.CallID] = struct{}{}
+			}
+		case "function_call_output", "custom_tool_call_output":
+			if _, ok := pending[row.Payload.CallID]; !ok {
+				continue
+			}
+			result := decodeCodexCommandResult(row.Payload.Output)
```
luhe19001: **[P2] Derive success from the nested command result**

`Script completed` only shows that the JavaScript wrapper finished; the nested `exec_command` may still return a nonzero exit code. For example, `gh pr create` can fail because a PR already exists while printing that PR URL, and this path would count it as newly created. Please use the nested command result or exit code before accepting the URL.

collector/internal/vendors/codex/parse.go
```diff
@@ -699,6 +703,48 @@ type codexCommandResult struct {
...
+			if result.succeeded {
+				for _, url := range pullRequestURLPattern.FindAllString(result.output, -1) {
+					urls[url] = struct{}{}
+				}
+			}
+			delete(pending, row.Payload.CallID)
```
luhe19001: **[P2] Keep pending creates across asynchronous polling**

When `exec_command` yields a running session, its initial output contains only the session ID. Deleting the pending create here loses the correlation: the later `write_stdin` call has a different call ID and returns the PR URL. Please retain the pending create by session ID and process polling output until the command completes.

collector/internal/vendors/codex/parse.go
calvintvu: Commit [c27693d](https://github.com/centauri-ai/coslash/commit/c27693d8d1a9a724d2310c6d57749d3c347f8157) fixes this error. The parser accepts a CLI PR URL only from a completed `CommandExecution` event with exit code `0`.

collector/internal/vendors/codex/parse.go
calvintvu: Commit [c27693d](https://github.com/centauri-ai/coslash/commit/c27693d8d1a9a724d2310c6d57749d3c347f8157) fixes this error. The parser reads the final `CommandExecution` event. Polling call IDs do not affect the result.

milanshen (approved):

PR #100, opened 2026-08-18
Show Codex sessions as Waiting when user action is required

calvintvu:
## Context

Codex sessions can remain Active while they wait for a user response or command approval. This change marks those sessions as Waiting until the user responds or the approval completes.

## Changes

- Mark an open Codex turn as Waiting for an unanswered `request_user_input` call.
- Mark an open Codex turn as Waiting for a pending sandbox approval.
- Clear Waiting after an answer, completed approval, or newer user message.
- Keep sessions Active for saved approval rules and ordinary commands.

## Test

- `go test parse.go source.go fork.go types.go metadata.go discovery.go waiting_test.go` — passed
- `go build ./...` — passed
- `git diff --check origin/main...HEAD` — passed
- Manual: An unanswered `request_user_input` prompt showed Waiting.

### Screenshots

- [x] Session board — Codex session shows Waiting during an unanswered question.
Screenshot 2026-08-18 at 10 49 59

- [x] Session board — Codex session shows Waiting during a pending approval.
Screenshot 2026-08-18 at 11 21 49

### DIFF
--- a/collector/internal/vendors/codex/parse.go
+++ b/collector/internal/vendors/codex/parse.go
@@ -340,7 +340,9 @@ func analyzeCodexSession(file string) (*codexSessionAnalysis, error) {
 				}
 				if questions := questionsFrom(row.Payload); len(questions) > 0 {
 					answers, answered := questionAnswers[row.Payload.CallID]
-					analysis.waitingForInput = analysis.waitingForInput || !answered
+					_, completed := completedCalls[row.Payload.CallID]
+					analysis.waitingForInput = analysis.waitingForInput ||
+						(!answered && (row.Payload.Name == "request_user_input" || !completed))
 					for _, question := range questions {
 						answer := strings.Join(answers[question.id].values(), ", ")
 						ts := int64(0)
--- a/collector/internal/vendors/codex/parse.go
+++ b/collector/internal/vendors/codex/parse.go
@@ -5,6 +5,9 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
+	"os"
+	"os/exec"
+	"path/filepath"
 	"regexp"
 	"strings"
 
@@ -43,6 +46,7 @@ type codexSessionAnalysis struct {
 	plan             []planStep
 	turnDepth        int
 	waitingForInput  bool
+	approvalPending  bool
 	lastTurnAborted  bool
 	inReview         bool
 }
@@ -50,6 +54,7 @@ type codexSessionAnalysis struct {
 // Review mode's own generated instruction is not a user turn
 func (analysis *codexSessionAnalysis) notePrompt(prompt string, timestamp *int64) {
 	analysis.waitingForInput = false
+	analysis.approvalPending = false
 	analysis.prompts++
 	if analysis.firstUserPrompt == "" {
 		analysis.firstUserPrompt = prompt
@@ -146,7 +151,7 @@ func parse(path string) (*parsedSession, error) {
 		Spawns:   analysis.spawns,
 		Commands: analysis.commands.Labelled(),
 	}
-	if analysis.waitingForInput && parsed.InTurn {
+	if (analysis.waitingForInput || analysis.approvalPending) && parsed.InTurn {
 		status := "waiting"
 		parsed.Session.Status = &status
 	}
@@ -169,6 +174,7 @@ func analyzeCodexSession(file string) (*codexSessionAnalysis, error) {
 		return nil, err
 	}
 	questionAnswers := questionAnswersByCall(rows)
+	completedCalls := completedCallIDs(rows)
 	ownID := SessionIDFromRollout(file)
 	var metas []codexMeta
 	analysis := &codexSessionAnalysis{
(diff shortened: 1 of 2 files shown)


collector/internal/vendors/codex/parse.go
```diff
@@ -325,15 +337,22 @@ func analyzeCodexSession(file string) (*codexSessionAnalysis, error) {
...
 						if timestamp != nil {
 							ts = *timestamp
 						}
 						analysis.digest.PushQuestion(analysis.prompts, question.text, answer, ts)
 					}
 				}
+				if prefix, escalated := execApprovalPrefix(row.Payload); escalated {
+					_, completed := completedCalls[row.Payload.CallID]
+					analysis.approvalPending = analysis.approvalPending ||
+						(!completed && prefixNeedsApproval(prefix))
```
milanshen: [P2] **One-time approvals remain Waiting while commands run**

`completedCalls` only becomes true when the command returns. After the user chooses Allow once, no rule is saved, so `prefixNeedsApproval` remains true and a long-running approved command is incorrectly shown as Waiting until completion.

collector/internal/vendors/codex/parse.go
```diff
@@ -694,6 +719,65 @@ func questionAnswersByCall(rows []codexRow) map[string]map[string]codexQuestionA
...
+func prefixNeedsApproval(prefix []string) bool {
+	if len(prefix) == 0 {
+		return true
+	}
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return true
+	}
+	rules := filepath.Join(home, ".codex", "rules", "default.rules")
+	arguments := append([]string{"execpolicy", "check", "--rules", rules}, prefix...)
```
milanshen: [P2] **Saved-rule check evaluates the wrong policy**

This checks only `~/.codex/rules/default.rules` and evaluates the proposed prefix rather than the actual command. Codex evaluates command arguments against every active user, team, and trusted-project rule file, so an already-approved command can be shown as Waiting. See the [official Codex Rules documentation](https://learn.chatgpt.com/docs/agent-configuration/rules).

collector/internal/vendors/codex/parse.go
```diff
@@ -325,15 +337,22 @@ func analyzeCodexSession(file string) (*codexSessionAnalysis, error) {
 					}
 				}
 				if questions := questionsFrom(row.Payload); len(questions) > 0 {
+					answers, answered := questionAnswers[row.Payload.CallID]
+					analysis.waitingForInput = analysis.waitingForInput || !answered
```
milanshen: [P2] **Completed wrapper calls can remain Waiting**

`answered` means the output parsed as an encoded answers object, not that the call completed. I found a real rollout where an `apply_patch` payload merely contained `tools.request_user_input(...)`; the outer call completed with content blocks, but this flag remained true while the turn continued. Use the matching completed-call signal and tighten wrapper-call recognition.

milanshen (commented):
Three status-correctness findings.

collector/internal/vendors/codex/parse.go
luhe19001: **Consider clearing completed input waits**

Could we treat any completed `request_user_input` call as no longer waiting, even when it has no decoded answer? For example, an error or cancellation completes the call without an answer; if the turn continues, this flag stays true and reports a stale waiting status.

collector/internal/vendors/codex/parse.go
```diff
@@ -694,6 +719,65 @@ func questionAnswersByCall(rows []codexRow) map[string]map[string]codexQuestionA
...
+
+func execApprovalPrefix(payload codexPayload) ([]string, bool) {
+	if payload.Name != "exec_command" {
+		text := payloadText(payload)
+		if payload.Name != "exec" || !strings.Contains(text, "tools.exec_command(") ||
+			!escalatedSandboxPattern.MatchString(text) {
+			return nil, false
+		}
+		var prefix []string
+		if match := prefixRulePattern.FindStringSubmatch(text); len(match) == 2 {
```
luhe19001: **Consider checking every nested approval prefix**

Could we evaluate all escalated `tools.exec_command` calls in the outer `exec`? `FindStringSubmatch` only reads the first prefix. If that prefix is allowed but a later one still needs approval, the outer call remains blocked while the session is not marked waiting.

collector/internal/vendors/codex/parse.go
calvintvu: Deliberate in `a01bc24`. The rollout does not record an Allow once decision before the command finishes. The alternative reads `logs_2.sqlite` and depends on internal log text. We keep the session Waiting until tool output exists. A future app-server integration can consume `serverRequest/resolved`.

collector/internal/vendors/codex/parse.go
calvintvu: Fixed in `5818a4a`. The policy evaluation now uses the actual command. It loads user, team, and project rule files.

collector/internal/vendors/codex/parse.go
calvintvu: Fixed in `9859566`. The parser now uses the matching completed-call signal. A completed wrapper no longer stays Waiting when its output has no answers object.

milanshen (approved):

PR #103, opened 2026-08-18
Show OpenCode sessions as Waiting for pending approvals

calvintvu:
## Context

OpenCode keeps pending permission requests in memory. It does not store this state in the transcript database. The default TUI does not expose its permission API to other processes. Therefore, the collector cannot distinguish a permission wait from an ordinary running tool. The plugin sends the permission lifecycle to coSlash.

## Changes

- Install a global OpenCode event plugin when coSlash starts. The global plugin also covers manually launched OpenCode sessions.
- Record only request IDs, session IDs, and process IDs. Do not record prompts, tool arguments, or decisions.
- Mark root sessions as Waiting for pending child or root approvals. Clear the state after a reply and ignore records from stopped processes.

## Test

- `go build ./...` — passed
- `node --check internal/vendors/opencode/coslash-waiting.js` — passed
- `git diff --check cvu/claude-wait...HEAD` — passed
- Manual: pending approval showed Waiting; rejection returned the session to Busy.

### Screenshots

- [x] Session board — OpenCode session shows Waiting during a pending approval.
Screenshot 2026-08-18 at 13 02 14

### DIFF
--- a/collector/internal/vendors/opencode/coslash-waiting.js
+++ b/collector/internal/vendors/opencode/coslash-waiting.js
@@ -1,0 +1,36 @@
+// managed by coSlash; changes are overwritten
+import { chmod, mkdir, rename, unlink, writeFile } from "node:fs/promises"
+import os from "node:os"
+import path from "node:path"
+
+const directory = path.join(os.homedir(), ".coslash", "opencode-permissions")
+
+async function record(properties) {
+  if (typeof properties.id !== "string" || typeof properties.sessionID !== "string") return
+  await mkdir(directory, { recursive: true, mode: 0o700 })
+  await chmod(directory, 0o700)
+  const target = path.join(directory, `${properties.id}.json`)
+  const temporary = `${target}.${process.pid}.tmp`
+  await writeFile(
+    temporary,
+    JSON.stringify({ sessionID: properties.sessionID, pid: process.pid }),
+    { mode: 0o600 },
+  )
+  await rename(temporary, target)
+}
+
+async function clear(properties) {
+  if (typeof properties.requestID !== "string") return
+  await unlink(path.join(directory, `${properties.requestID}.json`)).catch(() => {})
+}
+
+export const CoslashWaitingPlugin = async () => ({
+  event: async ({ event }) => {
+    try {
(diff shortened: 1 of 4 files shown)


collector/internal/vendors/opencode/coslash-waiting.js
```diff
@@ -0,0 +1,36 @@
...
+async function clear(properties) {
+  if (typeof properties.requestID !== "string") return
+  await unlink(path.join(directory, `${properties.requestID}.json`)).catch(() => {})
+}
+
+export const CoslashWaitingPlugin = async () => ({
+  event: async ({ event }) => {
+    try {
+      if (event?.type === "permission.asked") await record(event.properties ?? {})
+      if (event?.type === "permission.replied") await clear(event.properties ?? {})
```
milanshen, line 48: [P1] Clear permissions cancelled without a reply

OpenCode cancellation interrupts the runner and removes the in-memory permission without emitting permission.replied. Pressing Esc while an approval is open therefore never calls clear(). Since the OpenCode PID remains alive, markPendingPermissions keeps the session Waiting indefinitely. Clear records on a terminal event such as session.idle as well.

milanshen, line 48: [P2] Serialize permission lifecycle writes

OpenCode invokes plugin event hooks without awaiting them, so permission.asked and permission.replied callbacks can overlap. clear() can run before record() performs its final rename, leaving a stale JSON file afterward. I reproduced this by dispatching 20 asked/replied pairs concurrently; all 20 files remained. Serialize updates or otherwise guarantee a reply wins over an in-flight record.

calvintvu: Fixed in `f0a22fe`. The plugin now processes lifecycle events in delivery order. Thus, each reply runs after its record write. The regression test covers 20 concurrent pairs.

calvintvu: I rebased the branch. The fix is now in `f0a22fe`. On `session.idle`, the plugin removes tracked permission files for that session.

milanshen (approved):

PR #105, opened 2026-08-18
Show OpenCode Waiting plugin health in diagnostics

calvintvu:
## Context

OpenCode approval status depends on the Waiting plugin installed by coSlash. Diagnostics did not show whether the plugin was installed or whether running OpenCode sessions needed a restart.

## Changes

- Show the OpenCode Waiting plugin status in the existing diagnostics checklist and copied diagnostics.
- Warn when the plugin is missing, differs from the managed version, cannot be inspected, or requires an OpenCode restart.

## Test

- `go build ./...` — passed
- `go run ./cmd/coslash doctor --json | jq -e '.checks[] | select(.id == "opencode.waiting-plugin" and .status == "ok")'` — passed\n- `git diff --check cvu/opencode-wait...HEAD` — passed\n- Manual: `/api/diagnostics` reported the installed plugin as OK.\n\n### Screenshots\n\n- [ ] Diagnostics dialog — OpenCode Waiting plugin shows its current status.

### Screenshots
Screenshot 2026-08-18 at 13 39 17

### DIFF
--- a/collector/internal/vendors/opencode/plugin.go
+++ b/collector/internal/vendors/opencode/plugin.go
@@ -1,10 +1,12 @@
 package opencode
 
 import (
+	"bytes"
 	_ "embed"
 	"errors"
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
 )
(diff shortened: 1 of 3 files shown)


collector/internal/vendors/opencode/plugin.go
```diff
@@ -15,15 +17,68 @@ const pluginName = "coslash-waiting.js"
...
+		health.Err = err
+		return health
+	}
+	output, err := exec.Command("ps", "-ww", "-axo", "pid=,lstart=,command=").Output()
+	if err != nil {
+		health.Err = fmt.Errorf("list OpenCode processes: %w", err)
+		return health
+	}
+	for _, process := range parseTUIProcesses(string(output)) {
+		if processWorkingDirectory(process.pid) != "" && process.startedAt < info.ModTime().UnixMilli() {
```
milanshen, line 64: [P2] Account for plugin-disabled and server modes

This process-age check can report healthy when the plugin is not loaded in supported nonstandard modes. parseTUIProcesses includes opencode --pure even though OpenCode disables external plugins in that mode, and excludes serve/web processes that host plugin instances. Standard bare-TUI sessions are unaffected, but these modes can show Active approvals while diagnostics reports OK. Please use a diagnostics-specific process parser to warn on --pure and include serve/web restart candidates, or document this check as TUI-only.

milanshen (approved):

PR #106, opened 2026-08-18
feat: integrate opt-in Hub sharing

luhe19001:
## Summary

- extends the fixture-backed sharing UI from PR #98 into a live local-to-Hub workflow
- pairs the local device through the operating system credential store
- rebuilds the exact canonical session snapshot the user reviewed immediately before compression and upload
- supports safe retries, partial and complete failures, and navigation to the shared repository/week

## Safety and recovery

- device credentials never enter browser responses or macOS process arguments
- the Hub derives the workspace from the device credential; the browser-provided workspace is only checked against the reviewed destination
- source, destination, or credential changes refresh the relevant data and require explicit approval again
- live sharing loads the complete candidate set while the main page retains its selected time window
- upload success does not claim that summary generation finished; that state remains `pending`

## Verification

- `go test ./...`
- `go vet ./...`
- no Go `_test.go` files are included in this PR
- 27 frontend tests
- production frontend build
- lint (two pre-existing Fast Refresh warnings only)
- format check


frontend/src/pages/coslash/features/sharing/ShareToHubDialog.tsx
```diff
@@ -544,21 +706,21 @@ export function ShareToHubDialog({
...
-              </Button>
-            )}
+          {phase === 'result' && result?.state !== 'succeeded' && retryPlan != null && retryCount > 0 && (
+            <Button onClick={retryFailed}>
+              {retryPlan.renewedReview.size + destinationRefresh.size > 0
+                ? 'Review failed sessions again'
+                : 'Retry failed with same key'}
+            </Button>
+          )}
+          {phase === 'result' && result?.state === 'failed' && retryCount === 0 && (
```
calvintvu: **Bug:** A partial result whose remaining failures are all non-retryable has `retryCount === 0`, but this fallback is restricted to `state === 'failed'`. That leaves a partial result with no footer action. Show “Back to selection” for every non-succeeded result with no retry plan.

frontend/src/pages/coslash/features/sharing/ShareToHubDialog.tsx
```diff
@@ -544,21 +706,21 @@ export function ShareToHubDialog({
...
-            retryPlan != null &&
-            retryPlan.unchanged.size + retryPlan.renewedReview.size > 0 && (
-              <Button onClick={retryPartial}>
-                {retryPlan.renewedReview.size > 0
-                  ? 'Review failed sessions again'
-                  : 'Retry failed with same key'}
-              </Button>
-            )}
+          {phase === 'result' && result?.state !== 'succeeded' && retryPlan != null && retryCount > 0 && (
+            <Button onClick={retryFailed}>
```
calvintvu: **Bug:** The retry action ignores `item.error.retryAfterSeconds`, even though the adapter preserves the Hub’s `Retry-After` value. A rate-limited request can therefore be retried immediately and repeatedly fail. Disable the action until the applicable delay expires.

frontend/src/pages/coslash/features/sharing/ShareToHubDialog.tsx
```diff
@@ -147,6 +164,34 @@ export function ShareToHubDialog({
     });
   }, [candidates]);
 
+  useEffect(() => {
+    const pairingId = pairing?.pairingId;
+    if (!open || fixtureMode || pairing?.state !== 'pending' || !pairingId) return;
+    let stopped = false;
+    const interval = globalThis.setInterval(
```
calvintvu: **Bug:** This async `setInterval` can start another poll while the previous request is still pending. With the adapter’s 30-second HTTP timeout and a 2-second interval, many polls can overlap and race credential storage/destination refresh. Schedule the next poll after the current one settles, or guard against an in-flight request.

collector/internal/hubclient/client.go
```diff
@@ -0,0 +1,387 @@
...
+
+func (c *Client) httpClient() *http.Client {
+	if c.HTTP != nil {
+		return c.HTTP
+	}
+	return &http.Client{Timeout: 30 * time.Second}
+}
+
+func (c *Client) endpoint(path string) string {
+	return c.BaseURL.ResolveReference(&url.URL{Path: path}).String()
```
calvintvu: **Bug:** `COSLASH_HUB_URL` accepts path-prefixed base URLs, but every caller passes an absolute `/v1/...` path, so `ResolveReference` discards the configured prefix. Either reject non-root base paths or resolve endpoints while retaining the prefix.

frontend/src/pages/coslash/CoslashPage.tsx
```diff
@@ -267,9 +268,11 @@ export function CoslashPage() {
   const [vendor, setVendor] = useState<AgentVendor>('all');
   const [timeWindow, setTimeWindow] = useState<TimeWindow>('week');
   const shareParams = new URLSearchParams(window.location.search);
-  const shareBuildEnabled = shareParams.get('team-share') === '1';
+  const shareFixtureEnabled = shareParams.get('team-share') === '1';
+  const [hubDestination, setHubDestination] = useState<HubDestinationResult | null>(null);
+  const shareEnabled = shareFixtureEnabled || hubDestination?.configured === true;
```
calvintvu, line 274: **Bug:** Once a live destination makes `shareEnabled` true, the live candidate mapping below still sets `previouslyShared: index === 0`, which was fixture data. The first real session is therefore falsely labeled “Previously shared · re-share.” Restrict that marker to fixture mode or remove it.

frontend/src/pages/coslash/features/sharing/ShareToHubDialog.tsx
```diff
@@ -251,11 +298,59 @@ export function ShareToHubDialog({
     setPhase('result');
   };
 
-  const retryPartial = () => {
+  const submitReviewed = async () => {
+    if (fixtureMode) {
```
calvintvu, line 348: **Optional complexity reduction:** Now that the live adapter exists, consider deleting the production fixture execution branch and using mocked `/api/hub/*` responses in tests instead. Removing the URL-driven fixture state and result generator would cut roughly 100 lines and leave one workflow.

frontend/src/pages/coslash/features/sharing/model.ts
```diff
@@ -303,6 +303,20 @@ export function planShareRetry(result: ShareResult): {
   return plan;
 }
 
+export function destinationRefreshItems(result: ShareResult): Set<string> {
```
calvintvu: **Optional complexity reduction:** Retry policy is currently split between the server-provided `retryable` flag, `RETRY_RULES`, this destination-refresh function, and component-side plan mutation. Folding refresh requirements into `planShareRetry` and using the server flag as the retry source of truth would remove a second pass and much of the policy table.

collector/internal/hubclient/credential.go
```diff
@@ -0,0 +1,82 @@
...
+	"errors"
+	"fmt"
+	"os/exec"
+	"runtime"
+	"strings"
+)
+
+var ErrNotPaired = errors.New("hub device credential is not available")
+
+type CredentialStore interface {
```
calvintvu, line 15: **Optional complexity reduction:** This interface has one production implementation, and `Delete` has no caller. Consider storing `OSKeychain` directly and removing the unused operation; reintroduce an interface when a real second implementation or test fake is shipped.

collector/cmd/coslash/main.go
```diff
@@ -130,9 +135,10 @@ func newServer(
 	guard httpsec.Guard,
 	mgr *synthesis.Manager,
 	settingsStore *settings.Store,
+	hubClients ...*hubclient.Client,
```
calvintvu: **Optional complexity reduction:** This accepts a variadic list but only ever uses the first client. An explicit `*hubclient.Client` parameter (with tests passing `nil`) is smaller and makes the one-client invariant visible.

frontend/src/pages/coslash/features/sharing/api.ts
```diff
@@ -0,0 +1,149 @@
+import { apiFetch } from '@/pages/coslash/lib/api';
+import type { DestinationResult, ShareRequest, ShareResult } from './model';
+
+export type HubDestinationResult = DestinationResult & {
```
calvintvu: **Optional complexity reduction:** `DestinationResult` and `HubDestinationResult` represent the same endpoint response. Consider putting `configured` and `hubUrl` on the existing discriminated union so the adapter and dialog share one response type.

collector/internal/hubclient/client.go
```diff
@@ -0,0 +1,387 @@
...
+
+func (c *Client) configured() bool {
+	return c != nil && c.BaseURL != nil && c.Credentials != nil
+}
+
+func (c *Client) httpClient() *http.Client {
+	if c.HTTP != nil {
+		return c.HTTP
+	}
+	return &http.Client{Timeout: 30 * time.Second}
```
calvintvu: **Security — high:** The production client has no redirect policy. Go replays POST bodies across `307/308` redirects even when the host changes, and forwards `Authorization` on exact-host/subdomain redirects—including HTTPS→HTTP. This can expose pairing device codes, canonical session snapshots, or the persistent device credential. The smallest safe policy is to reject redirects (`CheckRedirect` returning `http.ErrUseLastResponse`), or strictly require the exact approved HTTPS origin.

collector/internal/hubclient/client.go
```diff
@@ -0,0 +1,387 @@
...
+	}
+	c.pairingMu.Lock()
+	if c.pairings == nil {
+		c.pairings = make(map[string]pairingSecret)
+	}
+	c.pairings[authorization.ID] = pairingSecret{deviceCode: authorization.DeviceCode, expiresAt: authorization.ExpiresAt}
+	c.pairingMu.Unlock()
+	return PairingResult{
+		State: "pending", PairingID: authorization.ID, UserCode: authorization.UserCode,
+		VerificationURI: authorization.VerificationURI, VerificationURIComplete: authorization.VerificationURIComplete,
```
calvintvu, line 155: **Security:** These Hub-provided verification URLs are returned to the browser and passed to `window.open` without scheme or origin validation. Validate an approved HTTPS pairing origin before exposing them, and reject a `verificationUriComplete` that embeds the private device code.

collector/internal/hubclient/client.go
```diff
@@ -0,0 +1,387 @@
...
+		}
+		return PairingResult{}, problem
+	}
+	var token struct {
+		DeviceID   string `json:"deviceId"`
+		Credential string `json:"credential"`
+		TokenType  string `json:"tokenType"`
+		Scope      string `json:"scope"`
+	}
+	if err := decodeBounded(response.Body, &token); err != nil || token.Credential == "" {
```
calvintvu: **Security:** The response stores any nonempty credential while ignoring `deviceId`, `tokenType`, and `scope`. Validate a nonempty device ID, the expected `Device` token type, and the least-privilege `ingest` scope before persisting the credential.

calvintvu (commented):
Behavioral, security, and optional complexity findings from the requested review.

collector/internal/hubclient/client.go
```diff
@@ -0,0 +1,387 @@
...
+			})
+		}
+		return result, nil
+	}
+	if err != nil {
+		return ShareResult{}, err
+	}
+	result := ShareResult{ContractVersion: ContractVersion, RequestID: input.RequestID, Results: make([]ShareItemResult, 0, len(input.Items))}
+	accepted := 0
+	for _, item := range input.Items {
```
luhe19001: **[P1] Bound the batch upload duration**

Please keep the whole batch within the server 3-minute write timeout, for example with bounded concurrency or a batch deadline. This loop may upload 100 items serially, and each Hub request can wait 30 seconds. After about six timeouts, the browser can lose the response while later approved items continue uploading.

frontend/src/pages/coslash/features/sharing/ShareToHubDialog.tsx
```diff
@@ -294,9 +402,36 @@ export function ShareToHubDialog({
...
-            <Button className="mt-5" onClick={onOpenSettings}>
-              {eligibility?.action}
-            </Button>
+            {!fixtureMode && pairing?.state === 'pending' ? (
+              <div className="mt-5 rounded-lg border p-4">
+                <p className="text-sm font-semibold">Approve code {pairing.userCode}</p>
+                <p className="text-muted-foreground pt-1 text-xs">
+                  A Hub sign-in window was opened. This page will update after approval.
+                </p>
+                {pairing.verificationUriComplete && (
```
luhe19001: **[P2] Always show a pairing fallback link**

Please link to `verificationUriComplete ?? verificationUri`. The complete URI is optional, and `window.open` may be blocked. For example, if Hub returns only the basic URI, this screen says a window opened but gives the user no way to finish pairing.

collector/internal/hubclient/client.go
```diff
@@ -0,0 +1,387 @@
...
+	SharedAt           time.Time `json:"sharedAt"`
+	BriefStatus        string    `json:"briefStatus"`
+	TeamURL            string    `json:"teamUrl,omitempty"`
+}
+
+func (c *Client) Share(ctx context.Context, input ShareRequest) (ShareResult, error) {
+	if !c.configured() || c.LoadSession == nil {
+		return ShareResult{}, errors.New("Hub sharing is not configured")
+	}
+	if input.ContractVersion != ContractVersion || strings.TrimSpace(input.RequestID) == "" || len(input.Items) == 0 || len(input.Items) > 100 {
```
luhe19001: **Consider capping share batches before submission**

Could we cap the selection at 100 sessions or split it into batches before submission? The all-time view can contain more than 100 sessions, so a user can review every payload only for this validation to reject the whole request as a generic upload failure.

frontend/src/pages/coslash/CoslashPage.tsx
```diff
@@ -299,6 +304,17 @@ export function CoslashPage() {
...
 
+  const refreshHubDestination = useCallback(async () => {
+    const destination = await loadHubDestination();
+    setHubDestination(destination);
+    return destination;
+  }, []);
+
+  useEffect(() => {
+    if (shareFixtureEnabled) return;
+    void refreshHubDestination().catch(() => setHubDestination(null));
```
luhe19001: **Consider keeping configured sharing visible after destination errors**

Could we keep sharing visible and show a retry option here? A temporary Hub outage currently resets the destination to null, which hides the Share button until the page is reloaded.

frontend/src/pages/coslash/CoslashPage.tsx
```diff
@@ -288,7 +291,9 @@ export function CoslashPage() {
   const [shareDialogOpen, setShareDialogOpen] = useState(false);
   const settingsState = useSettings();
   const settingsHaveError = settingsState.loadError != null || settingsState.response?.valid === false;
-  const shareDestination = fixtureDestination(window.location.search);
+  const shareDestination = shareFixtureEnabled
+    ? ({ ...fixtureDestination(window.location.search), configured: true } as HubDestinationResult)
+    : hubDestination;
```
luhe19001: **Consider removing the fixture share history from live sessions**

Could we avoid marking the first live session as previously shared unless the Hub reports that status? This creates false “Previously shared · re-share” consent text; until real history is available, live candidates could remain unmarked.

frontend/src/pages/coslash/features/sharing/ShareToHubDialog.tsx
luhe19001: **Consider honoring the server retry delay**

Could we disable retry until retryAfterSeconds has elapsed for rate-limited results? Retrying immediately can hit the same throttle again and extend the delay.

collector/internal/hubclient/client.go
```diff
@@ -0,0 +1,387 @@
...
+		return "unsupported_snapshot_version"
+	case "device_dormant":
+		return "credential_dormant"
+	case "device_revoked":
+		return "credential_revoked"
+	case "consent_stale", "destination_changed", "idempotency_conflict", "rate_limited", "snapshot_invalid", "snapshot_too_large", "invalid_share_request", "timeout", "temporary_unavailable":
+		return code
+	case "unauthorized":
+		return "unauthorized"
+	default:
```
luhe19001: **Consider treating deletion tombstones as terminal**

Could we map source_deleted to an actionable, non-retryable error before this fallback? A deletion tombstone cannot recover, but share_failed preserves the key and offers retries that will never succeed.

luhe19001 (commented):
Author follow-ups addressed in `8f91aa3`.

frontend/src/pages/coslash/features/sharing/ShareToHubDialog.tsx
luhe19001: Fixed. Every non-succeeded result with no retry plan now shows Back to selection, including partial results.

frontend/src/pages/coslash/features/sharing/ShareToHubDialog.tsx
luhe19001: Fixed. Retry now honors the longest applicable retryAfterSeconds value, disables the action, and shows a countdown.

frontend/src/pages/coslash/features/sharing/ShareToHubDialog.tsx
luhe19001: Fixed. Pairing uses a chained timeout and schedules the next poll only after the current request settles.

collector/internal/hubclient/client.go
luhe19001: Fixed. Endpoint and handoff URL construction now preserve configured base-path prefixes, with regression coverage.

frontend/src/pages/coslash/CoslashPage.tsx
luhe19001: Fixed. The previously-shared marker is now fixture-only; live sessions remain unmarked until real history exists.

frontend/src/pages/coslash/features/sharing/ShareToHubDialog.tsx
luhe19001: Thanks. I kept the fixture branch because it is an intentional local, no-upload acceptance path for manually exercising consent and result states without a Hub. Replacing it with test-only mocks would remove that diagnostic workflow.

frontend/src/pages/coslash/features/sharing/model.ts
luhe19001: Simplified. planShareRetry now uses the server retryable flag as the source of truth and returns destination-refresh requirements directly, removing the second policy pass.

collector/internal/hubclient/credential.go
luhe19001: Removed the unused Delete operation. I kept the two-method CredentialStore because the new credential-validation tests use an in-memory fake and must not invoke the OS keychain.

collector/cmd/coslash/main.go
luhe19001: Fixed. newServer and routes now take one explicit Hub client pointer; tests pass nil.

frontend/src/pages/coslash/features/sharing/api.ts
luhe19001: Fixed. configured and hubUrl now live on the existing DestinationResult union, so the adapter and dialog share one type.

collector/internal/hubclient/client.go
luhe19001: Fixed. The Hub client now rejects every redirect before Go can replay a body or forward authorization; a regression test verifies the target is never reached.

collector/internal/hubclient/client.go
luhe19001: Fixed. Pairing URLs must match the configured scheme and origin, and the complete URL is rejected if it contains the private device code.

frontend/src/pages/coslash/features/sharing/ShareToHubDialog.tsx
luhe19001: Good catch—confirmed fixed in `8f91aa3`: every non-succeeded result without a retry plan now shows Back to selection.

frontend/src/pages/coslash/features/sharing/ShareToHubDialog.tsx
luhe19001: Good catch—confirmed fixed in `8f91aa3`: retry honors the longest applicable `retryAfterSeconds` and stays disabled during the countdown.

frontend/src/pages/coslash/features/sharing/ShareToHubDialog.tsx
luhe19001: Good catch—confirmed fixed in `8f91aa3`: chained timeouts schedule the next pairing poll only after the current request settles.

collector/internal/hubclient/client.go
luhe19001: Good catch—confirmed fixed in `8f91aa3`: API endpoints and Hub handoffs both preserve configured base-path prefixes, with regression coverage.

frontend/src/pages/coslash/CoslashPage.tsx
luhe19001: Good catch—confirmed fixed in `8f91aa3`: `previouslyShared` is fixture-only, so live sessions are no longer mislabeled.

frontend/src/pages/coslash/features/sharing/ShareToHubDialog.tsx
luhe19001: Thanks—agreed this would reduce code. We are retaining the fixture branch as the intentional local, no-upload manual diagnostic workflow; mocked API tests would not replace that path.

frontend/src/pages/coslash/features/sharing/model.ts
luhe19001: Agreed—confirmed simplified in `8f91aa3`: `planShareRetry` trusts the server `retryable` flag and returns destination-refresh requirements in one pass.

collector/internal/hubclient/credential.go
luhe19001: Agreed—confirmed in `8f91aa3`: unused `Delete` was removed; the smaller interface remains because the shipped credential tests use an in-memory fake.

collector/internal/hubclient/client.go
luhe19001: Good catch—confirmed fixed and table-tested in `8f91aa3`: persistence requires a device ID, `Device` token type, and exact `ingest` scope.

collector/cmd/coslash/main.go
luhe19001: Agreed—confirmed fixed in `8f91aa3`: `newServer` and `routes` now accept one explicit Hub client pointer.

frontend/src/pages/coslash/features/sharing/api.ts
luhe19001: Agreed—confirmed fixed in `8f91aa3`: `configured` and `hubUrl` now belong to the shared `DestinationResult` union.

collector/internal/hubclient/client.go
luhe19001: Good catch—confirmed fixed and tested in `8f91aa3`: the client rejects all redirects before credentials or POST bodies can be forwarded.

collector/internal/hubclient/client.go
luhe19001: Good catch—confirmed fixed in `8f91aa3`: pairing URLs must match the configured origin, and complete URLs containing the private device code are rejected.

collector/internal/hubclient/client.go
```diff
@@ -0,0 +1,419 @@
...
+	if response.StatusCode != http.StatusCreated {
+		return PairingResult{}, remoteError(response)
+	}
+	var authorization authorizationResponse
+	if err := decodeBounded(response.Body, &authorization); err != nil {
+		return PairingResult{}, fmt.Errorf("decode Hub pairing: %w", err)
+	}
+	if authorization.ID == "" || authorization.DeviceCode == "" || authorization.UserCode == "" ||
+		time.Now().After(authorization.ExpiresAt) ||
+		!c.validVerificationURL(authorization.VerificationURI, "") ||
```
calvintvu: **Blocking · security:** The plain verification URL is validated with an empty forbidden value, so it may contain the secret device code even though `verificationUriComplete` rejects it. The UI opens either URL in the browser. I reproduced this with `verificationUri=https://hub.example/pair?device_code=private`; `BeginPairing` accepted it. Please pass `authorization.DeviceCode` here too and cover both URL variants with tests.

frontend/src/pages/coslash/features/sharing/ShareToHubDialog.tsx
```diff
@@ -147,6 +162,45 @@ export function ShareToHubDialog({
...
+    if (!open || fixtureMode || pairing?.state !== 'pending' || !pairingId) return;
+    let stopped = false;
+    let timeout = 0;
+    const delay = Math.max(2, pairing.intervalSeconds ?? 2) * 1000;
+    const poll = async () => {
+      let finished = false;
+      try {
+        const next = await pollHubPairing(pairingId);
+        if (stopped) return;
+        setPairing((current) => ({ ...current, ...next }));
```
calvintvu: **Bug:** `setPairing` changes `pairing.state` to `paired` before the destination refresh finishes. Because that state is an effect dependency, cleanup can set `stopped = true` while the refresh is in flight; a later refresh failure is then suppressed by the catch guard. The parent retains `pairing_required`, and the UI silently falls back to “Pair this device” even though pairing succeeded. Please complete or independently surface/retry the destination refresh instead of letting effect cleanup hide this failure.

frontend/src/pages/coslash/features/sharing/api.ts
```diff
@@ -0,0 +1,145 @@
...
+      typeof item.localSessionId !== 'string' ||
+      typeof item.idempotencyKey !== 'string'
+    ) {
+      return false;
+    }
+    if (item.state === 'failed') {
+      return (
+        isRecord(item.error) &&
+        typeof item.error.code === 'string' &&
+        item.error.code in RETRY_RULES &&
```
calvintvu: **Low priority bug:** `in` walks the prototype chain, so unknown error codes such as `toString` and `__proto__` pass this contract guard and enter retry planning. That contradicts the new unknown-code rejection test. Please use `Object.hasOwn(RETRY_RULES, item.error.code)`.

collector/internal/hubclient/client.go
luhe19001: Good catch—fixed in `87e5fd2`. Both plain and complete verification URLs now reject the private device code, with regression cases for each variant.

frontend/src/pages/coslash/features/sharing/ShareToHubDialog.tsx
luhe19001: Good catch—fixed in `87e5fd2`. Pairing is not marked complete until destination refresh succeeds; failures are surfaced with an independent retry action.

frontend/src/pages/coslash/features/sharing/api.ts
luhe19001: Good catch—fixed in `87e5fd2`. The guard now uses `Object.hasOwn`, with `toString` and `__proto__` regression cases.

calvintvu (approved):

PR #115, opened 2026-08-21
Add read-only SFTP session parsing

luhe19001:
## Context

Remote monitoring should use existing SSH access without installing or running coSlash on Linux. This establishes a shared parser boundary and bounded, read-only SFTP transport while preserving local collection behavior.

## Changes

- Parse Claude and Codex data from OS-backed or SFTP-backed sources on the Mac.
- Use system SSH configuration with containment, symlink rejection, and resource limits.
- Cover missing roots, malformed data, proxy configuration, cancellation, and process cleanup.

## Test

- `cd collector && go vet ./...` — passed
- Manual: read-only SFTP negotiation through a proxy-backed OpenSSH alias — passed


collector/internal/vendors/codex/source.go
```diff
@@ -14,6 +18,68 @@ func Collect(since int64) ([]*vendors.ParsedSession, *vendors.SessionMetadata, e
...
+	files, err := FilesSource(source, SessionsRoot(home))
+	if err != nil {
+		return vendors.RemoteCollection{}, err
+	}
+	candidates := len(files)
+	files, truncated := vendors.LimitNewestSourceFiles(
+		source, files, vendors.MaxCandidateFilesPerAgent,
+	)
+	if since > 0 {
+		files = FilesSinceSource(source, files, metadata.Live, since)
```
calvintvu, line 60: **[P2] Keep session families before the file limit**

When an agent has more than 2,000 files, this limit runs before `FilesSinceSource`. It can keep a recent child and remove its older root. The family scan cannot find the child without the root. As a result, the collector omits an active session. Apply the time and family selection before this limit, or limit complete families. The Claude remote collector has the same order.

collector/internal/remote/source.go
```diff
@@ -0,0 +1,240 @@
...
+	if file.exhausted {
+		return 0, fmt.Errorf("%w: %s", ErrFileLimit, file.path)
+	}
+	if int64(len(buffer)) > file.fileLeft+1 {
+		buffer = buffer[:file.fileLeft+1]
+	}
+	n, err := file.ReadCloser.Read(buffer)
+	file.fileLeft -= int64(n)
+	if file.total.Add(int64(n)) > file.totalLimit {
+		return n, ErrTotalLimit
```
calvintvu: **[P2] Reserve bytes before each read**

Each concurrent reader reads the full buffer before it adds the byte count. Near `MaxTotalBytes`, several readers can exceed 128 MiB before `ErrTotalLimit`. Limit each read to the remaining shared budget. Alternatively, reserve the requested bytes atomically before the read.

collector/internal/vendors/codex/source.go
luhe19001: Fixed in 56dedeb. Remote collection now filters by time before applying a newest-family cap, and the cap keeps each Claude or Codex root family intact. Added regression coverage for family-preserving selection.

collector/internal/remote/source.go
luhe19001: Fixed in 56dedeb. Reads now atomically reserve their allowed portion of the shared byte budget before calling the SFTP reader, releasing any unused reservation. Added concurrent-reader and exact-limit coverage.

milanshen (approved):

milanshen (changes requested):
please leave out the implementation docs

milanshen (approved):

PR #116, opened 2026-08-21
Collect remote sessions on the Mac

luhe19001:
## Context

Remote sessions are collected on macOS through SSH/SFTP, with all coSlash parsing and caching kept local. The architecture now clarifies the privacy-preserving cache boundary and compatibility behavior.

## Changes

- Document opaque fingerprinting and bounded reparsing for remote session refreshes.
- Specify that callers using the legacy session endpoint keep the existing local-session response.

## Test

- Not run — documentation-only PR update.


calvintvu:
I found two P1 cache problems:

1. **Keep the last complete cache after a limited refresh.**

When one agent fails or collection stops early, the manager saves and installs the incomplete result before it detects the limit. This action removes valid sessions from the last complete cache. Keep the last complete cache when `limitedResultReason(result)` reports a limit.

2. **Do not reuse a cache after an SSH alias change.**

A settings change can keep the source ID and replace `SSHAlias`. After restart, the manager loads the old host cache and labels those sessions with the new alias. If the new host stays unavailable, the incorrect stale sessions remain visible. Clear the cache after an alias change, or store each cache by SSH alias.

luhe19001:
Addressed both P1 cache issues in 405eddf.

- Limited refreshes now leave the installed and persisted last complete snapshot untouched, while reporting the limited state and displaying that snapshot as stale.
- Changing the SSH alias for an existing source ID clears its cached snapshot before the new host is loaded, so unavailable replacement hosts cannot surface sessions from the old host.

Added regression coverage for both cases. Verified with `cd collector && go vet ./... && go test ./...`.

luhe19001:
Rebased onto `ssh-mac/p1-parser-sftp` and resolved the merge conflicts. The resolution keeps the base branch’s session-family/byte-limit behavior and retains this PR’s selected-file fingerprints for cache persistence.

Verified with `cd collector && go vet ./... && go test ./...`.

milanshen (approved):

milanshen (approved):

PR #121, opened 2026-08-28
Collapse background re-homed Claude sessions into one card

calvintvu:
## Context

Claude Code can move a terminal session onto its background surface a short time after the session starts. It writes a second transcript that copies the conversation of the first, under a new session id. coSlash reads every root transcript as its own session, so one conversation renders as two identical cards and inflates the active-session count.

## Changes

- A background transcript that holds every conversation row of another root now supersedes that root. Only the newer card remains, and it keeps the session id that Resume needs.
- The tokens of the superseded transcript fold into the survivor, so the cost of the pair does not change.
- Subagent transcripts re-point to the surviving session, so they do not lose their parent.
- A deliberate `claude --resume` branch keeps its own card. A re-home of such a branch supersedes only the branch, not the root it resumed from.

## Test

- `go test ./...` — passed
- `go vet ./...` — passed
- Manual: ran the pass over a local transcript store of 79 root transcripts. It collapsed two re-home pairs and left two `claude --resume` pairs separate.

### Screenshots

- [x] Session list for a project with a re-homed session — BEFORE.
Screenshot 2026-08-28 at 12 54 40

- [x] Session list for a project with a re-homed session — AFTER.
Screenshot 2026-08-28 at 12 55 17


luhe19001:
Fixed in 080cf90. `GetSessionFamily` now identifies the successor’s re-home chain before selecting files, so it also parses subagents that remain under a superseded predecessor directory. Added a regression test covering the preview path.

collector/internal/vendors/claude/source.go
```diff
@@ -35,6 +35,7 @@ func GetSessionFamily(id string) ([]*vendors.ParsedSession, *vendors.SessionMeta
 func parseFiles(files []string) []*vendors.ParsedSession {
 	parsed := vendors.ParseFiles(files, parse)
 	applyForkedUsage(parsed)
+	parsed = collapseBackgroundRehomes(parsed)
```
luhe19001, line 189: `Collect` feeds every root + every subagent into this path, so ParentID repointing works. `GetSessionFamily` (preview) filters by path-derived root id *before* `parseFiles`, so when the card id is the successor, subagents that remain under the superseded predecessor directory are never selected and cannot be repointed here.

Worth either selecting predecessor-chain files for a re-homed id in `GetSessionFamily`, or documenting that preview of a re-homed session omits those subagents.

luhe19001 (commented):
Reviewed the re-home collapse approach — the uuid-containment + bg-only successor rules, token fold after `applyForkedUsage`, and fixture coverage look solid for the board/list path.

One follow-up on the preview family path (inline); otherwise LGTM for the stated goal.

collector/internal/vendors/claude/fork.go
```diff
@@ -106,8 +108,118 @@ func applyForkedUsage(parsed []*parsedSession) {
...
+// writing a fresh transcript that copies every row of the original, so both
+// files would otherwise serve as separate cards for one conversation.
+//
+// The successor survives: it holds the newer turns and the session id that is
+// actually live, so Resume targets the real session. applyForkedUsage ran
+// first and made the pair's usage disjoint, so folding tokens in is exact.
+//
+// Containment is measured over row uuids, not usage: a re-home whose only new
+// rows are a prompt or an aborted turn adds no usage-bearing message.id, so
+// the two files' usage sets come out equal and hide the duplicate.
```
milanshen: remove excessive inline comments

luhe19001 (approved):

PR #125, opened 2026-08-31
Fix remote connection results and expired-tab polling

luhe19001:
## Context

A successful remote SFTP probe could be reported as missing agent data because the connection check reused refresh coverage rules. Separately, an expired browser link continued polling the local API after authentication failed.

## Changes

- Report successful SSH/SFTP connection tests as ready without evaluating Claude or Codex session coverage.
- Stop session polling after an authentication failure while showing the existing expired-link message.

## Test

- `cd collector && go vet ./...` — passed
- `cd frontend && npm test -- --run` — passed
- `cd frontend && npm run lint` — passed with existing warnings
- `cd frontend && npm run format:check` — passed

### DIFF
--- a/collector/internal/remote/manager.go
+++ b/collector/internal/remote/manager.go
@@ -218,14 +218,7 @@ func (manager *Manager) TestAlias(ctx context.Context, alias string) (Health, er
 		health.DiagnosticStderr = redactDiagnostic(diagnostic)
 		return health, nil
 	}
-	if reason := limitedResultReason(result); reason != nil {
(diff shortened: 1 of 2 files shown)


luhe19001:
Resolved the TypeScript failure: the outer polling flag is now set only after narrowing the `unknown` error to `ApiAuthenticationError`, while the current request uses its own local boolean. Verified with `npm run build`, frontend tests, lint, formatting, and the full Go suite.

calvintvu (approved):

PR #127, opened 2026-08-31
Fix macOS Hub device pairing persistence

calvintvu:
## Context

macOS device pairing failed when coSlash ran from a terminal. The Keychain command inherited the terminal and waited for interactive password input. A canceled polling request could also stop credential storage after Hub approved the device.

## Changes

- Run the macOS Keychain command in a separate session.
- Provide both password lines required for a new Keychain item.
- Finish credential storage after request cancellation, with a 30-second limit.
- Add regression coverage for canceled polling requests.

## Test

- `go test ./internal/hubclient ./cmd/coslash -count=1` — passed
- `make release` — passed
- Manual: verified Keychain storage from a macOS PTY.

### DIFF
--- a/collector/internal/hubclient/credential.go
+++ b/collector/internal/hubclient/credential.go
@@ -8,6 +8,7 @@ import (
 	"os/exec"
 	"runtime"
 	"strings"
+	"syscall"
 )
 
 var ErrNotPaired = errors.New("hub device credential is not available")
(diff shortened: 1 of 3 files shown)


collector/internal/hubclient/credential.go
```diff
@@ -63,3 +63,14 @@ func (s OSKeychain) Save(ctx context.Context, credential string) error {
 	}
 	return nil
 }
+
+func darwinCredentialCommand(ctx context.Context, store OSKeychain, credential string) *exec.Cmd {
+	command := exec.CommandContext(ctx, "/usr/bin/security", "add-generic-password", "-U", "-s", store.Service, "-a", store.Account, "-w")
+	command.Stdin = darwinCredentialInput(credential)
+	command.SysProcAttr = &syscall.SysProcAttr{Setsid: true}
```
luhe19001, line 70: On macOS, `security add-generic-password ... -w` with no password argument invokes `getpass` twice, which reads from the controlling terminal rather than `Stdin`. `Setsid` creates a new session with no controlling terminal, so the command cannot read either password and every successful pairing fails when saving the credential. Please do not detach this command, or invoke `security` with a mechanism that supplies the password without terminal prompts.

calvintvu: Setsid is intentional. macOS getpass(3) first reads from /dev/tty, but when no controlling terminal is available it falls back to stdin. The child therefore consumes the two credential lines supplied through command.Stdin. Without Setsid, /dev/tty wins and security prompts in coSlash’s terminal which is the behavior this PR fixes. Passing the credential as the -w argument would expose it in the process arguments.

luhe19001 (approved):

PR #130, opened 2026-09-01
fix: fail fast when Go or npm is missing for make release

luhe19001:
## Summary
- Add `check-go` and `check-node` Makefile prerequisites so `make release` fails immediately with install hints instead of after the frontend build
- Clarify in README and CONTRIBUTING that building from source requires Go and Node, while end users should use the prebuilt install paths
- Add a troubleshooting section for source-build failures

## Context
A customer hit `go: not found` after running `make release` without Go installed. The build spent time on `npm ci`/`npm run build` first, then failed with an unhelpful error.

## Test plan
- [x] Verified `make release` without Go in PATH fails immediately with the new error message
- [ ] CI release job passes unchanged (toolchain already present in Actions)

### DIFF
--- a/collector/Makefile
+++ b/collector/Makefile
@@ -18,6 +18,7 @@ LITELLM_URL := https://raw.githubusercontent.com/BerriAI/litellm/main/model_pric
 OPENCODE_MODELS_URL := https://models.opencode.ai/api.json
 GO_REQUIRED := $(shell sed -n 's/^go //p' go.mod)
 NODE_REQUIRED := $(shell tr -d 'v\n' < $(FRONTEND)/.nvmrc)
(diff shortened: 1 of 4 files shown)


collector/Makefile
```diff
@@ -10,15 +10,33 @@ LDFLAGS := -X main.version=$(VERSION)
...
+check-go:
+	@command -v go >/dev/null 2>&1 || { \
+		echo "error: Go $(GO_REQUIRED)+ is required to build coSlash from source."; \
+		echo "Install Go from https://go.dev/dl/ or run: brew install go"; \
+		echo "To install a prebuilt binary instead, see the Install section in README.md."; \
+		exit 1; \
+	}
+
+check-node:
+	@command -v npm >/dev/null 2>&1 || { \
```
calvintvu: `check-node` accepts unsupported Node versions. It only makes sure that `npm` exists. For example, Node v23.11.0 passes this target, but `.nvmrc` requires v24.4.1 and `package.json` requires Node 24 or later. Then `make release` starts `npm ci` instead of reporting the required version. Compare `node --version` with `NODE_REQUIRED` in this target. Apply the same version check to `check-go`.

calvintvu (approved):

PR #132, opened 2026-09-01
Calm session list hierarchy and density

luhe19001:
## Summary
- Reduce session-list chrome: shortened IDs, quieter liveness, Autonomous-only modality, machine badge only when the visible set spans sources, and rolling time windows (`24h` / `7d` / `30d` / `All`).
- Increase density with hairline rows, quieter cost, louder waiting/active status, collapsed subagent chips, and a slimmer header/toolbar.
- Document the pass in `docs/ui-enhancement2.md` with before/after screenshots.

## Test plan
- [ ] Open the list view and confirm vendor labels show full names (Claude Code / Codex / OpenCode)
- [ ] Confirm time filters are only 24 hours / 7 days / 30 days / All
- [ ] Confirm Interactive no longer appears on every row; Autonomous still shows when relevant
- [ ] With local + remote sessions visible, machine labels appear; with a single-source filter set, they hide
- [ ] Subagents start collapsed as an `N subagents` chip and expand on click
- [ ] Waiting / Active stay visually louder than Idle / Liveness unknown / cost
- [ ] `cd frontend && npm test && npm run lint && npm run format:check`


luhe19001:
no need

PR #133, opened 2026-09-02
Integrate SSH monitoring in Machines and releases

luhe19001:
## Summary

- Add remote setup, helper lifecycle, and ownership APIs.
- Add Machines controls, diagnostics, fallback, and recovery states.
- Embed AMD64 and ARM64 helpers in release binaries.
- Add packaged smoke coverage and user documentation.
- Include follow-up fixes for freshness, setup probes, and reverification.

## Validation

- Full Go, vet, and targeted race suites pass.
- Frontend tests, lint, formatting, and build pass.
- Embedded-helper and packaged release smoke tests pass.
- Real-host lifecycle passed on Linux AMD64 and ARM64.

## Stack

- 3 of 3.
- Depends on #136, which depends on #135.
- Merge in stack order.

### DIFF
--- a/collector/internal/remote/helpermanager.go
+++ b/collector/internal/remote/helpermanager.go
@@ -73,10 +73,13 @@ const (
 type lifecycleFactory func(alias string) (Lifecycle, error)
 
 type helperTarget struct {
-	path    string
-	version string
-	state   LifecycleState
+	path     string
+	version  string
+	state    LifecycleState
+	artifact Artifact
 }
+
+type helperVerifyFunc func(context.Context, string, helperTarget) error
 
 type helperProbeState string
 
@@ -244,7 +247,7 @@ func (manager *Manager) runHelperDiscovery(ctx context.Context, config settings.
 		status := lifecycleStatus(result)
 		manager.helper = &status
 		if !useSFTPFallbackAfterDiscovery(result) {
-			manager.helperTarget = &helperTarget{path: result.Path, version: result.Artifact.Version, state: result.State}
+			manager.helperTarget = &helperTarget{path: result.Path, version: result.Artifact.Version, state: result.State, artifact: result.Artifact}
 			manager.helperVersion = result.Artifact.Version
 			// This records ownership locally, never remote execution authority.
 			if storeErr := manager.cache.StoreHelperVersion(config.ID, result.Artifact.Version, config.SSHAlias); storeErr != nil {
@@ -319,7 +322,7 @@ func (manager *Manager) SetupHelper(ctx context.Context, consent Consent) Health
 			manager.helperTarget = nil
 			return manager.healthLocked(manager.lastRequestedMs)
 		}
-		manager.helperTarget = &helperTarget{path: result.Path, version: result.Artifact.Version, state: result.State}
+		manager.helperTarget = &helperTarget{path: result.Path, version: result.Artifact.Version, state: result.State, artifact: result.Artifact}
 		manager.helperVersion = result.Artifact.Version
 		manager.helperProbe = helperProbeReady
 	} else {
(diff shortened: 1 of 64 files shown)


frontend/src/pages/coslash/components/MachinesSettingsSection.tsx
```diff
@@ -41,26 +88,63 @@ export function MachinesSettingsSection({
...
   };
 
-  const removeHost = () => {
-    if (!confirmRemove) {
-      setConfirmRemove(true);
+  const setupHelper = async (consent: 'install' | 'upgrade') => {
+    setHelperAction(consent);
+    setTestError(null);
+    try {
+      const result = await setupRemoteHelper(consent);
```
calvintvu: An unsaved alias can install the helper on the wrong host. `runTest` uses the draft alias, but `setupRemoteHelper` sends no alias. The server uses the saved alias from `manager.cfg`. If a user changes host A to host B, the test targets B, but installation targets A. Disable this action until the user saves the alias. Alternatively, send the tested alias. Make sure that the server rejects an alias mismatch.

collector/internal/remote/helpermanager.go
```diff
@@ -0,0 +1,518 @@
...
+		// protocol and inspected every supported root, but there is simply no
+		// agent data yet. Keep that useful board state while reporting the setup
+		// operation itself as successful.
+		return HelperTestResult{
+			Health:    manager.healthLocked(manager.lastRequestedMs),
+			Succeeded: *reason == ReasonNoSupportedData,
+			Reason:    reasonPtr(*reason),
+		}
+	}
+	manager.state = StateOK
```
calvintvu: `TestHelper` does not store `result.Snapshot` or `result.Sessions`. However, this code changes the board state to `StateOK`. A stale cached session then appears current and becomes eligible for aggregates until the next refresh. Keep the prior board state. If this method sets `StateOK`, store the probe result first.

collector/internal/remote/helpermanager.go
```diff
@@ -0,0 +1,518 @@
...
+	if err != nil {
+		status := unavailableHelperStatus(err)
+		manager.helper = &status
+		manager.helperProbe = helperProbeFallback
+	} else {
+		status := lifecycleStatus(result)
+		manager.helper = &status
+		if !useSFTPFallbackAfterDiscovery(result) {
+			manager.helperTarget = &helperTarget{path: result.Path, version: result.Artifact.Version, state: result.State, artifact: result.Artifact}
+			manager.helperVersion = result.Artifact.Version
```
calvintvu: This assignment occurs before `StoreHelperVersion`. If that write fails, the failure branch clears the target but keeps `helperVersion`. Health then reports recorded ownership that does not exist on disk. Alias changes stay blocked until restart, but restart loses the ownership protection. Assign `helperVersion` only after a successful write. Otherwise, clear it in the failure branch.

collector/internal/remote/helpermanager.go
```diff
@@ -0,0 +1,526 @@
...
+		manager.helper = &status
+		health := manager.healthLocked(manager.lastRequestedMs)
+		manager.mu.Unlock()
+		return health, nil
+	}
+	result := lifecycle.SetupWithLoader(ctx, document, consent, provider.LoadArtifact)
+	manager.mu.Lock()
+	defer manager.mu.Unlock()
+	if manager.cfg == nil || manager.cfg.ID != config.ID || manager.cfg.SSHAlias != config.SSHAlias {
+		return manager.healthLocked(manager.lastRequestedMs), nil
```
calvintvu, line 346: [P1] Block alias changes while helper setup is running

This check happens after SetupWithLoader can install the helper. Another request can change the saved SSH alias during that call because the manager lock is released. The install then finishes on the old host, this branch returns, and no helper ownership is saved. The app can no longer track or uninstall that helper.

I reproduced this by pausing LoadArtifact, changing the alias with ApplySettings, and then allowing setup to finish. The helper file exists on the old host, but LoadHelperOwnership returns no record.

Please prevent alias changes while setup is running, or keep enough ownership data to clean up the helper on the original host.

luhe19001: Fixed in b6be359. Helper setup now reserves the configured remote identity while SetupWithLoader runs; settings changes that would switch hosts are rejected until setup completes. Added a regression test that pauses artifact loading and verifies ownership remains recorded for the original alias.

milanshen (approved):

PR #135, opened 2026-09-02
Add incremental SFTP collection for remote sessions

luhe19001:
## Summary

- Define bounded remote facts and wire contracts.
- Read Claude Code and Codex data through existing SSH configuration.
- Cache normalized families for incremental refreshes.
- Preserve partial results and report explicit coverage.

## Validation

- Go tests and vet pass.
- Frontend tests, lint, formatting, and build pass.

## Stack

- 1 of 3.
- Review and merge this PR first.
- The optional helper follows in #136.

PR #136, opened 2026-09-02
Add a verified SSH collection helper

luhe19001:
## Summary

- Add a bounded Linux collection helper.
- Execute it through the existing SSH connection.
- Verify platform, digest, ownership, mode, and protocol.
- Require explicit lifecycle consent and retain SFTP fallback.

## Validation

- Go tests and vet pass.
- Remote, helper, and protocol race tests pass.

## Stack

- 2 of 3.
- Depends on #135.
- Machines integration and release packaging follow in #133.

### DIFF
--- a/collector/internal/remote/lifecycle.go
+++ b/collector/internal/remote/lifecycle.go
@@ -1,0 +1,657 @@
+package remote
+
+// This file deliberately keeps helper lifecycle operations behind a small
+// remote boundary.  An SSH/SFTP implementation must perform that boundary's
+// filesystem operations relative to a trusted directory descriptor with
+// no-follow semantics.  A sequence of remote shell `test`, `mv`, and `rm`
+// commands is not an acceptable implementation: it leaves a check/use race.
+
+import (
+	"bytes"
+	"context"
+	"crypto/ed25519"
+	"crypto/sha256"
+	"encoding/base64"
+	"encoding/binary"
+	"encoding/hex"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io/fs"
+	"regexp"
+	"slices"
+	"sort"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/centauri-ai/coslash/collector/internal/remotefacts"
+	"github.com/centauri-ai/coslash/collector/internal/remoteprotocol"
+)
+
+const (
+	// HelperInstallBase is the only remote directory lifecycle code may touch.
+	// It is user-owned and versioned, so a failed update cannot replace a known
+	// working helper.  A noexec mount is a deliberate SFTP fallback, not a
+	// reason to try a privileged or alternate installation location.
+	HelperInstallBase            = "~/.local/lib/coslash/helpers"
+	HelperFileName               = "coslash-helper"
+	MaxHelperArtifactBytes int64 = 128 << 20
+)
+
+var (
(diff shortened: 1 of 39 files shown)


collector/internal/remote/lifecycle.go
```diff
@@ -0,0 +1,657 @@
...
+	}
+	return document.Metadata, nil
+}
+
+func canonicalMetadataPayload(metadata ReleaseMetadata) []byte {
+	artifacts := slices.Clone(metadata.Artifacts)
+	sort.Slice(artifacts, func(i, j int) bool {
+		return artifactIdentity(artifacts[i]) < artifactIdentity(artifacts[j])
+	})
+	payload, _ := json.Marshal(ReleaseMetadata{Artifacts: artifacts})
```
calvintvu: `canonicalMetadataPayload` omits `Sequence` and `ExpiresAtUnix`. The signature therefore does not authenticate these fields.

An attacker can change the expiry or raise the sequence without invalidating the signature. A high sequence also blocks later valid metadata.

Include both fields in the canonical payload. Add a mutation test.

collector/internal/remotehelper/source.go
```diff
@@ -0,0 +1,217 @@
...
+// boundedFile stops a transcript read at the helper's per-file limit even when
+// the file grows while it is being parsed.
+type boundedFile struct {
+	*os.File
+	path string
+	left int64
+}
+
+func (file *boundedFile) Read(buffer []byte) (int, error) {
+	if file.left <= 0 {
```
calvintvu: `boundedFile.Read` returns `ErrFileLimit` when `left` is zero. It does this before it reads the underlying file for EOF.

As a result, a file whose size equals `MaxFileBytes` always fails. I reproduced this failure with a four-byte limit.

Permit one EOF read after the byte limit is reached. Add a boundary test for a file of the exact limit.

collector/internal/remote/lifecycle_ssh.go
```diff
@@ -0,0 +1,368 @@
...
+		if err != nil {
+			return fmt.Errorf("inspect helper filesystem: %w", err)
+		}
+		if vfs.Flag&linuxStatVFSNoExec != 0 {
+			return ErrHelperNoExec
+		}
+		if err := removeStaleLifecycleTemporary(session.client, temporary, request.OwnerUID); err != nil {
+			return err
+		}
+		file, err := session.client.OpenFile(temporary, os.O_WRONLY|os.O_CREATE|os.O_EXCL)
```
calvintvu: `validateLifecycleDirectories` validates each path component. Later, `OpenFile`, `PosixRename`, and `inspectLifecycleFile` use the path again.

A process on the remote host can replace a validated directory with a symlink between these operations. The later operation then uses another tree.

This behavior breaks the no-follow, directory-relative requirement in `LifecycleRemote`. Standard SFTP path operations do not close this race.

Use an operation that keeps a trusted directory handle for the complete install. Add a race test that swaps an intermediate directory.

luhe19001: Addressed in b6d6570. The verified staged helper now installs itself through Linux `openat`/`renameat` operations, retaining the final version-directory descriptor through temporary creation, sync, validation, and activation. It hashes `/proc/self/exe` against the signed artifact digest first. Added a Linux race test that swaps an intermediate directory for a symlink after the version descriptor is opened and verifies the install stays in the original tree.

milanshen (approved):

PR #139, opened 2026-09-02
Polish remote Machines SSH setup UX

luhe19001:
## Summary
- Soften first-time SSH failure messaging on the host strip and guide users to fix SSH, then Retry.
- After a successful connection test (or settings save), clear retry backoff and refresh the board automatically.
- Merge Machines **Test connection** / **Install helper** into a progressive flow: test first, then install as the next step.
- Shorten the initial remote retry backoff from 3 minutes to 30 seconds.

Stacked on top of #133 (`hlu/ssh-mix-07`).

## Test plan
- [ ] Configure a bad SSH alias → strip shows recoverable warning copy, not a vague red “connection failed”
- [ ] Fix SSH / run `ssh <alias>` once → **Test connection** succeeds → board strip clears without clicking Retry
- [ ] Save remote settings after SSH works → board refresh starts automatically
- [ ] Successful test reveals **Install helper** (optional) instead of a disabled Install beside Test
- [ ] `go test ./internal/remote/ -run TestSuccessfulAliasTestClearsBackoff`
- [ ] `npx vitest run src/pages/coslash/lib/host-strip.test.ts`

### DIFF
--- a/collector/internal/remote/limits.go
+++ b/collector/internal/remote/limits.go
@@ -15,8 +15,10 @@ const (
 	DefaultMaxDepth       = 16
 	DefaultMaxStderrBytes = 8 << 10
 
-	FreshnessInterval   = 3 * time.Minute
-	InitialRetryBackoff = 3 * time.Minute
+	FreshnessInterval = 3 * time.Minute
+	// Short first retry so fixing SSH (keys/host) recovers the board without a
+	// long wait or a manual Retry click; later failures still back off hard.
+	InitialRetryBackoff = 30 * time.Second
 	MaxRetryBackoff     = 30 * time.Minute
 	ActivityWindowMs    = 2 * 60_000
 	MaxDiagnosticBytes  = 2 << 10
--- a/frontend/src/pages/coslash/CoslashPage.tsx
+++ b/frontend/src/pages/coslash/CoslashPage.tsx
@@ -425,7 +425,18 @@ export function CoslashPage() {
 
   const saveSettings = async (...args: Parameters<typeof settingsState.save>) => {
     const ok = await settingsState.save(...args);
-    if (ok) retrySessions();
+    if (ok) {
+      // Kick a remote collect immediately after save so SSH recovery is not
+      // stuck behind the previous failure backoff.
+      setRemoteRetryInFlight(true);
+      void retryRemoteRefresh()
+        .catch(() => undefined)
+        .finally(() => {
+          setRemoteRetryInFlight(false);
+          retrySessions();
+          if (diagnosticsOpen) refreshDiagnostics();
+        });
+    }
     return ok;
   };
 
@@ -548,6 +559,7 @@ export function CoslashPage() {
         saveError={settingsState.saveError}
         isSaving={settingsState.isSaving}
         onSave={saveSettings}
+        onRemoteConnectionVerified={handleRemoteRetry}
       />
     </div>
   );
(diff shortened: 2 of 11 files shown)


frontend/src/pages/coslash/components/MachinesSettingsSection.tsx
```diff
@@ -90,13 +103,15 @@ export function MachinesSettingsSection({
...
-      setTestResult(await testRemoteAlias(alias));
+      const result = await testRemoteAlias(alias);
+      setTestResult(result);
       try {
         setHostStatus(await remoteStatus());
       } catch {
         // SFTP test success remains useful if the optional helper inspection
         // endpoint is temporarily unavailable.
       }
+      if (result.state === 'ok' && hasHost) onConnectionVerified?.();
```
calvintvu, line 114: This callback reloads the board before the remote collection finishes. The retry endpoint returns 202 immediately after it starts the collection. The board then shows `connecting` until the 60-second session poll runs. Poll until the machine leaves `connecting`, or send a board update after the collection finishes.

collector/internal/remote/limits.go
```diff
@@ -15,8 +15,10 @@ const (
 	DefaultMaxDepth       = 16
 	DefaultMaxStderrBytes = 8 << 10
 
-	FreshnessInterval   = 3 * time.Minute
-	InitialRetryBackoff = 3 * time.Minute
+	FreshnessInterval = 3 * time.Minute
+	// Short first retry so fixing SSH (keys/host) recovers the board without a
+	// long wait or a manual Retry click; later failures still back off hard.
+	InitialRetryBackoff = 30 * time.Second
```
calvintvu, line 21: This change makes `MaxRetryBackoff` unreachable. `retryBackoff` stops increasing after four shifts, so the new maximum is `30s << 4`, or 8 minutes. An unavailable host now retries every eight minutes instead of every 30 minutes. Increase the shift limit so the delay reaches `MaxRetryBackoff`.

frontend/src/pages/coslash/CoslashPage.tsx
```diff
@@ -425,7 +425,18 @@ export function CoslashPage() {
 
   const saveSettings = async (...args: Parameters<typeof settingsState.save>) => {
     const ok = await settingsState.save(...args);
-    if (ok) retrySessions();
+    if (ok) {
```
calvintvu, line 428: Optional: This block duplicates `handleRemoteRetry` at lines 414-424. Replace the block with `if (ok) handleRemoteRetry();`. This change removes about nine lines and keeps one retry flow.

frontend/src/pages/coslash/components/MachinesSettingsSection.tsx
luhe19001: Fixed in 243f23f. Remote retries now poll status until collection leaves connecting before reloading the board; added a regression test.

collector/internal/remote/limits.go
luhe19001: Fixed in 243f23f. The retry shift limit now reaches the 30-minute cap, with regression coverage for the capped delay.

frontend/src/pages/coslash/CoslashPage.tsx
luhe19001: Fixed in 243f23f. saveSettings now delegates to handleRemoteRetry, keeping a single retry flow.

milanshen (approved):

PR #142, opened 2026-09-03
Harden SSH remote collection, handoffs, and setup

luhe19001:
## Context

Remote sessions need complete inspection data, while SSH monitoring and connector setup must recover predictably from unavailable hosts and stale connections.

## Changes

- Preserve full remote session facts and commands for remote inspection.
- Keep complete multi-session families within the protocol record size limit by compacting oversized display details.
- Pass remote Codex handoff context correctly when launching a new session.
- Wire remote Start Fresh and Resume controls to the SSH launch path.
- Add machine health, connection freshness, cached-history age, and setup-progress API data.
- Bound SSH connection attempts and retry once after clearing a stale control connection.
- Install and verify the connector at its canonical path, then safely replace only the previously owned version.
- Allow local removal of an SSH host even when it is unreachable.

## Test

- `go test ./...` — passed
- `go vet ./...` — passed
- `gofmt -l ./cmd ./internal` — clean
- `git diff --check` — passed
- `npm test` — passed
- `npm run format:check` — passed
- `npm run build` — passed
- Manual: verified a multi-session changed-family record with 1.2 MiB of source display data remains under the 1 MiB protocol limit.

### DIFF
--- a/collector/cmd/coslash/api.go
+++ b/collector/cmd/coslash/api.go
@@ -238,21 +238,36 @@ func cleanupHandoffs() {
 	}
 }
 
-func handleLaunch(w http.ResponseWriter, r *http.Request, settingsStore *settings.Store) {
+func handleLaunch(w http.ResponseWriter, r *http.Request, settingsStore *settings.Store, remoteManager *remote.Manager) {
 	state := settingsStore.State()
 	if !state.Valid {
 		http.Error(w, state.Error+"; open Settings to repair it", http.StatusConflict)
 		return
 	}
 	query := r.URL.Query()
-	found, err := collector.GetSessionFacts(query.Get("id"))
+	sourceID, err := parseSourceID(query.Get("source"))
 	if err != nil {
-		log.Printf("launch: %v", err)
-		http.Error(w, "could not load session", http.StatusInternalServerError)
+		http.Error(w, err.Error(), http.StatusBadRequest)
 		return
 	}
+	var found *session.Session
+	alias := ""
+	if sourceID == localSourceID {
+		found, err = collector.GetSessionFacts(query.Get("id"))
+		if err != nil {
+			log.Printf("launch: %v", err)
+			http.Error(w, "could not load session", http.StatusInternalServerError)
+			return
+		}
+	} else {
+		found, alias, _ = remoteManager.LaunchSession(sourceID, query.Get("agent"), query.Get("id"))
+	}
 	if found == nil {
-		http.Error(w, "session not found", http.StatusNotFound)
+		if sourceID == localSourceID {
+			http.Error(w, "session not found", http.StatusNotFound)
+			return
+		}
+		http.Error(w, "remote host is offline or this session is no longer available", http.StatusConflict)
 		return
 	}
 	handoff, err := readHandoff(w, r)
@@ -262,7 +277,17 @@ func handleLaunch(w http.ResponseWriter, r *http.Request, settingsStore *setting
 		return
 	}
 	mode := query.Get("mode")
-	if err := launch.Terminal(state.Config.Launch.Terminal, found.Agent, found.WorkingDirectory, found.ID, mode, handoff); err != nil {
+	if sourceID != localSourceID {
+		health, testErr := remoteManager.TestAlias(r.Context(), alias)
+		if testErr != nil || health.State != remote.StateOK {
+			http.Error(w, "remote host is offline; wait for it to reconnect", http.StatusConflict)
+			return
+		}
+		err = launch.RemoteTerminal(state.Config.Launch.Terminal, alias, found.Agent, found.WorkingDirectory, found.ID, mode, handoff)
+	} else {
+		err = launch.Terminal(state.Config.Launch.Terminal, found.Agent, found.WorkingDirectory, found.ID, mode, handoff)
+	}
+	if err != nil {
 		log.Printf("launch: %v", err)
 		http.Error(w, "could not launch terminal", http.StatusInternalServerError)
 		return
--- a/collector/cmd/coslash/api.go
+++ b/collector/cmd/coslash/api.go
@@ -260,7 +260,11 @@ func handleLaunch(w http.ResponseWriter, r *http.Request, settingsStore *setting
 			return
 		}
 	} else {
-		found, alias, _ = remoteManager.LaunchSession(sourceID, query.Get("agent"), query.Get("id"))
+		found, alias, err = remoteManager.LaunchSession(sourceID, query.Get("agent"), query.Get("id"))
+		if errors.Is(err, remote.ErrRemoteSessionActive) {
+			http.Error(w, "remote session is already active", http.StatusConflict)
+			return
+		}
 	}
 	if found == nil {
 		if sourceID == localSourceID {
--- a/collector/cmd/coslash/api.go
+++ b/collector/cmd/coslash/api.go
@@ -265,6 +265,10 @@ func handleLaunch(w http.ResponseWriter, r *http.Request, settingsStore *setting
 			http.Error(w, "remote session is already active", http.StatusConflict)
 			return
 		}
+		if errors.Is(err, remote.ErrRemoteSessionUnavailable) {
+			http.Error(w, "remote session details are too large to launch", http.StatusConflict)
+			return
+		}
 	}
 	if found == nil {
 		if sourceID == localSourceID {
--- a/collector/cmd/coslash/api_remote.go
+++ b/collector/cmd/coslash/api_remote.go
@@ -28,6 +28,10 @@ type helperSetupResponse struct {
 	Error   string      `json:"error,omitempty"`
 }
 
+type remoteSetupStatusResponse struct {
+	Phase remote.SetupProgress `json:"phase"`
+}
+
 // helperSetupOutcome is separate from machine collection health. A host may
 // have a healthy SFTP cache while a requested helper action has failed; the
 // setup response must never present that operation as a green success.
@@ -100,7 +104,12 @@ func handleRemoteRetry(w http.ResponseWriter, _ *http.Request, manager *remote.M
 		writeAPIError(w, http.StatusConflict, errCodeRemoteDisabled, "remote disabled")
 		return
 	}
-	health = manager.Retry()
+	var started bool
+	health, started = manager.Retry()
+	if !started {
+		writeAPIError(w, http.StatusTooManyRequests, errCodeRemoteRetryThrottled, "remote refresh is already running; try again shortly")
+		return
+	}
 	w.Header().Set("Content-Type", "application/json")
 	w.WriteHeader(http.StatusAccepted)
 	_ = json.NewEncoder(w).Encode(machineFromHealth(health))
@@ -133,7 +142,9 @@ func handleRemoteHelperSetup(w http.ResponseWriter, request *http.Request, manag
 	}
 	ctx, cancel := context.WithTimeout(request.Context(), 2*time.Minute)
 	defer cancel()
-	setup, err := manager.SetupHelperForAlias(ctx, body.SSHAlias, remote.Consent{Install: body.Install, Upgrade: body.Upgrade})
+	// The single Add remote host action authorizes replacing a verified older
+	// connector as well as an initial install.
+	setup, err := manager.SetupHelperForAlias(ctx, body.SSHAlias, remote.Consent{Install: body.Install, Upgrade: body.Install || body.Upgrade})
 	if errors.Is(err, remote.ErrHelperAliasMismatch) {
 		writeAPIError(w, http.StatusConflict, "remote_alias_mismatch", "SSH alias changed; save settings and test that host before setting up its helper")
 		return
@@ -182,17 +193,6 @@ func handleRemoteStatus(w http.ResponseWriter, _ *http.Request, manager *remote.
 	writeJSON(w, machineFromHealth(health))
 }
 
-func handleRemoteHelperUninstall(w http.ResponseWriter, request *http.Request, manager *remote.Manager) {
-	health := manager.DiagnosticsHealth()
-	if health.SourceID == "" {
-		writeAPIError(w, http.StatusNotFound, errCodeRemoteNotConfigured, "remote not configured")
-		return
-	}
-	ctx, cancel := context.WithTimeout(request.Context(), 2*time.Minute)
-	defer cancel()
-	if err := manager.UninstallHelper(ctx); err != nil {
-		writeAPIError(w, http.StatusBadGateway, "remote_helper_uninstall_failed", "could not uninstall helper; host settings were kept")
-		return
-	}
-	w.WriteHeader(http.StatusNoContent)
+func handleRemoteSetupStatus(w http.ResponseWriter, _ *http.Request, manager *remote.Manager) {
+	writeJSON(w, remoteSetupStatusResponse{Phase: manager.SetupProgress()})
 }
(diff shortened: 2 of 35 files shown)


luhe19001:
Implemented in ac0807a. Remote sessions now expose Start Fresh with handoff and Resume controls, and the launch request preserves the remote source ID. Added local and remote launch-path coverage; frontend tests, lint, formatting, and production build pass.

calvintvu:
add screenshot(s) for any UI changes (i.e. `frontend/src/pages/coslash/components/SessionInspector.tsx`)

collector/internal/remote/helpermanager.go
```diff
@@ -339,7 +387,22 @@ func (manager *Manager) setupHelper(ctx context.Context, expectedAlias string, c
...
+	activeLifecycle := lifecycle
+	if retryableSetupTransportFailure(result) {
+		manager.resetControlMaster(config.SSHAlias)
+		if retryLifecycle, retryErr := factory(config.SSHAlias); retryErr == nil {
+			retryLifecycle.Progress = manager.setSetupProgress
+			result = retryLifecycle.SetupWithLoader(ctx, document, consent, provider.LoadArtifact)
+			activeLifecycle = retryLifecycle
+		}
+	}
+	if result.CanExecute && previousVersion != "" && previousVersion != result.Artifact.Version && activeLifecycle.Remote != nil {
```
calvintvu: P2: Keep the old helper until the new ownership record is durable.

`SetupWithLoader` removes the old helper before `StoreHelperVersion` writes the new version. If that write fails, the old helper is gone. However, local ownership still names the old helper.

This state disables helper collection and makes the recovery data incorrect.

Write the new ownership record before removal. If the write fails, keep the old helper and its ownership record.

collector/cmd/coslash/api_source.go
```diff
@@ -56,6 +59,7 @@ type machineFact struct {
 	HelperProbeState            string                   `json:"helperProbeState,omitempty"`
 	HelperOwnershipRecorded     bool                     `json:"helperOwnershipRecorded"`
 	HelperOwnershipCorrupt      bool                     `json:"helperOwnershipCorrupt"`
+	Refreshing                  bool                     `json:"refreshing,omitempty"`
```
calvintvu, line 61: P1: The client removes this new field, so the retry wait can finish before the refresh finishes.

A manual refresh can keep the `ok`, `stale`, or `error` state when a cached snapshot exists. `decodeMachineFact` does not copy `refreshing`. Also, `retryRemoteRefreshAndWait` only checks the `connecting` state.

As a result, the UI reloads old sessions while collection still runs.

Add `refreshing` to `MachineFact`. Then poll while its value is true.

calvintvu:
PR TLDR
- Keep each remote protocol record below 1 MiB through staged compaction.
- Enable remote Resume and Start Fresh by launching the agent through SSH.
- Preserve handoff context when starting remote sessions.
- Make SSH probing and connector installation time-bounded and recover from stale multiplexed connections.
- Automatically upgrade connector binaries that coSlash can prove it owns.
- Expose better remote health, freshness, retry, and setup-progress information.
- Allow users to forget an unreachable host locally without requiring remote cleanup first.

collector/cmd/coslash/api_source.go
```diff
@@ -45,6 +46,8 @@ type machineFact struct {
 	Complete                    bool                     `json:"complete"`
 	Reason                      *remote.Reason           `json:"reason,omitempty"`
 	LastSuccessAtMs             *int64                   `json:"lastSuccessAtMs,omitempty"`
+	LastCheckedAtMs             *int64                   `json:"lastCheckedAtMs,omitempty"`
+	SessionCount                int                      `json:"sessionCount"`
```
calvintvu: Optional: The frontend does not decode `lastCheckedAtMs` or `sessionCount`. It also calculates the remote session count from the session list. Consider removing these two API fields until a client uses them.

collector/cmd/coslash/api_remote.go
```diff
@@ -182,17 +193,6 @@ func handleRemoteStatus(w http.ResponseWriter, _ *http.Request, manager *remote.
...
-	}
-	ctx, cancel := context.WithTimeout(request.Context(), 2*time.Minute)
-	defer cancel()
-	if err := manager.UninstallHelper(ctx); err != nil {
-		writeAPIError(w, http.StatusBadGateway, "remote_helper_uninstall_failed", "could not uninstall helper; host settings were kept")
-		return
-	}
-	w.WriteHeader(http.StatusNoContent)
+func handleRemoteSetupStatus(w http.ResponseWriter, _ *http.Request, manager *remote.Manager) {
+	writeJSON(w, remoteSetupStatusResponse{Phase: manager.SetupProgress()})
```
calvintvu: Optional: This setup-progress API has no frontend consumer in this PR. The callback, manager state, endpoint, and route add a vertical slice of about 40 lines. Consider deleting the slice until the UI needs it.

calvintvu:
Optional cleanup for unchanged lines:

- Delete `uninstallRemoteHelper` in `frontend/src/pages/coslash/lib/remote-api.ts`. This PR removes its server route, and no caller uses the function.
- Delete the exported `SetupHelper` wrapper in `collector/internal/remote/helpermanager.go`. Only same-package tests use it, so they can call `setupHelper` directly.

These changes remove about ten more lines without changing production behavior.

collector/internal/remote/helpermanager.go
```diff
@@ -221,17 +223,46 @@ func (manager *Manager) InspectHelper() Health {
...
+			if autoUpdate && retryableSetupTransportFailure(result) {
+				manager.resetControlMaster(config.SSHAlias)
+				if retryLifecycle, retryErr := factory(config.SSHAlias); retryErr == nil {
+					result = retryLifecycle.SetupWithLoader(probeCtx, document, consent, provider.LoadArtifact)
+					lifecycle = retryLifecycle
+				}
+			}
+		}
+	}
+	if result.CanExecute && autoUpdate && previousVersion != "" && previousVersion != result.Artifact.Version && lifecycle.Remote != nil {
```
calvintvu: Optional: `Lifecycle.SetupWithLoader` already removes the previous helper after the replacement passes its capability check. This block repeats that removal. The manual setup path contains the same duplicate block at lines 401-405. Removing both blocks saves about ten lines and keeps ownership cleanup in one place.

collector/internal/remote/helpermanager.go
luhe19001: Fixed in 4761189. Lifecycle setup now retains a prior helper; the manager writes the new ownership record first and only then removes the recorded prior helper. A failed ownership write therefore leaves the old helper and ownership intact.

collector/cmd/coslash/api_source.go
luhe19001: Fixed in 4761189. MachineFact now decodes refreshing, and retryRemoteRefreshAndWait polls until both connecting and refreshing are false. Added coverage for an ok cached state while refresh is active.

collector/cmd/coslash/api_source.go
luhe19001: Removed in 4761189. lastCheckedAtMs and sessionCount are no longer emitted until a client consumes them.

collector/cmd/coslash/api_remote.go
luhe19001: Removed in 4761189. The unused setup-progress callback, manager state, endpoint, and route are gone.

collector/internal/remote/helpermanager.go
luhe19001: Resolved in 4761189 by centralizing prior-helper cleanup in the manager after durable ownership storage. Lifecycle setup now leaves prior helpers intact, avoiding premature deletion.

calvintvu (approved):
if deletion of the old helper fails after the new ownership record is stored, the old binary can remain remotely, not a blocker

collector/internal/remote/helpermanager.go
```diff
@@ -357,13 +397,30 @@ func (manager *Manager) setupHelper(ctx context.Context, expectedAlias string, c
 		manager.helperTarget = &helperTarget{path: result.Path, version: result.Artifact.Version, state: result.State, artifact: result.Artifact}
 		manager.helperVersion = result.Artifact.Version
 		manager.helperProbe = helperProbeReady
+		if previousVersion != "" && previousVersion != result.Artifact.Version && activeLifecycle.Remote != nil {
+			if previousPath, pathErr := helperPath(previousVersion); pathErr == nil {
+				_ = activeLifecycle.Remote.RemoveExact(ctx, previousPath)
```
milanshen: **P0 — an SSH round trip now runs while holding `manager.mu`.**

The dedup in `4761189` is right in spirit, but moving the cleanup into the ownership-recorded branch put it inside the critical section: the lock is taken at line 383 with `defer manager.mu.Unlock()`, and `RemoveExact` goes through `SSHLifecycleRemote.withSession`, which opens a real SFTP session.

Every concurrent `ListView` (`/api/sessions`), `remoteStatus`, `Retry` and `TestAlias` blocks on `manager.mu` until that round trip returns — seconds normally, and up to the context deadline if the host stops responding mid-upgrade (the sibling call at line 297 runs under `probeCtx`, which allows 2 minutes on the auto-update path).

Both calls ran outside the lock before this commit, and `setupHelper`'s own comment above notes that `SetupWithLoader` deliberately releases the mutex while uploading — this reintroduces exactly what that design avoided. Capturing what to remove under the lock and doing the removal after it unlocks keeps the dedup without the stall.

collector/internal/remote/manager.go
```diff
@@ -348,6 +359,10 @@ func (manager *Manager) ListView(remoteSinceMs int64) ListResult {
 	if !manager.cfg.Enabled {
 		return ListResult{Health: manager.healthLocked(remoteSinceMs)}
 	}
+	if manager.helperProbe == helperProbeFallback && manager.helperVersion != "" &&
```
milanshen, line 366: **P0 — this re-arm locks the user out of their own Retry setup.**

For a host with recorded ownership, `startHelperDiscoveryLocked` spawns `runHelperDiscovery` with `autoUpdate=true`, which sets `helperSetup = true` for up to `DefaultHelperAutoUpdateTimeout` (2 min). While that flag is set, `setupHelper` returns `ErrHelperSetupInProgress` → 409 *"helper setup is already running for this host"*, and `ApplySettings` rejects an SSH alias change.

With `RemoteRetryInterval` at 3 min and ListView polling roughly every 60s, an offline owned host re-arms this cycle indefinitely: the user is locked out of **Retry setup** for about 2 of every 3 minutes, and the UI gives no indication why the button just failed.

collector/internal/remote/helpermanager.go
```diff
@@ -240,6 +266,14 @@ func (manager *Manager) runHelperDiscovery(ctx context.Context, config settings.
 		return
 	}
 	if err != nil {
+		if errors.Is(err, context.DeadlineExceeded) || sshErrorStderr(err) != "" {
+			manager.lastCheckedAt = int64Ptr(manager.now().UnixMilli())
+			manager.helperProbe = helperProbeFallback
+			status := unavailableHelperStatus(err)
+			manager.helper = &status
+			manager.applyFailureLocked(classifyError(err), sshErrorStderr(err))
```
milanshen: **P1 — a helper-probe failure downgrades collection health that is actually fine.**

`applyFailureLocked` sets `state = StateStale` (or `StateError`), `complete = false`, `errorCopy`, and `failures++`. Those are *collection* health fields, but this call site is the helper discovery probe.

Scenario: SFTP collection succeeds at T (`StateOK`, fresh cards). At T+3min ListView re-arms discovery, whose SSH handshake exceeds the new 7s `DefaultConnectTimeout` on a slow or momentarily loaded host — note collection itself still gets `DefaultDeadline` (3 min), so the probe is far more likely to time out than the refresh that just succeeded. The board flips every remote card to stale and reports an error while SFTP is working.

The shared `failures` counter also means probe timeouts push out the collection retry schedule.

collector/cmd/coslash/api.go
```diff
@@ -238,21 +238,40 @@ func cleanupHandoffs() {
...
+			return
+		}
+	}
 	if found == nil {
-		http.Error(w, "session not found", http.StatusNotFound)
+		if sourceID == localSourceID {
+			http.Error(w, "session not found", http.StatusNotFound)
+			return
+		}
+		http.Error(w, "remote host is offline or this session is no longer available", http.StatusConflict)
```
milanshen, line 278: **P1 — a compacted family is reported to the user as an offline host.**

`compact()` stage 2 sets `Sessions[i].Display = session.Session{}`, so `WorkingDirectory` is empty after `Parsed()`. `LaunchSession`'s `item.WorkingDirectory != ""` guard then skips the session and returns `(nil, "", nil)` — a nil session with a **nil error**, indistinguishable here from a genuinely offline host.

The user sees 409 *"remote host is offline or this session is no longer available"* on a fully connected host and goes off to debug SSH, when the real cause is that their family exceeded `MaxFamilyBytes` and had its display data dropped. Worth distinguishing the two so the message can say the session's details were too large to transfer.

frontend/src/pages/coslash/components/SessionInspector.tsx
```diff
@@ -990,10 +990,16 @@ function InspectorFooter({ detail }: { detail: SessionDetail }) {
...
-        </span>
+      <SheetFooter className="bg-muted flex-row items-center justify-between gap-4 border-t">
+        <div className="flex min-w-0 flex-col">
+          <span className="text-xs">Resume this exact remote session</span>
+          <span className="text-muted-foreground text-xs font-light">
+            Reopens this session in {getVendor(detail.agent).label} on {detail.sourceLabel} via SSH.
+          </span>
+        </div>
+        <div className="flex shrink-0 items-center gap-2">
+          <ResumeSessionButton detail={detail} />
```
milanshen, line 1003: **P1 — Resume/Start fresh are never gated on the host being reachable.**

`ResumeSessionButton` here (and `StartNewSessionButton` at line 422) render unconditionally for remote sessions — no check on machine state and none on `workingDirectory` being present. With the host offline the button looks enabled, and the failure only surfaces as a 409 toast after the click, once `TestAlias` fails server-side.

`docs/troubleshooting.md` in this same PR states: *"They are disabled while the host is offline; wait for it to reconnect, then try again."* The shipped UI does not match that.

collector/internal/remote/helpermanager.go
luhe19001: Fixed in 8990c14. The manager captures the old helper path while holding its mutex, then releases the mutex before the SFTP RemoveExact call in both automatic and manual setup paths. Ownership is still written before cleanup.

collector/internal/remote/manager.go
luhe19001: Fixed in 8990c14. Automatic updates now use separate automatic-setup state, so they no longer return ErrHelperSetupInProgress to a user setup request or block an SSH-alias settings update.

collector/internal/remote/helpermanager.go
luhe19001: Fixed in 8990c14. A helper-discovery transport failure now updates only helper-probe state and a dedicated helper-probe retry time. It preserves collection health, collection backoff, and fresh cards.

collector/cmd/coslash/api.go
luhe19001: Fixed in 8990c14. LaunchSession now returns a distinct error for a matching remote session whose compacted facts lack its working directory, and the API tells the user that its details are too large to launch. Remote board records also expose launchable for the client.

frontend/src/pages/coslash/components/SessionInspector.tsx
luhe19001: Fixed in 8990c14. Resume and Start fresh with handoff are disabled for stale remote sessions and for sessions whose remote facts are not launchable.

PR #143, opened 2026-09-03
Improve remote monitoring and terminal actions

luhe19001:
## Context

Remote monitoring should make per-machine activity and normal offline states understandable, while SSH setup should be a concise guided flow. Live remote sessions should also support the same terminal actions as local sessions.

## Changes

- Replace the global activity summary with per-machine status chips and activity counts.
- Show checking, offline, connection freshness, cached-data age, and actionable setup-failure states.
- Simplify SSH setup to one add/install/verify flow with concise progress, retry, and local-only removal.
- Auto-save regular settings changes and use a Close-only settings footer.
- Enable remote Resume and Start fresh with handoff actions when the SSH host is live, with useful unavailable-state guidance.
- Correct remote Codex handoff quoting so the fresh session loads its handoff.

## Test

- `go test ./...` — passed
- `go vet ./...` — passed
- `gofmt -l ./cmd ./internal` — passed
- `npm test` — passed
- `npm run format:check` — passed
- `npm run build` — passed
- Manual: checked the machine activity display, SSH monitoring settings flow, and remote handoff command.

### Screenshots

SSH monitoring settings


collector/internal/remote/manager.go
```diff
@@ -432,6 +437,27 @@ func (manager *Manager) DiagnosticsHealth() Health {
...
+func (manager *Manager) LaunchSession(sourceID, agent, sessionID string) (*session.Session, string, error) {
+	manager.mu.Lock()
+	defer manager.mu.Unlock()
+	if manager.cfg == nil || !manager.cfg.Enabled || manager.cfg.ID != sourceID ||
+		(manager.state != StateOK && manager.state != StateLimited) {
+		return nil, "", nil
+	}
+	for _, item := range manager.sessions {
+		if item.Agent == agent && item.ID == sessionID && item.WorkingDirectory != "" {
+			if item.Status != nil && *item.Status == "busy" {
```
calvintvu: **Blocking (P1):** `LaunchSession` rejects every launch for a busy session. The UI disables only Resume. As a result, Start fresh remains enabled but returns HTTP 409. If `mode=resume`, apply this restriction.

frontend/src/pages/coslash/hooks/use-sessions.ts
```diff
@@ -235,17 +251,45 @@ export function useSessions({ localWindow, remoteWindow }: SessionsQuery) {
...
+  useEffect(() => {
+    if (!remoteRefreshing) return;
+    let cancelled = false;
+    let timer: ReturnType<typeof setTimeout> | undefined;
+    const poll = () => {
+      void remoteStatus()
+        .then((machine) => {
+          if (cancelled) return;
+          setMachines((current) =>
+            current.map((item) => (item.sourceId === machine.sourceId ? machine : item)),
```
calvintvu: **Blocking (P2):** This poll updates only `machines`. It does not reload `/api/sessions` after the remote refresh ends. New or changed remote sessions stay stale for up to 60 seconds. Reload the session list after the poll reaches a final state.

collector/internal/remote/helpermanager.go
```diff
@@ -221,19 +223,46 @@ func (manager *Manager) InspectHelper() Health {
...
+			if autoUpdate && retryableSetupTransportFailure(result) {
+				manager.resetControlMaster(config.SSHAlias)
+				if retryLifecycle, retryErr := factory(config.SSHAlias); retryErr == nil {
+					result = retryLifecycle.SetupWithLoader(probeCtx, document, consent, provider.LoadArtifact)
+					lifecycle = retryLifecycle
+				}
+			}
+		}
+	}
+	if result.CanExecute && autoUpdate && previousVersion != "" && previousVersion != result.Artifact.Version && lifecycle.Remote != nil {
```
calvintvu: **Blocking (P2):** This code removes the previous helper before it saves the new ownership record. If `StoreHelperVersion` fails, the usable helper is gone and the local record still names the old version. Save the ownership record first. Then remove the previous helper.

collector/internal/launch/launch.go
```diff
@@ -75,6 +76,33 @@ func Terminal(terminal, agent, workingDirectory, sessionID, mode, handoff string
...
+// RemoteTerminal opens the selected local terminal and runs an agent CLI on a
+// configured SSH host.
+func RemoteTerminal(terminal, alias, agent, workingDirectory, sessionID, mode, handoff string) error {
+	if alias == "" {
+		return errors.New("launch: SSH alias is required")
+	}
+	if workingDirectory == "" {
+		return fmt.Errorf("launch: session has no working directory")
+	}
+	adapter, err := terminalFor(terminal)
```
calvintvu, line 87: **Optional:** `Terminal` and `RemoteTerminal` repeat terminal selection, platform checks, availability checks, and terminal startup. Move this common code into one helper. Keep only remote-command construction in this function.

collector/internal/remote/helpermanager.go
```diff
@@ -221,19 +223,46 @@ func (manager *Manager) InspectHelper() Health {
...
+	if autoUpdate {
+		timeout = DefaultHelperAutoUpdateTimeout
+	}
+	probeCtx, cancel := context.WithTimeout(ctx, timeout)
 	defer cancel()
 	manager.mu.Lock()
-	provider, factory := manager.releaseProvider, manager.lifecycleFactory
+	provider, factory, previousVersion := manager.releaseProvider, manager.lifecycleFactory, manager.helperVersion
 	manager.mu.Unlock()
 	document, err := provider.LoadMetadata(probeCtx)
```
calvintvu, line 243: **Optional:** This function and `setupHelper` repeat metadata loading, lifecycle creation, transport retry, helper removal, and ownership storage. Use one internal lifecycle runner for both paths.

frontend/src/pages/coslash/hooks/use-sessions.ts
```diff
@@ -235,17 +251,45 @@ export function useSessions({ localWindow, remoteWindow }: SessionsQuery) {
...
+          );
+          if (machine.refreshing || machine.state === 'connecting' || machine.reason === 'initial_refresh') {
+            timer = setTimeout(poll, REMOTE_REFRESH_POLL_INTERVAL_MS);
+          }
+        })
+        .catch(() => {
```
calvintvu: **Optional:** This polling loop overlaps `retryRemoteRefreshAndWait`. Use one function that polls until the remote operation reaches a final state.

frontend/src/pages/coslash/components/MachineActivity.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+function connectorFailed(machine: MachineFact) {
+  return (
+    machine.sourceId !== LOCAL_SOURCE_ID &&
+    machine.helper?.compatible === false &&
+    machine.helper.reason != null
+  );
+}
+
+function connectorFailureCopy(machine: MachineFact) {
```
calvintvu, line 52: **Optional:** `MachinesSettingsSection` contains the same helper-failure formatting and similar machine-state text. Put the shared text mapping next to `MachineFact`.

frontend/src/pages/coslash/components/SettingsDialog.tsx
```diff
@@ -201,39 +201,81 @@ export function SettingsDialog({
     response ? initialSettingsDraft(response) : null,
   );
   const [disclosureOpen, setDisclosureOpen] = useState(false);
-  const [remoteOwnershipAction, setRemoteOwnershipAction] = useState<RemoteOwnershipAction | null>(null);
+  const [remoteOperationInProgress, setRemoteOperationInProgress] = useState(false);
+  const initializedForOpen = useRef(false);
+  const saveChain = useRef(Promise.resolve());
   const isFirstRun = requiresFirstRunConsent(response);
   const requiresConsent = mode === 'synthesis-consent' && isFirstRun;
   const showFullSettings = mode === 'full-settings';
```
calvintvu: **Optional:** `showFullSettings`, `setRemoteBusy`, and `close` only rename one-line expressions. Remove these aliases and use the expressions at their call sites.

calvintvu (changes requested):
I found three blocking behavior errors. The other comments are optional simplifications. Rebase before addressing comments.

collector/internal/remote/manager.go
luhe19001: Fixed: busy state now rejects Resume only; Start fresh remains available.

frontend/src/pages/coslash/hooks/use-sessions.ts
luhe19001: Fixed: once remote polling reaches a final state, the session list reloads immediately.

collector/internal/remote/helpermanager.go
luhe19001: Fixed: ownership is stored successfully before removing the prior helper.

collector/internal/launch/launch.go
luhe19001: Refactored: local and remote launch paths now share terminal selection, availability checks, and opening.

collector/internal/remote/helpermanager.go
luhe19001: Kept discovery and consented setup separate because their lifecycle/progress semantics differ; both retain the corrected ownership ordering.

frontend/src/pages/coslash/hooks/use-sessions.ts
luhe19001: Refactored: both flows now use the shared wait-for-final-remote-refresh poller.

frontend/src/pages/coslash/components/MachineActivity.tsx
luhe19001: Kept the contextual machine copy local to each surface; the failure formatting remains intentionally minimal.

frontend/src/pages/coslash/components/SettingsDialog.tsx
luhe19001: Simplified: removed the one-line settings aliases.

collector/internal/remote/helpermanager.go
```diff
@@ -297,6 +292,11 @@ func (manager *Manager) runHelperDiscovery(ctx context.Context, config settings.
 			} else {
 				manager.helperVersion = result.Artifact.Version
 				manager.helperProbe = helperProbeReady
+				if autoUpdate && previousVersion != "" && previousVersion != result.Artifact.Version && lifecycle.Remote != nil {
+					if previousPath, pathErr := helperPath(previousVersion); pathErr == nil {
+						_ = lifecycle.Remote.RemoveExact(probeCtx, previousPath)
```
calvintvu: **[P1]:** `RemoveExact` opens an SSH/SFTP session while `manager.mu` is locked. Board refreshes, settings writes, and remote launches wait for this request. Save the ownership record under the lock. Then unlock before you remove the old helper.

luhe19001: Fixed: the ownership record is saved under the manager lock, then the lock is released before prior-helper cleanup performs remote I/O.

frontend/src/pages/coslash/components/SettingsDialog.tsx
```diff
@@ -514,14 +546,15 @@ export function SettingsDialog({
...
           <div className="flex items-center gap-2">
-            {!requiresConsent && (
-              <Button variant="outline" onClick={close} disabled={isSaving}>
-                Cancel
+            {requiresConsent ? (
+              <Button onClick={() => void save()} disabled={!draft || isSaving || !synthesisBackendAvailable}>
+                Save settings
+              </Button>
+            ) : (
+              <Button variant="outline" onClick={close} disabled={remoteOperationInProgress}>
```
milanshen: **P0 — this calls `window.close()`, not the dialog's close.**

The local `const close = () => { setTheme(...); onOpenChange(false); }` was deleted in this commit, but the footer still passes `close`. The identifier appears exactly once in this file — here — with no declaration in module scope, so it resolves to the DOM global `close` (`declare function close(): void` in `lib.dom.d.ts`). That is why `tsc -b` and oxlint are both green.

At runtime, clicking the footer button calls `window.close()`. The browser blocks that for a tab it did not script-open, so the Settings dialog just never closes. This is now the *only* footer action, so it is the primary dismissal path.

```suggestion
              <Button variant="outline" onClick={() => onOpenChange(false)} disabled={remoteOperationInProgress}>
```

frontend/src/pages/coslash/hooks/use-sessions.ts
```diff
@@ -235,17 +249,38 @@ export function useSessions({ localWindow, remoteWindow }: SessionsQuery) {
...
       controller.abort();
     };
   }, [localWindow, remoteWindow, retryCount]);
 
+  const remoteRefreshing = remoteRefreshInProgress(machines);
+
+  useEffect(() => {
+    if (!remoteRefreshing) return;
+    let cancelled = false;
+    void waitForRemoteRefresh()
```
milanshen: **P0 — these pollers are uncancellable and accumulate.**

`waitForRemoteRefresh` (`remote-api.ts:55`) loops `while (remoteRefreshInProgress(current))`, hitting `/api/remote/status` every 400ms with no abort signal, no timeout and no iteration cap. The cleanup below sets `cancelled = true`, which suppresses the `setMachines`/`setRetryCount` calls but cannot stop that loop.

Two consequences:

1. A single refresh on a real host takes ~90s, so one cycle is roughly 225 status requests.
2. Every `remoteRefreshing` false→true transition starts a *new* loop while earlier ones are still polling. They never stop, so the request rate grows for as long as the page is open.

The effect already has the shape for this — thread the `cancelled` flag (or the `AbortController` from the load effect) into `waitForRemoteRefresh` and break the loop on it.

collector/internal/remote/helpermanager.go
```diff
@@ -420,6 +422,11 @@ func (manager *Manager) setupHelper(ctx context.Context, expectedAlias string, c
 		manager.helperTarget = &helperTarget{path: result.Path, version: result.Artifact.Version, state: result.State, artifact: result.Artifact}
 		manager.helperVersion = result.Artifact.Version
 		manager.helperProbe = helperProbeReady
+		if previousVersion != "" && previousVersion != result.Artifact.Version && activeLifecycle.Remote != nil {
+			if previousPath, pathErr := helperPath(previousVersion); pathErr == nil {
+				_ = activeLifecycle.Remote.RemoveExact(ctx, previousPath)
```
milanshen: **P0 — an SSH round trip while holding `manager.mu`.**

This commit fixes exactly this problem in `runHelperDiscovery` — it captures `removePreviousPath`, unlocks, then calls `RemoveExact` at line 314 — but applies the opposite change here: the pre-lock block was deleted and the call re-added *inside* the `manager.mu.Lock()` / `defer manager.mu.Unlock()` region opened just above.

`RemoveExact` goes through `SSHLifecycleRemote.withSession` and opens a real SFTP session. After any helper upgrade, every concurrent `ListView` (`/api/sessions`), `remoteStatus`, `Retry` and `TestAlias` blocks on `manager.mu` until that round trip returns — seconds normally, and up to the context deadline if the host stops responding mid-upgrade.

The capture-then-unlock pattern you already used at line 314 applies here unchanged.

frontend/src/pages/coslash/lib/remote-api.ts
```diff
@@ -109,9 +121,13 @@ export async function remoteStatus(): Promise<MachineFact> {
   return decodeMachineFact(await response.json());
 }
 
-export async function uninstallRemoteHelper(): Promise<void> {
-  const response = await apiFetch('/api/remote/helper/uninstall', { method: 'POST' });
-  if (response.ok) return;
-  const apiError = await readApiError(response);
-  throw new Error(apiError?.error || `Helper uninstall failed (${response.status})`);
+export async function remoteSetupProgress(): Promise<SetupProgress> {
```
milanshen: **P1 — this endpoint no longer exists on the base branch.**

This PR is based on `ac0807a`, but `helu/remote-monitoring-setup-ux` is now at `4761189` ("Address remote monitoring review feedback"), which deleted the whole setup-progress mechanism: the `GET /api/remote/setup/status` route, the `SetupProgress` type, `Lifecycle.Progress`, and `manager.SetupProgress()`.

This PR re-adds the consumer — `remoteSetupProgress()` polled every 500ms from `installConnector`. After PR #142 merges, `helpermanager.go` and `lifecycle.go` conflict, and resolving those by taking the deletion leaves this poll hitting a 404. `MachinesSettingsSection`'s `.catch(() => {})` swallows it, so the install message silently freezes on "Installing connector…" for the entire install with no error anywhere.

Worth rebasing onto the current base head and settling whether setup progress is being kept or dropped — right now the two PRs disagree.

frontend/src/pages/coslash/CoslashPage.tsx
```diff
@@ -466,8 +475,11 @@ export function CoslashPage() {
               <LoadingSpinner isLoading={isLoading}>
                 <SessionsStats
                   sessions={sessionsForVendor}
```
milanshen, line 479: **P1 — the machine chips count only filter-visible sessions.**

`sessionsForVendor` is already narrowed by both the machine filter and the vendor filter, and `SessionsStats` forwards it straight to `MachineActivity`, whose `activityFor(sessions, sourceId)` derives each chip's active/waiting counts from it.

Set the machine filter to **Local** while agent-box has 3 busy sessions: every remote session is filtered out, so the agent-box chip reads "0 active · 0 waiting" while the host is connected and busy. Picking a single vendor does the same thing per vendor.

These chips replaced the global activity summary, so they should read from `sessionsInWindow` (unfiltered) rather than the filtered list — otherwise the filter hides the very activity the chip exists to surface.

frontend/src/pages/coslash/components/MachinesSettingsSection.tsx
```diff
@@ -1,386 +1,265 @@
...
-        const status = await remoteStatus();
-        if (cancelled) return;
-        setHostStatus(status);
-        if (status.helperProbeState === 'probing') timer = setTimeout(() => void load(), 400);
-      } catch {
-        if (!cancelled) setHostStatus(null);
-      }
-    };
-    void load();
+    void remoteStatus()
```
milanshen: **P1 — the host status is fetched once and never refreshes.**

The previous implementation re-polled every 400ms while `helperProbeState === 'probing'`; this is now a single one-shot fetch keyed on `remote?.id`.

Open Settings while collection is running: the fetch returns `refreshing: true`, the row renders **Checking**, and nothing ever fetches again — it stays "Checking" until the user closes and reopens the dialog, long after the refresh finished.

Compounding it, `addRemoteHost` sets `draft.remote` to `{ sshAlias, enabled: true }` with no server-assigned `id`, and this effect early-returns on `!remote?.id`. So for a host the user just added, the status fetch is disabled for the remainder of the dialog session.

collector/internal/remote/manager.go
```diff
@@ -448,7 +448,7 @@ func (manager *Manager) LaunchSession(sourceID, agent, sessionID string) (*sessi
 	}
 	for _, item := range manager.sessions {
 		if item.Agent == agent && item.ID == sessionID && item.WorkingDirectory != "" {
-			if item.Status != nil && *item.Status == "busy" {
+			if mode == "resume" && item.Status != nil && *item.Status == "busy" {
```
milanshen: **P1 — the busy guard hangs on a string literal.**

`launch.ResumeSession` is defined as `"resume"`, but this compares against a literal, and `handleLaunch` forwards `query.Get("mode")` with no validation. This comparison is the only thing preventing a user from resuming an agent session that is mid-turn.

Any drift — the launch package renaming `ResumeSession`, or a caller sending `"Resume"` — makes the condition silently false with no compile error, and a second CLI attaches to a busy remote session.

Referencing `launch.ResumeSession`, or validating `mode` against the known set before calling `LaunchSession`, makes the coupling explicit and breaks loudly instead of quietly.

frontend/src/pages/coslash/components/SettingsDialog.tsx
luhe19001: Fixed: the footer now calls the dialog onOpenChange handler directly.

frontend/src/pages/coslash/hooks/use-sessions.ts
luhe19001: Fixed: remote refresh polling now receives an AbortSignal and stops on effect cleanup.

collector/internal/remote/helpermanager.go
luhe19001: Fixed: both helper cleanup paths capture the remote target, unlock, then perform remote I/O.

frontend/src/pages/coslash/lib/remote-api.ts
luhe19001: Rebased onto the current base and removed the obsolete setup-progress client polling.

frontend/src/pages/coslash/CoslashPage.tsx
luhe19001: Fixed: machine chips now receive unfiltered sessions from the active time window.

frontend/src/pages/coslash/components/MachinesSettingsSection.tsx
luhe19001: Fixed: settings polls status through refresh completion and keys it on the saved SSH alias.

collector/internal/remote/manager.go
luhe19001: Fixed: the busy-session guard now compares against launch.ResumeSession.

frontend/src/pages/coslash/components/MachinesSettingsSection.tsx
```diff
@@ -1,386 +1,238 @@
...
-        if (status.helperProbeState === 'probing') timer = setTimeout(() => void load(), 400);
-      } catch {
-        if (!cancelled) setHostStatus(null);
-      }
+    const poll = () => {
+      void remoteStatus()
+        .then((status) => {
+          if (cancelled) return;
+          setMachine(status);
+          if (status.refreshing || status.state === 'connecting' || status.reason === 'initial_refresh') {
```
calvintvu: [P2] Continue polling during connector discovery

`remoteStatus()` can return `helperProbeState: "probing"` while `refreshing` is false and `state` is not `connecting`. This condition stops polling too soon. The screen can miss the final connector status until the user reopens Settings. Continue polling while `status.helperProbeState === "probing"`.

frontend/src/pages/coslash/components/MachineActivity.tsx
```diff
@@ -0,0 +1,132 @@
...
+  };
+}
+
+function isChecking(machine: MachineFact) {
+  return machine.refreshing || machine.reason === 'initial_refresh' || machine.state === 'connecting';
+}
+
+function machineCopy(machine: MachineFact, checking: boolean) {
+  if (machine.sourceId === LOCAL_SOURCE_ID) return 'This Mac is up to date.';
+  const lastChecked = machine.lastCheckedAtMs == null ? 'not yet' : formatTimeAgo(machine.lastCheckedAtMs);
```
calvintvu, line 21: [P2] Return the status fields that this UI uses

The API does not return `lastCheckedAtMs` or `sessionCount`. This text always shows “Last checked not yet.” The empty-session text also does not appear. Add both fields to `machineFact` and copy them from `remote.Health`.

frontend/src/pages/coslash/components/MachinesSettingsSection.tsx
luhe19001: Fixed: Settings continues polling while helper discovery reports probing.

frontend/src/pages/coslash/components/MachineActivity.tsx
luhe19001: Fixed: machine facts now include lastCheckedAtMs and sessionCount from remote health.

luhe19001:
Addressed the requested behavior fixes: busy remote sessions now block Resume only, sessions reload when remote refresh completes, and helper ownership is stored before prior-helper cleanup. I also fixed the follow-up polling, dialog-close, activity-count, and lock-scope issues, then rebased onto the current base. Please re-review when convenient.

frontend/src/pages/coslash/components/SessionInspector.tsx
```diff
@@ -352,16 +354,32 @@ function LaunchError({ message }: { message: string | null }) {
...
+        <TooltipTrigger asChild>
+          <span tabIndex={0}>{children}</span>
+        </TooltipTrigger>
+        <TooltipContent>{hint}</TooltipContent>
+      </Tooltip>
+    </TooltipProvider>
+  );
+}
+
+function ResumeSessionButton({ detail, disabledHint }: { detail: SessionDetail; disabledHint?: string }) {
```
calvintvu, line 371: **[P2]** Disable Resume for an active remote session

The UI shows “This session is already active,” but Resume stays enabled. A click then causes a backend conflict. Include the remote busy state in `disabled`. Keep “Start fresh” enabled.

calvintvu (approved):

frontend/src/pages/coslash/components/SessionInspector.tsx
luhe19001: will postpone this for further UI refinement.

frontend/src/pages/coslash/components/SettingsDialog.tsx
```diff
@@ -418,7 +462,7 @@ export function SettingsDialog({
                       onClick={() => {
                         const theme = draft.appearance.theme === 'dark' ? 'light' : 'dark';
                         setTheme(theme);
```
luhe19001: **Theme toggled before save succeeds** · medium

The dark-mode switch calls `setTheme` immediately and then auto-saves via `saveChange`. If the PUT fails, the UI stays on the new theme while `settings.json` and the loaded settings response still hold the previous value, so appearance diverges from persisted settings until a full reload or manual repair.

Consider applying the theme only after `saveChange` / `saveSettings` succeeds (same pattern already used inside `saveChange`), or rolling the theme back on save failure.

PR #146, opened 2026-09-03
Show OpenCode session client in badges

calvintvu:
## Context

OpenCode session badges previously showed the selected agent role, which does not identify the surface used to run a session. Surface labels now distinguish Desktop and CLI while preserving the existing modality labels for other agents.

## Changes

- Record the OpenCode client surface for session activity and display `Desktop` or `CLI` in session badges.
- Omit the OpenCode modality when no client record exists, rather than showing the misleading Build/Plan role.
- Remove client markers when an OpenCode session is deleted.

Note: old inactive opencode sessions created before this change will show no modality badge until that session is reopened. retroactively iterating through all old sessions and updating the entrypoint will add significant complexity.

## Test

- `GOCACHE=/private/tmp/coslash-go-cache go test ./...` — passed
- `GOCACHE=/private/tmp/coslash-go-cache go build ./...` — passed
- `npm test` — passed
- `npm run build` — passed

### Screenshots

- [x] Before — OpenCode only distinguishes between Build/Plan mode.
Screenshot 2026-09-03 at 11 30 49

- [x] After — OpenCode can distinguish between sessions created in either CLI or Desktop app.
Screenshot 2026-09-03 at 11 18 20
Screenshot 2026-09-03 at 11 16 53

### DIFF
--- a/collector/internal/vendors/opencode/coslash-plugin.js
+++ b/collector/internal/vendors/opencode/coslash-plugin.js
@@ -58,7 +58,7 @@ async function clearSession(properties) {
   )
 }
 
-export const CoslashWaitingPlugin = async () => ({
+export const CoslashPlugin = async () => ({
   event: ({ event }) => {
     lifecycle = lifecycle.then(async () => {
       if (["session.created", "session.updated", "message.updated"].includes(event?.type)) {
--- a/collector/cmd/coslash/main.go
+++ b/collector/cmd/coslash/main.go
@@ -82,8 +82,8 @@ func main() {
 	}
 
 	settingsStore := settings.Open()
-	if err := opencode.EnsureWaitingPlugin(); err != nil {
-		log.Printf("install OpenCode status plugin: %v", err)
+	if err := opencode.EnsurePlugin(); err != nil {
+		log.Printf("install OpenCode coSlash plugin: %v", err)
 	}
 	settingsState := settingsStore.State()
 	var runner synthesis.Runner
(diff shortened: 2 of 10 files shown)


luhe19001 (commented):
Bugbot review on the PR branch — one finding below for the author to address.

collector/internal/vendors/opencode/coslash-plugin.js
```diff
@@ -4,9 +4,26 @@ import os from "node:os"
 import path from "node:path"
 
 const directory = path.join(os.homedir(), ".coslash", "opencode-permissions")
+const clientDirectory = path.join(os.homedir(), ".coslash", "opencode-clients")
 const requestsBySession = new Map()
 let lifecycle = Promise.resolve()
 
+async function recordClient(properties) {
+  if (typeof properties.sessionID !== "string") return
```
luhe19001, line 12: **[medium] Plugin ignores `info` session id**

`recordClient` and `clearClient` only accept top-level `properties.sessionID`, but OpenCode `session.created`, `session.updated`, and `session.deleted` events publish the session id on `properties.info.id`. On those events the handlers no-op, so:

1. Client state is not removed when a session is deleted
2. Client labels may never appear on OpenCode versions where `message.updated` also omits top-level `sessionID` and only exposes it on `properties.info.sessionID`

Please resolve the session id from both shapes (e.g. top-level `sessionID` and `properties.info.id` / `properties.info.sessionID`) before writing or clearing client records.

calvintvu: Checked the current OpenCode event schemas: `session.created`, `session.updated`, `session.deleted`, and `message.updated` all define `sessionID` at the top level alongside `info`. The plugin therefore reads the documented identifier for each event, including deletion, and `info.id` / `info.sessionID` fallback handling is not needed for the supported schema.

luhe19001 (approved):

PR #153, opened 2026-09-03
feat(snapshot): export resolved commit object IDs

luhe19001:
## Summary
- Export optional `session.commitShas`: verified, locally resolved Git object IDs alongside the existing human-facing `session.commits` subjects.
- Keep commit subjects and identifiers deliberately separate: subjects remain display text and are never parsed as hashes.
- Emit at most 200 lowercase, full SHA-1 (40-character) or SHA-256 (64-character) object IDs.

## Why
Coslash can already report commit subjects, but a subject is not a dependable commit identity: it can be duplicated, changed, or fail to identify a commit precisely. This change gives the snapshot a deterministic, machine-usable reference to commits actually present in the local repository. It preserves the useful human-readable commit list while making the export ready for precise downstream correlation.

The collector never guesses: an observed hash is included only after it resolves against local repository history to a full object ID. If it cannot be resolved, the subject may remain for people, but no identifier is exported.

## Compatibility and rollout
`session.commitShas` is additive and optional. Consumers that reject unknown snapshot fields must accept this property before collectors begin sending it.

## Verification
- `go test -count=1 ./...` from `collector`
- `go vet ./...` from `collector`
- `git diff --check`


collector/snapshot/v1/snapshot.go
```diff
@@ -108,8 +111,11 @@ type Session struct {
 	Todos            []Todo     `json:"todos"`
 	FileEdits        []FileEdit `json:"fileEdits"`
 	Commits          []string   `json:"commits"`
-	Git              *GitDrift  `json:"git,omitempty"`
-	Subagents        []Subagent `json:"subagents"`
+	// CommitSHAs is optional for the additive v1 rollout. When supplied, each
+	// value is a resolved lowercase full Git object ID (SHA-1 or SHA-256).
+	CommitSHAs []string   `json:"commitShas,omitempty"`
```
calvintvu, line 115: **P1.** `Decode` calls `DisallowUnknownFields`, and the hub runs this same validator. Upgrade the hub before the collectors ship, or hold this field behind a collector flag.

collector/snapshot/v1/snapshot.go
```diff
@@ -108,8 +111,11 @@ type Session struct {
 	Todos            []Todo     `json:"todos"`
 	FileEdits        []FileEdit `json:"fileEdits"`
 	Commits          []string   `json:"commits"`
-	Git              *GitDrift  `json:"git,omitempty"`
-	Subagents        []Subagent `json:"subagents"`
+	// CommitSHAs is optional for the additive v1 rollout. When supplied, each
```
calvintvu, line 112: **P1.** Add one sentence here: `commitShas` has no positional relation to `commits`. A consumer that reads the two arrays by index attributes a SHA to the wrong subject.

collector/snapshot/v1/schema.json
```diff
@@ -142,6 +142,7 @@
         "todos": {"type": "array", "maxItems": 200, "items": {"$ref": "#/$defs/todo"}},
         "fileEdits": {"type": "array", "maxItems": 2000, "items": {"$ref": "#/$defs/fileEdit"}},
         "commits": {"type": "array", "maxItems": 200, "items": {"type": "string", "maxLength": 2048, "x-maxUtf8Bytes": 2048}},
+        "commitShas": {"type": "array", "maxItems": 200, "items": {"type": "string", "pattern": "^(?:[0-9a-f]{40}|[0-9a-f]{64})$"}},
```
calvintvu: **P2.** Add `"uniqueItems": true` and `"minItems": 1` to this entry. `Decode` rejects duplicates and rejects an empty array, but the schema accepts both.

collector/snapshot/v1/internal/fixturegen/main.go
```diff
@@ -112,13 +112,14 @@ func codex() snapshotv1.Snapshot {
...
-			FileEdits: []snapshotv1.FileEdit{{Path: "collector/snapshot/v1/snapshot.go", Additions: 100, Edits: 1, IsNew: true}},
-			Commits:   []string{"feat: add snapshot v1 contract"},
-			Subagents: []snapshotv1.Subagent{},
+			Counts:     snapshotv1.Counts{EditedFiles: 2, Turns: 3, ToolUses: 4, Commands: 2},
+			Usage:      snapshotv1.Usage{Models: []snapshotv1.ModelUsage{{Model: "gpt-5", InputTokens: 1000, OutputTokens: 200, EstimatedCostMicroUSD: 250000}}, EstimatedCostMicroUSD: 250000, UnpricedModels: []string{}},
+			Digest:     []snapshotv1.Digest{{Turn: 1, Category: "user", Description: "Requested a bounded export contract."}},
+			Todos:      []snapshotv1.Todo{{Text: "Run compatibility fixtures", Done: true}},
+			FileEdits:  []snapshotv1.FileEdit{{Path: "collector/snapshot/v1/snapshot.go", Additions: 100, Edits: 1, IsNew: true}},
+			Commits:    []string{"feat: add snapshot v1 contract"},
+			CommitSHAs: []string{strings.Repeat("a", 40)},
```
calvintvu, line 121: **P2.** Add `MaxCommitSHAItems` to `boundaryItems` here and to the table in `boundary_test.go:121`. Every other collection has an exact-limit and a limit-plus-one case, and this one has neither.

collector/internal/session/commits.go
```diff
@@ -103,17 +103,34 @@ type repositoryCommit struct {
...
+// CommitFacts keeps human commit subjects and machine Git object IDs on
+// separate paths. Hashes are emitted only when an observed value resolves
+// against local repository history to a full object ID.
+type CommitFacts struct {
+	Subjects []string
+	SHAs     []string
+}
+
 // NewCommitReconciler caches repository history for one reconciliation pass.
 func NewCommitReconciler() func([]CommitObservation, string, *string) []string {
```
calvintvu: **Optional.** `NewCommitReconciler` and `reconcileCommits` now have no callers, so delete both. `ReconcileCommitFacts` has no caller outside this file, so unexport it.

collector/snapshot/v1/snapshot.go
```diff
@@ -517,6 +536,18 @@ func validateSession(s Session) error {
 	return nil
 }
 
+func validCommitSHA(value string) bool {
```
calvintvu: **Optional.** Line 302 already validates lowercase hex with `hex.DecodeString` and `strings.ToLower`. Use the same idiom here and delete these nine lines.

collector/snapshot/v1/snapshot.go
```diff
@@ -21,28 +21,31 @@ import (
...
+	MaxGoalBytes             = 16 * 1024
+	MaxPromptBytes           = 16 * 1024
+	MaxDigestItems           = 200
+	MaxDigestTextBytes       = 4 * 1024
+	MaxTodoItems             = 200
+	MaxTodoTextBytes         = 2 * 1024
+	MaxFileEditItems         = 2000
+	MaxCommitItems           = 200
+	MaxCommitTextBytes       = 2 * 1024
+	// Commit SHAs are deterministic Git object locators, intentionally
```
calvintvu: **Optional.** This comment splits the alignment run of gofmt and re-indents 22 untouched constants. Delete it, and the const block diff becomes one added line.

calvintvu (commented):
first pass review

collector/snapshot/v1/snapshot.go
luhe19001: Reused the existing `hex.DecodeString` plus lowercase validation idiom inline and removed `validCommitSHA`.

collector/internal/session/commits.go
luhe19001: Applied the cleanup: removed the unused `NewCommitReconciler` and `reconcileCommits`, and made the repository-backed commit-facts helper private.

collector/snapshot/v1/internal/fixturegen/main.go
luhe19001: Added `MaxCommitSHAItems` to both `boundaryItems` and the exact-limit/limit-plus-one boundary test table, with unique valid full object IDs. Regenerated the boundary fixture and manifest.

collector/snapshot/v1/snapshot.go
luhe19001: The Hub was upgraded first in centauri-ai/coslash-server@913e33b: its strict snapshot validator now accepts `commitShas`, and ingest projects the field explicitly. That commit is on the Hub `main` branch; current Hub CI and release-image publication also pass. No collector rollout flag is needed. https://github.com/centauri-ai/coslash-server/commit/913e33ba37d450cf9fda67dc161923c568479f22

collector/snapshot/v1/schema.json
luhe19001: Added `minItems: 1` and `uniqueItems: true` to the `commitShas` schema entry.

collector/snapshot/v1/snapshot.go
luhe19001: Removed the interrupting comment so gofmt restores the single aligned const block and the diff is limited to the new constant.

collector/snapshot/v1/snapshot.go
luhe19001: Added the explicit contract sentence: `CommitSHAs` has no positional relationship to `Commits` and consumers must not join the arrays by index.

collector/snapshot/v1/README.md
```diff
@@ -70,7 +72,11 @@ silently after review.
 
 The allow-list includes bounded session metadata, aggregate counts, raw token
 counts, frozen estimated costs, digest/planning evidence, todos, file-change
-statistics, commit subjects, git drift, and bounded subagent facts.
+statistics, human-facing commit subjects, resolved commit object IDs, git
+drift, and bounded subagent facts. `session.commits` remains subject text;
+`session.commitShas` is optional during its additive rollout and contains only
+locally resolved lowercase full Git object IDs. Consumers must never parse
+subjects as identifiers.
```
calvintvu, line 79: **P2.** Add the rule from the godoc here: `commitShas` has no positional relation to `commits`. A third-party consumer that reads only this README can still join the two arrays by index.

calvintvu (commented):
second pass review

calvintvu (approved):

collector/snapshot/v1/README.md
luhe19001: Addressed in follow-up PR #165: https://github.com/centauri-ai/coslash/pull/165

PR #160, opened 2026-09-08
Redact transcript-derived text from remote cache

calvintvu:
## Context

Remote cache snapshots could persist transcript-derived display text intended only for live session views. Store a privacy-safe snapshot while retaining derived facts needed for useful offline cards and incremental refresh.

## Changes

- Remove prompts, summaries, goals, commands, todos, digest text, file paths, synthesis, and subagent text from persisted remote snapshots.
- Invalidate older cache snapshots so previously stored transcript-derived text is not reused.
- Leave live in-memory session details unchanged.

## Test

- `GOCACHE=/private/tmp/coslash-go-cache go test ./internal/remote -run '^(TestCacheV2StoreOmitsTranscriptDerivedText|TestCache|TestKnownFamilies)' -count=1` — passed
- `GOCACHE=/private/tmp/coslash-go-cache go test ./cmd/coslash -count=1` — passed

### DIFF
--- a/collector/internal/remote/cache.go
+++ b/collector/internal/remote/cache.go
@@ -256,7 +256,7 @@ func age(fetchedAtMs int64, now time.Time) time.Duration {
 // stays visible (marked stale) until the first v2 generation commits, and a
 // v1 fingerprint is never reinterpreted as v2 baseline state — the first v2
 // refresh always starts from an empty generation.
-const cacheV2Version = 3
+const cacheV2Version = 4
 const maxCacheV2Bytes = 64 << 20
 
 // CachedFamilyV2 is one durable family entry. Vendor and FamilyID are stored
@@ -501,6 +501,7 @@ func validCachedSnapshotV2(cached CachedSnapshotV2) bool {
 // closed before an atomic rename replaces the prior snapshot, and the
 // containing directory is synced afterward where the platform supports it.
 func (c *Cache) StoreV2(sourceID string, cached CachedSnapshotV2) error {
+	cached = privacySafeSnapshot(cached)
 	cached.Version = cacheV2Version
 	dir, err := c.sourceDir(sourceID)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


collector/internal/remote/cache.go
```diff
@@ -558,6 +559,31 @@ func (c *Cache) StoreV2(sourceID string, cached CachedSnapshotV2) error {
 	return nil
 }
 
+func privacySafeSnapshot(cached CachedSnapshotV2) CachedSnapshotV2 {
+	cached.Families = append([]CachedFamilyV2(nil), cached.Families...)
+	for i := range cached.Families {
+		facts := cached.Families[i].Facts
+		facts.Sessions = append([]remotefacts.Session(nil), facts.Sessions...)
+		for j := range facts.Sessions {
```
luhe19001, line 567: **[P1] Clear prompt-derived session names before caching**

For Codex root sessions, the parser assigns `parsed.Name = analysis.firstUserPrompt`, and `FromParsed` persists that value in `facts.Sessions[j].Name`. This sanitizer leaves that field intact, so the user's full prompt is still written to `snapshot-v2.json` despite clearing `Display.FirstPrompt`; clear or replace the fact-level name before storing.

collector/internal/remote/cache.go
```diff
@@ -558,6 +559,31 @@ func (c *Cache) StoreV2(sourceID string, cached CachedSnapshotV2) error {
...
+			display.DeclaredGoal = nil
+			display.Synthesis = nil
+			display.Commands = nil
+			display.Commits = nil
+			display.Todos = nil
+			display.Digest = nil
+			display.FileEdits = nil
+			display.Subagents = nil
+			facts.Sessions[j].Display = display
+			facts.Sessions[j].Commands = nil
```
luhe19001, line 582: **[P1] Clear fallback raw command labels from cached facts**

`CommandLog.Note` uses the raw command as its label when a vendor provides no human-readable label, and `FromParsed` copies those labels into `CommandLabels`. Clearing only `facts.Sessions[j].Commands` therefore still persists commands such as credential-bearing `curl` invocations in `snapshot-v2.json`; clear `CommandLabels` as well.

collector/internal/remote/cache.go
calvintvu: Fixed in 363896e. The cache sanitizer now clears Session.Name before serialization. The focused regression test failed before the fix and passed after it.

collector/internal/remote/cache.go
calvintvu: Fixed in 363896e. The cache sanitizer now clears Session.CommandLabels before serialization. The focused regression test failed before the fix and passed after it.

luhe19001 (approved):

PR #163, opened 2026-09-08
Require consent before installing remote helpers

calvintvu:
## Context

Adding a remote host currently proceeds from SSH validation directly to helper installation. Users should explicitly choose whether coSlash may write the helper or continue with the existing SFTP fallback.

## Changes

- Show inline Install connector and Use SFTP only choices after validating and saving a host.
- Reuse verified helpers without prompting, and wait for background discovery before an approved installation.
- Return failed setup retries to the consent choice instead of installing automatically.

## Test

- `npm test -- remote-api.test.ts` — passed
- `npx tsc -p tsconfig.eng1399.json` — passed with a temporary production-only config
- `npx vite build` — passed
- `npx prettier --check src/pages/coslash/components/MachinesSettingsSection.tsx src/pages/coslash/lib/remote-api.ts` — passed
- `npx oxlint src/pages/coslash/components/MachinesSettingsSection.tsx src/pages/coslash/lib/remote-api.ts` — passed

### Screenshots

- [x] Machines settings — inline helper-install consent choices
Screenshot 2026-09-08 at 15 34 45

### DIFF
--- a/frontend/src/pages/coslash/components/MachinesSettingsSection.tsx
+++ b/frontend/src/pages/coslash/components/MachinesSettingsSection.tsx
@@ -2,10 +2,15 @@ import { useEffect, useState } from 'react';
 import { Button } from '@/components/ui/button';
 import { cn } from '@/lib/utils';
 import type { MachineFact } from '@/pages/coslash/lib/machines';
-import { remoteStatus, setupRemoteHelper, testRemoteAlias } from '@/pages/coslash/lib/remote-api';
+import {
+  remoteStatus,
+  setupRemoteHelper,
+  testRemoteAlias,
+  waitForRemoteRefresh,
+} from '@/pages/coslash/lib/remote-api';
 import type { RemoteHostSettings } from '@/pages/coslash/lib/settings';
 
-type SetupStage = 'idle' | 'testing' | 'saving' | 'installing' | 'ready' | 'error' | 'removing';
+type SetupStage = 'idle' | 'testing' | 'saving' | 'consent' | 'installing' | 'ready' | 'error' | 'removing';
 
 function testResultCopy(machine: MachineFact) {
   return machine.state === 'ok'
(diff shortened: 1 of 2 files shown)


frontend/src/pages/coslash/components/MachinesSettingsSection.tsx
```diff
@@ -117,21 +129,30 @@ export function MachinesSettingsSection({
         setMessage('Could not add this SSH host.');
         return;
       }
-      await installConnector(sshAlias);
+      if (test.helper?.compatible) {
```
luhe19001: **[P1] Respect the SFTP-only selection for existing helpers**

`testRemoteAlias` never returns helper details, so this branch is unreachable and an already-installed helper always leads to the consent UI. If discovery then finds that helper, clicking “Use SFTP only” merely changes local UI state; the subsequent retry (and later refreshes) still selects and executes the discovered helper. Either make this choice disable helper transport or label it as skipping a new installation rather than SFTP-only.

frontend/src/pages/coslash/components/MachinesSettingsSection.tsx
```diff
@@ -76,16 +82,22 @@ export function MachinesSettingsSection({
   const installConnector = async (sshAlias: string) => {
     setStage('installing');
     setMessage('Installing connector…');
-    const setup = await setupRemoteHelper(sshAlias, 'install');
-    setMachine(setup.machine);
-    if (setup.error != null) {
+    try {
+      await waitForRemoteRefresh();
```
luhe19001: **[P2] Avoid blocking connector installation on remote collection**

When adding a host, saving settings already starts `handleRemoteRetry`, which begins an SFTP refresh. This wait blocks the install request until that refresh completes; on a slow or large remote tree this can take the collection deadline (up to minutes), while the UI says “Installing connector” and prevents closing the dialog even though `/api/remote/helper/setup` has not been called. Wait only for helper discovery/conflicts, or allow setup to proceed independently of collection.

frontend/src/pages/coslash/components/MachinesSettingsSection.tsx
calvintvu: Fixed in 56cf818. The decline action now says “Skip installation” and does not promise SFTP-only transport. I also removed the unreachable helper check.

frontend/src/pages/coslash/components/MachinesSettingsSection.tsx
calvintvu: Fixed in 56cf818. Connector installation now starts immediately after consent. It no longer waits for remote collection to finish. The clean frontend build passes.

luhe19001 (approved):

PR #164, opened 2026-09-09
Show working directory when remote repository is unavailable

calvintvu:
## Context

Remote session cards can lack repository metadata even when their working directory is known. This leaves the card’s location field blank despite having a useful folder to display.

## Changes

- Fall back to the session working directory when repository metadata is unavailable.
- Continue showing an em dash when neither value exists.

## Test

- `npm test -- --run src/pages/coslash/lib/session.test.ts` — passed
- `npx prettier --check src/pages/coslash/components/SessionCard.tsx src/pages/coslash/lib/session.ts src/pages/coslash/lib/session.test.ts` — passed
- `npm run lint -- src/pages/coslash/components/SessionCard.tsx src/pages/coslash/lib/session.ts src/pages/coslash/lib/session.test.ts` — passed
- Manual: verified a remote card without repository metadata displays `/tmp`.

### Screenshots

- [x] Remote session card — missing folder structure and no repo
Screenshot 2026-09-08 at 18 01 04

- [x] Remote session card — working-directory fallback displayed
Screenshot 2026-09-08 at 18 08 33

### DIFF
--- a/frontend/src/pages/coslash/lib/session.ts
+++ b/frontend/src/pages/coslash/lib/session.ts
@@ -110,6 +110,10 @@ export function environmentFact(value: string | null | undefined): string {
   return trimmed ? trimmed : '—';
 }
 
+export function sessionLocationFact(session: Pick<Session, 'repo' | 'cwd'>): string {
+  return environmentFact(session.repo ?? session.cwd);
+}
+
 export function withLocalSourceDefaults<T extends { agent: string; id: string }>(
   session: T,
 ): T & Pick<Session, 'sourceId' | 'sourceLabel' | 'eligibleForAggregates' | 'displayStale'> {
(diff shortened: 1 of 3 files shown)


frontend/src/pages/coslash/lib/session.ts
```diff
@@ -110,6 +110,10 @@ export function environmentFact(value: string | null | undefined): string {
   return trimmed ? trimmed : '—';
 }
 
+export function sessionLocationFact(session: Pick<Session, 'repo' | 'cwd'>): string {
+  return environmentFact(session.repo ?? session.cwd);
```
luhe19001: **[P2] Fall back when repository metadata is blank**

Could you take a look at blank repository values here? When a remote session has `repo: ''` or whitespace, `??` selects it and `environmentFact` renders `—` instead of using the available `cwd`. Since blank values are otherwise treated as unavailable environment facts, it may be worth selecting `repo` only when its trimmed value is nonempty before falling back. Non-blocking; just flagging it for consideration.

calvintvu: Fixed in 97797b9. The location helper now trims the repository value and uses the working directory when the result is blank. Tests cover empty and whitespace-only values.

luhe19001 (approved):

PR #166, opened 2026-09-09
Keep command-heavy remote sessions instead of dropping them

milanshen:
## Context

Remote collection could skip a valid session when its normalized facts had more command labels than the protocol allows. The helper treated that validation failure as a read error, so the session disappeared from the board with no last-good card.

## Changes

- Truncate over-limit per-session model, spawn, and command lists in transcript order when building remote facts, so the family still publishes.
- Leave Validate as a reject-not-repair gate for helper and cache input.
- Report fact-assembly failures as `invalid_data` instead of `read_failed`.

## Test

- `go test ./internal/remotefacts/` — passed
- `go test ./internal/remotehelper/` — passed
- Manual: rebuilt and upgraded the Linux helper; helper collection published a previously missing command-heavy Codex session.

### DIFF
--- a/collector/internal/remotefacts/facts.go
+++ b/collector/internal/remotefacts/facts.go
@@ -425,6 +425,20 @@ func FromParsed(vendor, familyID, parserVersion, state, staleReason string, pars
 			fact.CommandLabels = append(fact.CommandLabels, truncate(command.Label, MaxDisplayBytes))
 		}
 		fact.Commands = append([]session.SubagentCommand(nil), p.Commands...)
+		// Validate rejects over-limit lists. Keep transcript order and drop the tail
+		// so a command-heavy session still publishes instead of vanishing.
+		if len(fact.Usage) > MaxModelsPerSession {
+			fact.Usage = fact.Usage[:MaxModelsPerSession]
+		}
+		if len(fact.Spawns) > MaxSpawnsPerSession {
+			fact.Spawns = fact.Spawns[:MaxSpawnsPerSession]
+		}
+		if len(fact.CommandLabels) > MaxCommands {
+			fact.CommandLabels = fact.CommandLabels[:MaxCommands]
+		}
+		if len(fact.Commands) > MaxCommands {
+			fact.Commands = fact.Commands[:MaxCommands]
+		}
 		f.Sessions = append(f.Sessions, fact)
 	}
 	sort.Slice(f.Sessions, func(i, j int) bool { return f.Sessions[i].ID < f.Sessions[j].ID })
--- a/collector/internal/remotefacts/facts.go
+++ b/collector/internal/remotefacts/facts.go
@@ -421,24 +421,24 @@ func FromParsed(vendor, familyID, parserVersion, state, staleReason string, pars
 			spawn := p.Spawns[key]
 			fact.Spawns = append(fact.Spawns, Spawn{Key: key, Turn: cloneInt(spawn.Turn), Completed: spawn.Completed})
 		}
-		for _, command := range p.Commands {
+		n := min(len(p.Commands), MaxCommands)
+		for _, command := range p.Commands[:n] {
 			fact.CommandLabels = append(fact.CommandLabels, truncate(command.Label, MaxDisplayBytes))
 		}
-		fact.Commands = append([]session.SubagentCommand(nil), p.Commands...)
-		// Validate rejects over-limit lists. Keep transcript order and drop the tail
-		// so a command-heavy session still publishes instead of vanishing.
+		fact.Commands = append([]session.SubagentCommand(nil), p.Commands[:n]...)
+		fact.Display.Commands = fact.Display.Commands[:min(len(fact.Display.Commands), MaxCommands)]
 		if len(fact.Usage) > MaxModelsPerSession {
+			if fact.RecordedCostMicros == nil {
+				var total int64
+				for _, usage := range fact.Usage {
+					total += usage.CostMicroUSD
+				}
+				total = min(total, MaxCostMicroUSD)
+				fact.RecordedCostMicros = &total
+			}
 			fact.Usage = fact.Usage[:MaxModelsPerSession]
 		}
-		if len(fact.Spawns) > MaxSpawnsPerSession {
-			fact.Spawns = fact.Spawns[:MaxSpawnsPerSession]
-		}
-		if len(fact.CommandLabels) > MaxCommands {
-			fact.CommandLabels = fact.CommandLabels[:MaxCommands]
-		}
-		if len(fact.Commands) > MaxCommands {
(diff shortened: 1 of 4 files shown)


collector/internal/remotefacts/facts.go
```diff
@@ -425,6 +425,20 @@ func FromParsed(vendor, familyID, parserVersion, state, staleReason string, pars
...
+		if len(fact.Usage) > MaxModelsPerSession {
+			fact.Usage = fact.Usage[:MaxModelsPerSession]
+		}
+		if len(fact.Spawns) > MaxSpawnsPerSession {
+			fact.Spawns = fact.Spawns[:MaxSpawnsPerSession]
+		}
+		if len(fact.CommandLabels) > MaxCommands {
+			fact.CommandLabels = fact.CommandLabels[:MaxCommands]
+		}
+		if len(fact.Commands) > MaxCommands {
```
calvintvu: P1: The cap does not include `Display.Commands`, so a 300-command session reports 128 commands here and 300 in `Display`. Cap `fact.Display.Commands` to `MaxCommands` also.

collector/internal/remotefacts/facts.go
```diff
@@ -425,6 +425,20 @@ func FromParsed(vendor, familyID, parserVersion, state, staleReason string, pars
 			fact.CommandLabels = append(fact.CommandLabels, truncate(command.Label, MaxDisplayBytes))
 		}
 		fact.Commands = append([]session.SubagentCommand(nil), p.Commands...)
+		// Validate rejects over-limit lists. Keep transcript order and drop the tail
```
calvintvu: P1: The truncation leaves `f.State` as `StateComplete`, so the Mac caches lossy data as authoritative and never re-fetches it. Set `f.State = StatePartial` here, as line 395 already does for a missing parent.

collector/internal/remotefacts/facts.go
```diff
@@ -425,6 +425,20 @@ func FromParsed(vendor, familyID, parserVersion, state, staleReason string, pars
...
+		// so a command-heavy session still publishes instead of vanishing.
+		if len(fact.Usage) > MaxModelsPerSession {
+			fact.Usage = fact.Usage[:MaxModelsPerSession]
```
calvintvu, line 439: P1: `Parsed()` rebuilds `s.Tokens` from `fact.Usage`, but `s.Cost` stays inside `fact.Display`. A session with 37 models at $1 each then reports `Display.Cost=37.00` against a summed usage cost of 32.00.

collector/internal/remotefacts/facts.go
```diff
@@ -425,6 +425,20 @@ func FromParsed(vendor, familyID, parserVersion, state, staleReason string, pars
...
+		}
+		if len(fact.Spawns) > MaxSpawnsPerSession {
+			fact.Spawns = fact.Spawns[:MaxSpawnsPerSession]
```
calvintvu: P1: A dropped spawn orphans the child that references it, and `subagents.go:66` then shows a finished subagent as Running forever. Keep the spawns that a present child references before you drop any.

collector/internal/remotefacts/facts.go
```diff
@@ -425,6 +425,20 @@ func FromParsed(vendor, familyID, parserVersion, state, staleReason string, pars
...
+		}
+		if len(fact.CommandLabels) > MaxCommands {
+			fact.CommandLabels = fact.CommandLabels[:MaxCommands]
+		}
+		if len(fact.Commands) > MaxCommands {
+			fact.Commands = fact.Commands[:MaxCommands]
+		}
```
calvintvu: Optional: These four blocks build the full list and then drop the tail. Cap at the source with `min()`, as `synthesis/runner.go:275` does, for 9 fewer lines.

collector/internal/remotehelper/collect.go
```diff
@@ -354,16 +354,20 @@ func familyFacts(
...
 	if len(sessions) < len(item.sessionIDs) {
 		state = remotefacts.StatePartial
 	}
-	return remotefacts.FromParsed(
+	family, err := remotefacts.FromParsed(
 		scanned.vendor, item.id, vendors.ParserVersion, state, "",
 		sessions, scanned.metadata, item.fingerprints, item.headerMappings,
 	)
+	if err != nil {
+		return remotefacts.Family{}, fmt.Errorf("%w: %v", vendors.ErrInvalidData, err)
```
calvintvu, line 368: Optional: No consumer compares `StaleReasonInvalidData` against `StaleReasonReadFailed`, so this wrap relabels a string that nothing reads. Keep it only if you will branch on the reason.

collector/internal/remotefacts/facts.go
milanshen: Agreed — `Parsed()` copies `Display` first, so the board’s `commands` field stayed uncapped. `Display.Commands` is now truncated to `MaxCommands` with the other command lists.

collector/internal/remotefacts/facts.go
milanshen: Leaving this as `complete`. Unchanged refresh is fingerprint equality (`cached == item.fingerprint`), not `Facts.State`, so marking partial would not cause a re-fetch while the files are unchanged. `StatePartial` already means a missing parent in the family graph; a truncated command list is still a complete family.

collector/internal/remotefacts/facts.go
milanshen: Agreed the mismatch is real: `Parsed()` rebuilds `Tokens` from capped `Usage`, then `AttachCost` would under-report for Claude/Codex (no recorded cost). Matching `Cost` to the truncated sum would hide spend. When usage is capped we now store the pre-cap total as `RecordedCostMicros` so the card keeps the full cost.

collector/internal/remotefacts/facts.go
milanshen: Agreed — a tail drop can leave a Claude child without `parent.Spawns[spawnKey]`, so `subagents.go:66` treats a finished agent as Running. `capSpawns` now keeps referenced keys first, then fills remaining slots in sorted order.

collector/internal/remotefacts/facts.go
milanshen: Done — command copies now cap with `min()` at the source instead of build-then-slice.

collector/internal/remotehelper/collect.go
milanshen: Keeping the wrap. SFTP already maps `FromParsed` failures to `invalid_data`; this is the same classification on the helper path for a remaining producer failure. Nothing branches on the two strings for user copy — both still mark the last-good card stale — but the skip reason should not look like an I/O error.

calvintvu (approved):

PR #172, opened 2026-09-12
feat(remote): add guided SSH authentication

luhe19001:
## Summary
- add explicit terminal-assisted SSH authentication and reconnect flow
- distinguish host-key confirmation from changed-key remediation
- coordinate auth attempts with SSH control-master lifecycle and cleanup
- add durable auth outcomes, machine action state, and lifecycle tests

## UI
image

## Validation
- `go test ./...`
- `npm run build`
- focused frontend authentication API tests
- `make release`
- `make smoke`

Closes ENG-1412.

### DIFF
--- a/collector/cmd/coslash/api_remote.go
+++ b/collector/cmd/coslash/api_remote.go
@@ -6,8 +6,10 @@ import (
 	"errors"
 	"io"
 	"net/http"
+	"os"
 	"time"
 
+	"github.com/centauri-ai/coslash/collector/internal/launch"
 	"github.com/centauri-ai/coslash/collector/internal/remote"
 	"github.com/centauri-ai/coslash/collector/internal/settings"
 )
@@ -20,6 +22,15 @@ type remoteHelperSetupRequest struct {
 	SSHAlias string `json:"sshAlias"`
 	Install  bool   `json:"install"`
 	Upgrade  bool   `json:"upgrade"`
+}
+
+type remoteAuthResponse struct {
+	ID    string           `json:"id"`
+	State remote.AuthState `json:"state"`
+}
+
+func terminalAuthenticationPermitted(health remote.Health) bool {
+	return health.Reason != nil && (*health.Reason == remote.ReasonAuthentication || *health.Reason == remote.ReasonHostKeyConfirmation)
 }
 
 type helperSetupResponse struct {
@@ -76,18 +87,81 @@ func handleRemoteTest(w http.ResponseWriter, request *http.Request, manager *rem
 	var body remoteTestRequest
 	decoder := json.NewDecoder(io.LimitReader(request.Body, 4<<10))
 	decoder.DisallowUnknownFields()
-	if err := decoder.Decode(&body); err != nil || !settings.ValidSSHAlias(body.SSHAlias) {
-		http.Error(w, "invalid remote test request", http.StatusBadRequest)
+	if err := decoder.Decode(&body); err != nil || decoder.Decode(&struct{}{}) != io.EOF || !settings.ValidSSHAlias(body.SSHAlias) {
+		writeAPIError(w, http.StatusBadRequest, "invalid_destination", "Enter an SSH alias or user@host.")
 		return
 	}
 	ctx, cancel := context.WithTimeout(request.Context(), remote.DefaultConnectTimeout+5*time.Second)
 	defer cancel()
 	health, err := manager.TestAlias(ctx, body.SSHAlias)
 	if err != nil {
-		http.Error(w, "invalid sshAlias", http.StatusBadRequest)
+		writeAPIError(w, http.StatusBadRequest, "invalid_destination", "Enter an SSH alias or user@host.")
 		return
 	}
 	writeJSON(w, machineFromHealth(health))
+}
+
+func handleRemoteAuthStart(w http.ResponseWriter, request *http.Request, manager *remote.Manager, store *settings.Store) {
+	var body remoteTestRequest
+	decoder := json.NewDecoder(io.LimitReader(request.Body, 4<<10))
+	decoder.DisallowUnknownFields()
+	if err := decoder.Decode(&body); err != nil || decoder.Decode(&struct{}{}) != io.EOF || !settings.ValidSSHAlias(body.SSHAlias) {
+		writeAPIError(w, http.StatusBadRequest, "invalid_destination", "Enter an SSH alias or user@host.")
+		return
+	}
+	if remote.AuthAttemptActive(body.SSHAlias) {
+		writeAPIError(w, http.StatusConflict, "authentication_in_progress", "Authentication is already waiting in Terminal.")
+		return
+	}
+	ctx, cancel := context.WithTimeout(request.Context(), remote.DefaultConnectTimeout+5*time.Second)
+	defer cancel()
+	health, err := manager.TestAlias(ctx, body.SSHAlias)
+	if err != nil || !terminalAuthenticationPermitted(health) {
+		writeAPIError(w, http.StatusConflict, "authentication_not_required", "Terminal authentication is not required for this host.")
+		return
+	}
+	id, err := remote.CreateAuthAttempt(body.SSHAlias)
+	if errors.Is(err, remote.ErrAuthAttemptActive) {
+		writeAPIError(w, http.StatusConflict, "authentication_in_progress", "Authentication is already waiting in Terminal.")
+		return
+	}
+	if err != nil {
+		writeAPIError(w, http.StatusInternalServerError, "authentication_unavailable", "Could not prepare terminal authentication.")
+		return
+	}
+	executable, err := os.Executable()
+	if err == nil {
+		err = launch.SSHAuthentication(store.State().Config.Launch.Terminal, executable, id)
+	}
+	if err != nil {
+		_ = remote.CancelAuthAttempt(id)
+		writeAPIError(w, http.StatusConflict, "terminal_unavailable", "Could not open the selected terminal for SSH authentication.")
+		return
+	}
+	writeJSON(w, remoteAuthResponse{ID: id, State: remote.AuthWaiting})
+}
+
+func handleRemoteAuthStatus(w http.ResponseWriter, request *http.Request) {
+	id := request.URL.Query().Get("id")
+	state, err := remote.AuthAttemptState(request.Context(), id)
+	if err != nil {
+		writeAPIError(w, http.StatusBadRequest, "invalid_authentication_attempt", "Authentication attempt is unavailable.")
+		return
+	}
+	writeJSON(w, remoteAuthResponse{ID: id, State: state})
+}
+
+func handleRemoteAuthCancel(w http.ResponseWriter, request *http.Request) {
+	var body struct {
+		ID string `json:"id"`
+	}
+	decoder := json.NewDecoder(io.LimitReader(request.Body, 4<<10))
+	decoder.DisallowUnknownFields()
+	if err := decoder.Decode(&body); err != nil || decoder.Decode(&struct{}{}) != io.EOF || remote.CancelAuthAttempt(body.ID) != nil {
+		writeAPIError(w, http.StatusBadRequest, "invalid_authentication_attempt", "Authentication attempt is unavailable.")
+		return
+	}
+	writeJSON(w, remoteAuthResponse{ID: body.ID, State: remote.AuthCancelled})
 }
 
 func handleRemoteRetry(w http.ResponseWriter, _ *http.Request, manager *remote.Manager) {
--- a/collector/internal/remote/auth.go
+++ b/collector/internal/remote/auth.go
@@ -178,20 +178,9 @@ func CancelAuthAttempt(id string) error {
 	}
 	_, err = updateAuthAttemptIfWaiting(id, AuthCancelled)
 	if err == nil {
-		removeAuthAttempt(id)
 		exitAuthControlMaster(attempt.Destination)
 	}
 	return err
-}
-
-// removeAuthAttempt removes only the validated record path. The single
-// coordinator lock is retained because unlinking a live advisory lock could
-// let another process create a different inode and bypass synchronization.
-func removeAuthAttempt(id string) {
-	path, err := authAttemptPath(id)
-	if err == nil {
-		_ = os.Remove(path)
-	}
 }
 
 // CancelAuthAttemptsForDestination prevents a removed host from gaining a
@@ -216,7 +205,10 @@ func CancelAuthAttemptsForDestination(destination string) {
 			if _, err := authAttemptPath(attempt.ID); err != nil {
 				continue
 			}
-			removeAuthAttempt(attempt.ID)
+			attempt.State = AuthCancelled
+			if err := writeAuthAttempt(attempt); err != nil {
+				return err
+			}
 			removed = true
 		}
 		return nil
@@ -334,6 +326,12 @@ func RunAuthAttempt(ctx context.Context, id string) error {
 		if err != nil {
 			return err
 		}
+		if latest.State == AuthReady {
+			// A status poll may see the new socket before this terminal process
+			// reacquires the coordinator. That is a successful attempt, not a
+			// cancellation, so the shared master must remain available.
+			return nil
+		}
 		if latest.State != AuthWaiting {
 			closeMaster = true
 			return nil
@@ -360,7 +358,6 @@ func AuthAttemptState(ctx context.Context, id string) (AuthState, error) {
 		return "", err
 	}
 	if attempt.State != AuthWaiting {
-		removeAuthAttempt(id)
 		return attempt.State, nil
 	}
 	if time.Since(attempt.CreatedAt) > AuthAttemptTTL {
@@ -368,7 +365,6 @@ func AuthAttemptState(ctx context.Context, id string) (AuthState, error) {
 		if updateErr != nil {
 			return "", updateErr
 		}
-		removeAuthAttempt(id)
 		return state, nil
 	}
 	args, err := controlCheckArgs(attempt.Destination)
@@ -382,7 +378,6 @@ func AuthAttemptState(ctx context.Context, id string) (AuthState, error) {
 		if updateErr != nil {
 			return "", updateErr
 		}
-		removeAuthAttempt(id)
 		return state, nil
 	}
 	return AuthWaiting, nil
--- a/collector/internal/remote/auth.go
+++ b/collector/internal/remote/auth.go
@@ -1,0 +1,389 @@
+package remote
+
+import (
+	"context"
+	"crypto/rand"
+	"encoding/hex"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"regexp"
+	"syscall"
+	"time"
+
+	"github.com/centauri-ai/coslash/collector/internal/settings"
+)
+
+const AuthAttemptTTL = 5 * time.Minute
+
+var authAttemptIDPattern = regexp.MustCompile(`^[0-9a-f]{32}$`)
+
+var ErrAuthAttemptActive = errors.New("authentication attempt already active")
+
+type AuthState string
+
+const (
+	AuthNotRequired AuthState = "not_required"
+	AuthRequired    AuthState = "required"
+	AuthLaunching   AuthState = "launching"
+	AuthWaiting     AuthState = "waiting"
+	AuthReady       AuthState = "ready"
+	AuthTimedOut    AuthState = "timed_out"
+	AuthCancelled   AuthState = "cancelled"
+	AuthFailed      AuthState = "failed"
+)
+
+type authAttempt struct {
+	ID          string    `json:"id"`
+	Destination string    `json:"destination"`
+	CreatedAt   time.Time `json:"createdAt"`
+	State       AuthState `json:"state"`
+}
+
+func authAttemptPath(id string) (string, error) {
+	if !authAttemptIDPattern.MatchString(id) {
+		return "", errors.New("invalid authentication attempt")
+	}
+	return filepath.Join(settings.Home(), "ssh", "auth-"+id+".json"), nil
+}
+
+func createAuthAttempt(destination string) (string, error) {
+	if _, err := parseDestination(destination); err != nil {
+		return "", err
+	}
+	var id string
+	err := withDestinationCoordinator(destination, func() error {
+		entries, err := filepath.Glob(filepath.Join(settings.Home(), "ssh", "auth-*.json"))
+		if err != nil {
+			return err
+		}
+		for _, path := range entries {
+			data, readErr := os.ReadFile(path)
+			var existing authAttempt
+			if readErr != nil || json.Unmarshal(data, &existing) != nil || time.Since(existing.CreatedAt) > AuthAttemptTTL {
+				_ = os.Remove(path)
+				continue
+			}
+			if existing.Destination == destination && existing.State == AuthWaiting {
+				return ErrAuthAttemptActive
+			}
+		}
+		raw := make([]byte, 16)
+		if _, err := rand.Read(raw); err != nil {
+			return fmt.Errorf("generate authentication attempt: %w", err)
+		}
+		id = hex.EncodeToString(raw)
+		path, err := authAttemptPath(id)
+		if err != nil {
+			return err
+		}
+		data, err := json.Marshal(authAttempt{ID: id, Destination: destination, CreatedAt: time.Now().UTC(), State: AuthWaiting})
+		if err != nil {
+			return err
+		}
+		file, err := os.OpenFile(path, os.O_WRONLY|os.O_CREATE|os.O_EXCL, 0o600)
+		if err != nil {
+			return fmt.Errorf("create authentication attempt: %w", err)
+		}
+		if _, err := file.Write(data); err != nil {
+			_ = file.Close()
+			_ = os.Remove(path)
+			return fmt.Errorf("write authentication attempt: %w", err)
+		}
+		if err := file.Close(); err != nil {
+			_ = os.Remove(path)
+			return fmt.Errorf("close authentication attempt: %w", err)
+		}
+		return nil
+	})
+	return id, err
+}
+
+// withDestinationCoordinator serializes an interactive attempt with every
+// control-master check/start. The single lock is stronger than a per-
+// destination lock, and avoids leaving a new lock file for every attempted
+// destination. flock is released by the kernel if a process exits.
+func withDestinationCoordinator(_ string, callback func() error) error {
+	if err := ensureSSHControlDir(); err != nil {
+		return err
+	}
+	lockPath := filepath.Join(settings.Home(), "ssh", ".auth-coordinator.lock")
+	lock, err := os.OpenFile(lockPath, os.O_WRONLY|os.O_CREATE, 0o600)
+	if err != nil {
+		return err
+	}
+	defer func() { _ = lock.Close() }()
+	if err := syscall.Flock(int(lock.Fd()), syscall.LOCK_EX); err != nil {
+		return fmt.Errorf("lock SSH destination: %w", err)
+	}
+	defer func() { _ = syscall.Flock(int(lock.Fd()), syscall.LOCK_UN) }()
+	return callback()
+}
+
+var runInteractiveSSH = func(ctx context.Context, args []string) error {
+	cmd := exec.CommandContext(ctx, "ssh", args...)
+	cmd.Stdin, cmd.Stdout, cmd.Stderr = os.Stdin, os.Stdout, os.Stderr
+	return cmd.Run()
+}
+
+var exitAuthControlMaster = exitControlMasterBestEffort
+
+func loadAuthAttempt(id string) (authAttempt, error) {
+	path, err := authAttemptPath(id)
+	if err != nil {
+		return authAttempt{}, err
+	}
+	data, err := os.ReadFile(path)
+	if err != nil {
+		return authAttempt{}, err
+	}
+	var attempt authAttempt
+	if err := json.Unmarshal(data, &attempt); err != nil || attempt.ID != id {
+		return authAttempt{}, errors.New("invalid authentication attempt")
+	}
+	if _, err := parseDestination(attempt.Destination); err != nil {
+		return authAttempt{}, errors.New("invalid authentication destination")
+	}
+	return attempt, nil
+}
+
+// CreateAuthAttempt records only the already-validated destination in a
+// mode-0600, short-lived local file. The terminal receives only its opaque ID.
+func CreateAuthAttempt(destination string) (string, error) { return createAuthAttempt(destination) }
+
+// AuthAttemptActive is used by background work to avoid racing an explicit
+// terminal prompt. Corrupt and expired records are not considered active.
+func AuthAttemptActive(destination string) bool {
+	entries, err := filepath.Glob(filepath.Join(settings.Home(), "ssh", "auth-*.json"))
+	if err != nil {
+		return false
+	}
+	for _, path := range entries {
+		data, err := os.ReadFile(path)
+		var attempt authAttempt
+		if err == nil && json.Unmarshal(data, &attempt) == nil && attempt.Destination == destination && attempt.State == AuthWaiting && time.Since(attempt.CreatedAt) <= AuthAttemptTTL {
+			return true
+		}
+	}
+	return false
+}
+
+func CancelAuthAttempt(id string) error {
+	attempt, err := loadAuthAttempt(id)
+	if err != nil {
+		return err
+	}
+	_, err = updateAuthAttemptIfWaiting(id, AuthCancelled)
+	if err == nil {
+		removeAuthAttempt(id)
+		exitAuthControlMaster(attempt.Destination)
+	}
+	return err
+}
+
+// removeAuthAttempt removes only the validated record path. The single
+// coordinator lock is retained because unlinking a live advisory lock could
+// let another process create a different inode and bypass synchronization.
+func removeAuthAttempt(id string) {
+	path, err := authAttemptPath(id)
+	if err == nil {
+		_ = os.Remove(path)
+	}
+}
+
+// CancelAuthAttemptsForDestination prevents a removed host from gaining a
+// late interactive master. The terminal command re-checks its record under
+// this same destination coordinator before treating SSH success as ready.
+func CancelAuthAttemptsForDestination(destination string) {
+	if destination == "" {
+		return
+	}
+	removed := false
+	_ = withDestinationCoordinator(destination, func() error {
+		entries, err := filepath.Glob(filepath.Join(settings.Home(), "ssh", "auth-*.json"))
+		if err != nil {
+			return err
+		}
+		for _, path := range entries {
+			data, err := os.ReadFile(path)
+			var attempt authAttempt
+			if err != nil || json.Unmarshal(data, &attempt) != nil || attempt.Destination != destination {
+				continue
+			}
+			if _, err := authAttemptPath(attempt.ID); err != nil {
+				continue
+			}
+			removeAuthAttempt(attempt.ID)
+			removed = true
+		}
+		return nil
+	})
+	if removed {
+		exitAuthControlMaster(destination)
+	}
+}
+
+// updateAuthAttemptIfWaiting preserves a terminal result chosen by another
+// process (for example, Cancel in the app racing a terminal SSH failure).
+func updateAuthAttemptIfWaiting(id string, state AuthState) (AuthState, error) {
+	initial, err := loadAuthAttempt(id)
+	if err != nil {
+		return "", err
+	}
+	var result AuthState
+	err = withDestinationCoordinator(initial.Destination, func() error {
+		attempt, err := loadAuthAttempt(id)
+		if err != nil {
+			return err
+		}
+		if attempt.State != AuthWaiting {
+			result = attempt.State
+			return nil
+		}
+		attempt.State = state
+		if err := writeAuthAttempt(attempt); err != nil {
+			return err
+		}
+		result = state
+		return nil
+	})
+	return result, err
+}
+
+func writeAuthAttempt(attempt authAttempt) error {
+	path, err := authAttemptPath(attempt.ID)
+	if err != nil {
+		return err
+	}
+	data, err := json.Marshal(attempt)
+	if err != nil {
+		return err
+	}
+	temporary, err := os.CreateTemp(filepath.Dir(path), ".auth-*")
+	if err != nil {
+		return err
+	}
+	if err := temporary.Chmod(0o600); err == nil {
+		_, err = temporary.Write(data)
+	}
+	if closeErr := temporary.Close(); err == nil {
+		err = closeErr
+	}
+	if err != nil {
+		_ = os.Remove(temporary.Name())
+		return err
+	}
+	return os.Rename(temporary.Name(), path)
+}
+
+func interactiveMasterArgs(destination string) ([]string, error) {
+	parsed, err := parseDestination(destination)
+	if err != nil {
+		return nil, err
+	}
+	args := []string{
+		"-f", "-N",
+		"-o", "BatchMode=no",
+		"-o", "ConnectTimeout=" + fmt.Sprint(int(DefaultConnectTimeout.Seconds())),
+		"-o", "ControlMaster=yes",
+		"-o", "ControlPath=" + controlSocketPath(),
+		"-o", "ControlPersist=" + defaultControlPersist,
+	}
+	return append(args, parsed.Args()...), nil
+}
+
+// RunAuthAttempt is the ssh-auth subcommand entry point. Its stdin and output
+// belong to the terminal, never to the collector HTTP process.
+func RunAuthAttempt(ctx context.Context, id string) error {
+	attempt, err := loadAuthAttempt(id)
+	if err != nil {
+		return err
+	}
+	if time.Since(attempt.CreatedAt) > AuthAttemptTTL {
+		_, _ = updateAuthAttemptIfWaiting(id, AuthTimedOut)
+		return errors.New("authentication attempt expired")
+	}
+	deadline := attempt.CreatedAt.Add(AuthAttemptTTL)
+	ctx, cancel := context.WithDeadline(ctx, deadline)
+	defer cancel()
+	args, err := interactiveMasterArgs(attempt.Destination)
+	if err != nil {
+		return err
+	}
+	if err := runInteractiveSSH(ctx, args); err != nil {
+		state := AuthFailed
+		if errors.Is(ctx.Err(), context.DeadlineExceeded) {
+			state = AuthTimedOut
+		}
+		if errors.Is(ctx.Err(), context.Canceled) {
+			state = AuthCancelled
+		}
+		_, _ = updateAuthAttemptIfWaiting(id, state)
+		return fmt.Errorf("authenticate SSH: %w", err)
+	}
+	closeMaster := false
+	err = withDestinationCoordinator(attempt.Destination, func() error {
+		latest, err := loadAuthAttempt(id)
+		if errors.Is(err, os.ErrNotExist) {
+			closeMaster = true
+			return nil
+		}
+		if err != nil {
+			return err
+		}
+		if latest.State != AuthWaiting {
+			closeMaster = true
+			return nil
+		}
+		latest.State = AuthReady
+		return writeAuthAttempt(latest)
+	})
+	if err != nil {
+		return fmt.Errorf("record SSH authentication: %w", err)
+	}
+	if closeMaster {
+		exitAuthControlMaster(attempt.Destination)
+		return errors.New("authentication attempt was cancelled")
+	}
+	return nil
+}
+
+func AuthAttemptState(ctx context.Context, id string) (AuthState, error) {
+	attempt, err := loadAuthAttempt(id)
+	if errors.Is(err, os.ErrNotExist) {
+		return AuthCancelled, nil
+	}
+	if err != nil {
+		return "", err
+	}
+	if attempt.State != AuthWaiting {
+		removeAuthAttempt(id)
+		return attempt.State, nil
+	}
+	if time.Since(attempt.CreatedAt) > AuthAttemptTTL {
+		state, updateErr := updateAuthAttemptIfWaiting(id, AuthTimedOut)
+		if updateErr != nil {
+			return "", updateErr
+		}
+		removeAuthAttempt(id)
+		return state, nil
+	}
+	args, err := controlCheckArgs(attempt.Destination)
+	if err != nil {
+		return "", err
+	}
+	checkCtx, cancel := context.WithTimeout(ctx, 3*time.Second)
+	defer cancel()
+	if err := runSSHCommand(checkCtx, OpenOptions{}, args); err == nil {
+		state, updateErr := updateAuthAttemptIfWaiting(id, AuthReady)
+		if updateErr != nil {
+			return "", updateErr
+		}
+		removeAuthAttempt(id)
+		return state, nil
+	}
+	return AuthWaiting, nil
+}
--- a/collector/internal/remote/manager.go
+++ b/collector/internal/remote/manager.go
@@ -260,6 +260,7 @@ func (manager *Manager) ApplySettings(remote *settings.RemoteSettings) error {
 	previous := manager.cfg
 	aliasChanged := previous != nil && previous.SSHAlias != remote.SSHAlias
 	if aliasChanged {
+		CancelAuthAttemptsForDestination(previous.SSHAlias)
 		exitControlMasterBestEffort(previous.SSHAlias)
 		if err := manager.cache.RemoveSource(remote.ID); err != nil {
 			return err
@@ -283,6 +284,7 @@ func (manager *Manager) ApplySettings(remote *settings.RemoteSettings) error {
 		manager.complete = true
 		manager.errorCopy = ""
 		manager.transport = TransportSFTP
+		CancelAuthAttemptsForDestination(remote.SSHAlias)
 		exitControlMasterBestEffort(remote.SSHAlias)
 		return nil
 	}
@@ -397,6 +399,9 @@ func (manager *Manager) Retry() (Health, bool) {
 	manager.mu.Lock()
 	defer manager.mu.Unlock()
 	if manager.cfg == nil || !manager.cfg.Enabled {
+		return manager.healthLocked(manager.lastRequestedMs), false
+	}
+	if AuthAttemptActive(manager.cfg.SSHAlias) {
 		return manager.healthLocked(manager.lastRequestedMs), false
 	}
 	if manager.refreshing || (!manager.lastManualRetryAt.IsZero() &&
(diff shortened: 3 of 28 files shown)


luhe19001 (commented):
Found two concrete issues in the new SSH-auth/destination support.

Review profile: Helu

collector/internal/remote/auth.go
```diff
@@ -0,0 +1,389 @@
...
+		return "", err
+	}
+	checkCtx, cancel := context.WithTimeout(ctx, 3*time.Second)
+	defer cancel()
+	if err := runSSHCommand(checkCtx, OpenOptions{}, args); err == nil {
+		state, updateErr := updateAuthAttemptIfWaiting(id, AuthReady)
+		if updateErr != nil {
+			return "", updateErr
+		}
+		removeAuthAttempt(id)
```
luhe19001: [medium] A status poll can observe the newly-created control master before `RunAuthAttempt` reaches its post-SSH coordinator block. This path marks the attempt `ready` and deletes its record; `RunAuthAttempt` then sees the record missing (or `ready` rather than `waiting`) at lines 329/337 and treats it as cancellation, explicitly exits the control master. The UI receives `ready` but its immediate test/retry can race with or follow a destroyed master, so successful terminal authentication intermittently fails. Keep a successful attempt distinguishable from cancellation and do not tear down the master when the state is `ready`.

collector/internal/settings/settings.go
```diff
@@ -169,9 +172,42 @@ func ValidSynthesisModel(model string) bool {
...
+}
+
+func ParseSSHDestination(value string) (SSHDestination, error) {
+	if len(value) == 0 || len(value) > 255 {
+		return SSHDestination{}, errors.New("SSH destination must be 1 to 255 characters")
+	}
+	if sshAliasPattern.MatchString(value) {
+		return SSHDestination{Display: value, Host: value}, nil
+	}
+	user, host, found := strings.Cut(value, "@")
```
luhe19001, line 199: [low] The implementation now accepts and persists `user@host`, but the advertised `settings.schema.json` still restricts `remote.sshAlias` to `^[A-Za-z0-9][A-Za-z0-9._-]*$`. A settings file produced by the new UI therefore fails validation in schema-aware editors/tools (and contradicts the README-linked file format). Update the schema pattern and description to cover the same bounded `user@host` grammar.

collector/internal/remote/auth.go
luhe19001: Fixed in 2330743. Terminal outcomes now remain available through the terminal’s final coordinator check. A pre-existing `ready` state is treated as success and leaves the shared master intact; only cancelled, timed-out, failed, or missing attempts trigger teardown. Added a regression test for the status-poll-before-terminal-finalization ordering.

collector/internal/settings/settings.go
luhe19001: Fixed in 2330743. `settings.schema.json` now accepts the same bounded alias-or-`user@host` grammar as the Go parser, and its description plus README documentation have been updated.

frontend/src/pages/coslash/components/MachinesSettingsSection.tsx
```diff
@@ -34,13 +73,32 @@ export function MachinesSettingsSection({
...
   const [message, setMessage] = useState<string | null>(null);
   const [machine, setMachine] = useState<MachineFact | null>(null);
-  const busy = stage === 'testing' || stage === 'saving' || stage === 'installing' || stage === 'removing';
+  const [authAlias, setAuthAlias] = useState<string | null>(null);
+  const [authAttemptID, setAuthAttemptID] = useState<string | null>(null);
+  const [authReconnect, setAuthReconnect] = useState(false);
+  const authenticationRun = useRef<AuthenticationRun | null>(null);
+  const busy =
+    stage === 'testing' ||
+    stage === 'authenticating' ||
```
luhe19001: [medium] Including `authenticating` in `busy` activates `SettingsDialog`'s `remoteOperationInProgress` guard, which applies `pointer-events-none` to the entire dialog content and prevents closing via the close button or Escape. That also disables the Cancel button rendered for this stage, so a user who starts authentication cannot cancel or close Settings while an SSH prompt remains open (up to the five-minute attempt TTL). Keep this stage cancelable, or exempt its Cancel control/dialog close path from the operation lock.

luhe19001 (commented):
Found one actionable UI regression in the terminal-authentication flow.

Review profile: Helu

frontend/src/pages/coslash/components/MachinesSettingsSection.tsx
luhe19001: Fixed in 6344352. `authenticating` no longer triggers the dialog-wide operation lock, so Cancel and dialog dismissal remain available. The destructive Remove control itself is explicitly disabled during authentication; closing the dialog invokes the existing cleanup, which cancels the server attempt.

frontend/src/pages/coslash/components/MachinesSettingsSection.tsx
```diff
@@ -34,13 +73,28 @@ export function MachinesSettingsSection({
   const [stage, setStage] = useState<SetupStage>('idle');
   const [message, setMessage] = useState<string | null>(null);
   const [machine, setMachine] = useState<MachineFact | null>(null);
+  const [authAlias, setAuthAlias] = useState<string | null>(null);
+  const [authAttemptID, setAuthAttemptID] = useState<string | null>(null);
+  const [authReconnect, setAuthReconnect] = useState(false);
+  const authenticationRun = useRef<AuthenticationRun | null>(null);
   const busy = stage === 'testing' || stage === 'saving' || stage === 'installing' || stage === 'removing';
```
luhe19001, line 80: [medium] `authenticating` is excluded from `busy`, so while adding a new host the alias input and “Add remote host” button remain usable (they are disabled only by `busy`). A user can begin terminal auth for host A, then submit host B; the original `waitForAuthentication` still holds A and calls `saveVerifiedHost(A)` after it succeeds, overwriting the later choice. Disable/reject competing setup actions while an authentication run is active (while preserving the Cancel action).

luhe19001 (commented):
One concrete race in the new authentication setup flow.

Review profile: Helu

frontend/src/pages/coslash/components/MachinesSettingsSection.tsx
luhe19001: Fixed in 0b7c875. Active terminal authentication now locks the alias input, Add, and Remove controls via a local setup-actions guard, and `addHost` rejects programmatic/re-entrant invocation as well. The dialog-wide busy lock remains off during auth so Cancel and dismissal still work.

luhe19001 (commented):

Review profile: Helu

frontend/src/pages/coslash/components/MachinesSettingsSection.tsx
```diff
@@ -94,7 +150,126 @@ export function MachinesSettingsSection({
...
+  };
+
+  const authenticateInTerminal = async () => {
+    if (authAlias == null || authenticationRun.current != null) return;
+    const run: AuthenticationRun = { cancelled: false, controller: new AbortController(), attemptID: null };
+    authenticationRun.current = run;
+    setStage('authenticating');
+    setMessage('Terminal opened. Complete the SSH prompt there; setup will continue automatically.');
+    try {
+      const attempt = await startRemoteAuthentication(authAlias);
```
luhe19001, line 231: [medium] The authentication start request is not given `run.controller.signal`. Consequently, clicking Cancel while the server is still running its initial SSH/SFTP probe (up to ~12 seconds) only updates the UI; the request continues and can create the attempt and open a Terminal window after cancellation. Pass `run.controller.signal` to `startRemoteAuthentication` (and ensure the server stops work on request cancellation) so Cancel prevents a late terminal launch.

luhe19001 (commented):
Found one cancellation regression in the guided SSH authentication flow.

Review profile: Helu

collector/cmd/coslash/api_remote.go
```diff
@@ -76,20 +87,83 @@ func handleRemoteTest(w http.ResponseWriter, request *http.Request, manager *rem
...
+		return
+	}
+	if remote.AuthAttemptActive(body.SSHAlias) {
+		writeAPIError(w, http.StatusConflict, "authentication_in_progress", "Authentication is already waiting in Terminal.")
+		return
+	}
+	ctx, cancel := context.WithTimeout(request.Context(), remote.DefaultConnectTimeout+5*time.Second)
+	defer cancel()
+	health, err := manager.TestAlias(ctx, body.SSHAlias)
+	if err != nil || !terminalAuthenticationPermitted(health) {
```
calvintvu: **[P2]** -  Allow authentication for password-only and keyboard-interactive hosts

The new gate rejects hosts that report `Permission denied (password).` or `Permission denied (keyboard-interactive).` The existing classifier maps both errors to `connection_failed`, so the UI offers no authentication button. This endpoint also returns `authentication_not_required`, although these hosts need a password or an interactive prompt.

Extend the authentication error classification to include these methods before this gate uses the result.

A focused test reproduced both error cases on this commit.

collector/internal/remote/auth.go
```diff
@@ -0,0 +1,384 @@
...
+		return errors.New("authentication attempt expired")
+	}
+	deadline := attempt.CreatedAt.Add(AuthAttemptTTL)
+	ctx, cancel := context.WithDeadline(ctx, deadline)
+	defer cancel()
+	args, err := interactiveMasterArgs(attempt.Destination)
+	if err != nil {
+		return err
+	}
+	if err := runInteractiveSSH(ctx, args); err != nil {
```
calvintvu: **[P2]** -  Check the attempt state before SSH starts

If the user cancels before Terminal starts the command, the attempt file contains `cancelled`. `RunAuthAttempt` checks expiration but does not check that state before it starts interactive SSH. The cancelled attempt still asks for credentials and checks cancellation only after authentication succeeds.

Reject cancelled attempts before the call to `runInteractiveSSH`.

A focused test reproduced this: `RunAuthAttempt` called the SSH runner after `CancelAuthAttempt` completed.

collector/internal/remote/auth.go
```diff
@@ -0,0 +1,384 @@
...
+}
+
+func CancelAuthAttempt(id string) error {
+	attempt, err := loadAuthAttempt(id)
+	if err != nil {
+		return err
+	}
+	_, err = updateAuthAttemptIfWaiting(id, AuthCancelled)
+	if err == nil {
+		exitAuthControlMaster(attempt.Destination)
```
calvintvu, line 202: **[P2]** -   Keep the SSH connection after authentication succeeds

`updateAuthAttemptIfWaiting` returns no error for an attempt that is already `ready`. This code then closes the shared SSH connection.

After authentication succeeds, the frontend keeps the attempt ID until the reconnect refresh ends. If the user closes Settings during that refresh, cleanup calls `CancelAuthAttempt` and closes the connection that the refresh needs. The host loses its successful authentication and requires another prompt.

Do not close the SSH connection when cancellation does not change a waiting attempt to cancelled.

A focused test reproduced this: cancelling a `ready` attempt called `exitAuthControlMaster`.

milanshen:
@agent-review: milan

milanshen (commented):

Review profile: Milan

Requested in [PR comment](https://github.com/centauri-ai/coslash/pull/172#issuecomment-5673114177)

frontend/src/pages/coslash/components/MachinesSettingsSection.tsx
luhe19001: Fixed in c584177. The auth-start request now receives the run’s AbortSignal, so Cancel aborts the server probe before it can create an attempt or launch Terminal. Added request-signal coverage.

collector/cmd/coslash/api_remote.go
luhe19001: Fixed in c584177. Password-only and keyboard-interactive permission failures now classify as authentication failures, enabling the guided auth flow while changed host keys remain blocked. Added focused coverage for both messages.

collector/internal/remote/auth.go
luhe19001: Fixed in c584177. RunAuthAttempt now rejects every non-waiting state before invoking interactive SSH. Added a regression test proving a cancelled attempt never calls the SSH runner.

collector/internal/remote/auth.go
luhe19001: Fixed in c584177. CancelAuthAttempt now acts only on a waiting attempt and closes the master only when the transition to cancelled succeeds. Added coverage that cancelling a ready attempt preserves the connection.

milanshen (commented):

Review profile: Milan

collector/internal/remote/manager.go
```diff
@@ -260,6 +260,7 @@ func (manager *Manager) ApplySettings(remote *settings.RemoteSettings) error {
 	previous := manager.cfg
 	aliasChanged := previous != nil && previous.SSHAlias != remote.SSHAlias
 	if aliasChanged {
+		CancelAuthAttemptsForDestination(previous.SSHAlias)
```
milanshen: **[P0]** — Cross-process flock taken while holding `manager.mu`

`ensureControlMaster` is now wrapped in `withDestinationCoordinator` (`sftp.go:160`) and holds the single `.auth-coordinator.lock` across both the control-master check *and* start — up to 2×`DefaultConnectTimeout` (~14s) — from a refresh goroutine that does **not** hold `manager.mu`.

If the user saves settings during that window, `ApplySettings` locks `manager.mu` at `manager.go:216` and then blocks on `syscall.Flock` here (and again at line 287, and in `removeLocked` at line 571). Every `/api/sessions` poll (`ListView`, line 369), `Retry`, and `DiagnosticsHealth` then queue behind `manager.mu` and the UI hangs for the full connect timeout.

`Shutdown` (line 527) deliberately drops `manager.mu` before calling the same function; `ApplySettings` and `removeLocked` should do the same — capture the alias under the lock, cancel attempts after releasing it.

collector/internal/remote/auth.go
```diff
@@ -0,0 +1,390 @@
...
+	if err := ensureSSHControlDir(); err != nil {
+		return err
+	}
+	lockPath := filepath.Join(settings.Home(), "ssh", ".auth-coordinator.lock")
+	lock, err := os.OpenFile(lockPath, os.O_WRONLY|os.O_CREATE, 0o600)
+	if err != nil {
+		return err
+	}
+	defer func() { _ = lock.Close() }()
+	if err := syscall.Flock(int(lock.Fd()), syscall.LOCK_EX); err != nil {
```
milanshen: **[P1]** — Blocking `LOCK_EX` ignores the request context

`syscall.Flock(..., LOCK_EX)` is not context-aware, so every HTTP handler that reaches `withDestinationCoordinator` can block well past its own deadline.

Concretely: `POST /api/remote/auth/cancel` → `CancelAuthAttempt` → `updateAuthAttemptIfWaiting` → here. If a refresh goroutine is inside `ensureControlMaster` while `ssh` negotiates a slow or unreachable host, the cancel handler blocks for up to ~14s even though the user just clicked Cancel. The 600ms status poll hits the same path and can pile up behind it. The `context.WithTimeout` on the request has no effect.

Consider a non-blocking `LOCK_EX|LOCK_NB` retry loop that honours `ctx.Done()`, so callers can give up.

frontend/src/pages/coslash/components/MachinesSettingsSection.tsx
```diff
@@ -201,11 +373,16 @@ export function MachinesSettingsSection({
                   Retry setup
                 </Button>
               ) : null}
+              {machine?.actionRequired === 'authenticate' && stage !== 'authenticating' ? (
```
milanshen: **[P1]** — Two identical "Authenticate in Terminal" buttons; this one is a no-op

This button calls `reconnect()` (line 263), which only sets `authAlias`/`authReconnect` and `stage = 'authentication_required'` — no terminal opens and no attempt is created.

After the click, `machine.actionRequired` is still `'authenticate'` and `stage` is not `'authenticating'`, so this button stays rendered while a **second identical** "Authenticate in Terminal" button appears in the message area at line 435. The user sees two same-labelled buttons and has to click the second one to actually launch the terminal.

Either have this button call `authenticateInTerminal()` after seeding the alias, or hide it once `stage === 'authentication_required'`.

frontend/src/pages/coslash/components/MachineActivity.tsx
```diff
@@ -26,6 +26,12 @@ function machineCopy(machine: MachineFact, checking: boolean) {
   }
   if (connectorFailed(machine))
     return `Setup failed: ${connectorFailureCopy(machine)}. Open Settings to retry.`;
+  if (machine.reason === 'authentication_failed') {
+    return `Authentication required. Saved history from ${savedHistory}. Open Settings to reconnect.`;
+  }
+  if (machine.reason === 'host_key_changed') {
```
milanshen, line 35: **[P2]** — `host_key_confirmation_required` missing from `machineCopy`

Branches were added for `authentication_failed` and `host_key_changed`, but not for `host_key_confirmation_required`. A first-time host with an unknown key classifies as `ReasonHostKeyConfirmation`, and `WithAuthenticationStatus` sets `actionRequired = 'authenticate'`, so the badge renders "Authentication required · Reconnect" (line 138) — but `machineCopy` falls through to the state branches and returns "Offline. Last checked …" or "Connection needs attention.", giving no hint that the host key needs confirming.

Note `needsAttention` (line 48) *does* list all three reasons, so the omission here looks unintentional.

collector/cmd/coslash/api_remote.go
milanshen: **[P2]** — A recovered host dead-ends the flow with a 409

(Distinct from the earlier password/keyboard-interactive classification thread, which is fixed.)

The user types an alias, the first test returns `actionRequired: 'authenticate'`, and the UI moves to `stage = 'authentication_required'`. If a control master becomes available before they click — another coSlash action, or an `ssh` they ran by hand — `TestAlias` now returns `StateOK`, `terminalAuthenticationPermitted` is false, and this returns 409 "Terminal authentication is not required for this host."

The frontend sets `stage = 'error'` with that text, and since the Authenticate button only renders at `stage === 'authentication_required'`, the user is left on an error with no path forward — even though the host is now ready to be added. Treating a healthy re-test as "proceed with setup" would be friendlier than a 409.

milanshen (commented):
Review of the guided SSH authentication flow at `c584177`. `go build`, `go vet`, and `go test ./internal/remote/... ./internal/settings/... ./cmd/...` all pass. The destination parsing, argv restructuring to `-l user host`, the reason split, and the ready/cancel handling in `RunAuthAttempt` all look sound.

Five findings, none overlapping the earlier review threads:

**P0 (blocking)**
1. `manager.go:263` — `ApplySettings`/`removeLocked` block on the new cross-process flock while holding `manager.mu`, freezing the whole manager for the control-master timeout.

**P1**
2. `auth.go:119` — the blocking `LOCK_EX` ignores the request context, so auth cancel/status handlers block past their own deadlines.
3. `MachinesSettingsSection.tsx:376` — the header "Authenticate in Terminal" button is a no-op relabel; a second identical button then renders.

**P2**
4. `MachineActivity.tsx:32` — `machineCopy` has no `host_key_confirmation_required` branch, so the tooltip contradicts the badge.
5. `api_remote.go:119` — a host that recovers between test and click dead-ends the flow with a 409.

Details inline.

Review profile: Milan

frontend/src/pages/coslash/components/MachineActivity.tsx
```diff
@@ -83,7 +95,10 @@ export function MachineActivity({
             isChecking(machine) || (machine.sourceId !== LOCAL_SOURCE_ID && remoteRetryInFlight);
           const retryable =
             machine.sourceId !== LOCAL_SOURCE_ID &&
+            machine.actionRequired == null &&
```
milanshen: **[P1]** — Changed host key can no longer be retried from the UI

`actionRequired == null` is now required for Retry. `WithAuthenticationStatus` sets `actionRequired: 'verify_host_key'` for `host_key_changed`, so this chip is disabled. Settings has no Retry either — only Remove.

On `main`, `host_key_failed` was retryable. After the user removes the offending `known_hosts` line (exactly what the new copy tells them to do), coSlash will not reconnect until the 3-minute `RemoteRetryInterval` background backoff, or they remove and re-add the host.

Keep Retry enabled for `verify_host_key`. Still do not offer terminal authentication for that reason.

frontend/src/pages/coslash/components/MachinesSettingsSection.tsx
```diff
@@ -34,13 +73,30 @@ export function MachinesSettingsSection({
...
   const [message, setMessage] = useState<string | null>(null);
   const [machine, setMachine] = useState<MachineFact | null>(null);
+  const [authAlias, setAuthAlias] = useState<string | null>(null);
+  const [authAttemptID, setAuthAttemptID] = useState<string | null>(null);
+  const [authReconnect, setAuthReconnect] = useState(false);
+  const authenticationRun = useRef<AuthenticationRun | null>(null);
   const busy = stage === 'testing' || stage === 'saving' || stage === 'installing' || stage === 'removing';
+  // Authentication must not use the dialog-wide busy state because that
+  // disables Cancel. Lock competing setup actions locally instead.
+  const setupActionsLocked = busy || stage === 'authenticating';
```
milanshen: **[P1]** — `authentication_required` is not covered by the setup-actions lock

`0b7c875` locked Add/alias/Remove during `authenticating`, which closed the in-flight `waitForAuthentication(A)` vs `addHost(B)` race. After a failed test the stage is `authentication_required`, so this guard is still false.

The alias field and **Add remote host** stay enabled, while Authenticate uses `authAlias` from the last test — not the current input. A user can edit `alice@host1` → `bob@host2` and click the banner button, then authenticate and save host1 while the field shows host2.

Include `authentication_required` in `setupActionsLocked`, or always start auth from `alias.trim()` / `remote.sshAlias`.

collector/internal/remote/manager.go
```diff
@@ -946,14 +956,19 @@ func classifyError(err error) Reason {
...
 		return ReasonInvalidData
+	case errors.Is(err, ErrAuthAttemptActive):
+		return ReasonAuthentication
 	}
 	message := strings.ToLower(err.Error() + " " + sshErrorStderr(err))
-	if strings.Contains(message, "host key verification failed") ||
-		strings.Contains(message, "remote host identification has changed") ||
-		strings.Contains(message, "no host key is known") {
-		return ReasonHostKey
+	if strings.Contains(message, "remote host identification has changed") || strings.Contains(message, "offending") {
```
milanshen: **[P2]** — `"offending"` is too broad for changed-host-key classification

`"remote host identification has changed"` already covers the OpenSSH MITM/changed-key warning. `strings.Contains(..., "offending")` also matches that message (`Offending ECDSA key in known_hosts`), but it matches unrelated stderr too — e.g. `Could not resolve hostname offending-box`.

Those hosts then get `verify_host_key`, which blocks guided auth and Retry.

Match something like `offending … key` / `known_hosts`, and add tests for changed vs unknown vs a hostname that merely contains `"offending"`.

frontend/src/pages/coslash/components/MachinesSettingsSection.tsx
```diff
@@ -94,7 +150,126 @@ export function MachinesSettingsSection({
...
+      );
+    }
+  };
+
+  const authenticateInTerminal = async () => {
+    if (authAlias == null || authenticationRun.current != null) return;
+    const run: AuthenticationRun = { cancelled: false, controller: new AbortController(), attemptID: null };
+    authenticationRun.current = run;
+    setStage('authenticating');
+    setMessage('Terminal opened. Complete the SSH prompt there; setup will continue automatically.');
```
milanshen: **[P2]** — Copy claims Terminal is open before `/auth/start` finishes

This message is set before `startRemoteAuthentication`. That RPC still runs `TestAlias` first (up to `DefaultConnectTimeout`, ~7s). A hanging probe shows “Terminal opened. Complete the SSH prompt there” with no window.

Keep “Checking SSH…” until launch succeeds, then switch to this copy.

milanshen (commented):
Additional review notes on `c584177`. Skipping findings already on the PR (flock/`manager.mu`, blocking `LOCK_EX`, the no-op header Authenticate button, `machineCopy` host-key confirmation, recovered-host 409).

No additional **P0** (blocking). Remaining items:

**P1**
- `MachineActivity.tsx` — `verify_host_key` disables Retry after the user has already verified the host.
- `MachinesSettingsSection.tsx` — `authentication_required` still leaves Add/alias unlocked (leftover of the `0b7c875` setup-actions guard).

**P2**
- `manager.go` — `classifyError` treats any stderr containing `"offending"` as a changed host key.
- `MachinesSettingsSection.tsx` — copy says Terminal is open before `/auth/start` finishes its probe.

Details inline.

collector/internal/remote/manager.go
luhe19001: Moved authentication cancellation and control-master teardown after releasing manager.mu, with a regression test covering blocked coordinator cleanup.

Resolved in `03ed2a80f2308f4b10c0ae0a1f7c3d64cb987dd9`.

collector/internal/remote/auth.go
luhe19001: Changed coordinator locking to a non-blocking retry loop that honors context cancellation, and threaded contexts through request-driven callers.

Resolved in `03ed2a80f2308f4b10c0ae0a1f7c3d64cb987dd9`.

frontend/src/pages/coslash/components/MachinesSettingsSection.tsx
luhe19001: The existing-host button now hides during authentication_required, leaving only the actionable banner button.

Resolved in `03ed2a80f2308f4b10c0ae0a1f7c3d64cb987dd9`.

frontend/src/pages/coslash/components/MachineActivity.tsx
luhe19001: Added specific guidance for first-time host-key confirmation.

Resolved in `03ed2a80f2308f4b10c0ae0a1f7c3d64cb987dd9`.

collector/cmd/coslash/api_remote.go
luhe19001: A healthy authentication re-test now returns not_required successfully, allowing the frontend to continue setup or mark reconnection complete.

Resolved in `03ed2a80f2308f4b10c0ae0a1f7c3d64cb987dd9`.

frontend/src/pages/coslash/components/MachineActivity.tsx
luhe19001: Retry is now enabled for verify_host_key while terminal authentication remains unavailable for changed keys.

Resolved in `03ed2a80f2308f4b10c0ae0a1f7c3d64cb987dd9`.

frontend/src/pages/coslash/components/MachinesSettingsSection.tsx
luhe19001: Setup actions, including alias editing and removal, are now locked during authentication_required.

Resolved in `03ed2a80f2308f4b10c0ae0a1f7c3d64cb987dd9`.

collector/internal/remote/manager.go
luhe19001: Narrowed changed-key detection to offending-key/known_hosts messages and added changed, unknown, and offending-hostname regression cases.

Resolved in `03ed2a80f2308f4b10c0ae0a1f7c3d64cb987dd9`.

frontend/src/pages/coslash/components/MachinesSettingsSection.tsx
luhe19001: The UI now shows Checking SSH until authentication launch succeeds, then reports that Terminal opened.

Resolved in `03ed2a80f2308f4b10c0ae0a1f7c3d64cb987dd9`.

collector/internal/remote/auth.go
```diff
@@ -0,0 +1,408 @@
...
+		return fmt.Errorf("authentication attempt is %s", attempt.State)
+	}
+	deadline := attempt.CreatedAt.Add(AuthAttemptTTL)
+	ctx, cancel := context.WithDeadline(ctx, deadline)
+	defer cancel()
+	args, err := interactiveMasterArgs(attempt.Destination)
+	if err != nil {
+		return err
+	}
+	if err := runInteractiveSSH(ctx, args); err != nil {
```
milanshen: **[P1] Cancel does not terminate the interactive SSH process**

`CancelAuthAttempt` changes the attempt file and runs `ssh -O exit`, but `RunAuthAttempt` remains synchronously blocked here in a separate Terminal process. Before authentication completes, OpenSSH has no control-master socket for `-O exit` to reach, so clicking Cancel while a password or host-key prompt is open leaves that prompt accepting input until the user closes it or the five-minute deadline expires, despite the app reporting cancellation.

Make the terminal runner observe cancellation state and cancel the SSH command context, or persist enough process information to signal the active process.

collector/cmd/coslash/api_remote.go
```diff
@@ -76,20 +87,87 @@ func handleRemoteTest(w http.ResponseWriter, request *http.Request, manager *rem
...
+	var body struct {
+		ID string `json:"id"`
+	}
+	decoder := json.NewDecoder(io.LimitReader(request.Body, 4<<10))
+	decoder.DisallowUnknownFields()
+	if err := decoder.Decode(&body); err != nil || decoder.Decode(&struct{}{}) != io.EOF || remote.CancelAuthAttempt(request.Context(), body.ID) != nil {
+		writeAPIError(w, http.StatusBadRequest, "invalid_authentication_attempt", "Authentication attempt is unavailable.")
+		return
+	}
+	writeJSON(w, remoteAuthResponse{ID: body.ID, State: remote.AuthCancelled})
```
milanshen: **[P2] Return the actual state when cancellation loses the race**

If a status poll marks the attempt `ready` immediately before this request, `CancelAuthAttempt` deliberately leaves the ready state and master intact, but this handler still responds with `state: "cancelled"`. That makes the API response contradict the persisted state and any subsequent status response.

Have cancellation return the resulting attempt state and encode that value here.

milanshen (commented):
## Review summary

**Carried over:** 16 resolved, 1 unresolved (17 prior findings).

| --- | --- | --- | --- |
| auth.go: status poll destroys successful master | P1 | Resolved | `RunAuthAttempt` treats an existing `ready` state as success and preserves the master. |
| settings schema rejects `user@host` | P2 | Resolved | Schema and parser now share the bounded alias-or-`user@host` grammar. |
| Settings dialog disables Cancel | P1 | Resolved | Authentication is excluded from dialog-wide `busy`; Cancel remains interactive. |
| Competing setup during authentication | P1 | Resolved | Setup controls and re-entrant `addHost` calls are blocked during authentication. |
| Auth-start cancellation permits late Terminal | P1 | Unresolved | The probe and attempt creation honor request cancellation, but `SSHAuthentication` calls context-free `osascript`; cancellation while Apple Events is blocked can still open Terminal late. |
| Password/keyboard-interactive classification | P2 | Resolved | Both permission-denied forms classify as authentication failures. |
| Cancelled attempt starts SSH | P2 | Resolved | `RunAuthAttempt` rejects non-waiting state before invoking SSH. |
| Cancelling ready attempt closes master | P2 | Resolved | `CancelAuthAttempt` preserves ready attempts and their master. |
| `manager.mu` held across coordinator lock | P0 | Resolved | Authentication cleanup runs after `manager.mu` is released. |
| Coordinator lock ignores context | P1 | Resolved | Nonblocking flock retries observe `ctx.Done()`. |
| Duplicate Authenticate buttons | P1 | Resolved | Existing-host action is hidden during authentication-required/authenticating stages. |
| Missing host-key confirmation copy | P2 | Resolved | `machineCopy` includes specific first-time confirmation guidance. |
| Healthy re-test dead-ends with 409 | P2 | Resolved | Healthy auth-start returns `not_required`, which the frontend handles. |
| Changed host key cannot retry | P1 | Resolved | `verify_host_key` remains retryable without enabling guided authentication. |
| Authentication-required setup remains editable | P1 | Resolved | `authentication_required` is included in `setupActionsLocked`. |
| Overbroad `offending` classification | P2 | Resolved | Detection now requires offending-key and `known_hosts` context. |
| Terminal-open copy shown too early | P2 | Resolved | The message changes only after auth-start returns successfully. |

The cross-layer guided SSH authentication design is generally sound, including structured destination parsing and coordinated control-master state. Cancellation remains incomplete around terminal launching and an already-running interactive SSH prompt. The 31-file, roughly 1,270-production-line change was reviewed as a static single-pass fallback because review workers and the Go/Node toolchains were unavailable; all production hunks and relevant tests were inspected, but tests and builds could not be executed.

**Findings:** 0 blocking (P0), 2 should-fix (P1), 1 nice-to-have (P2) — one P1 is carried over; new issues are in inline comments.

Review profile: Milan

collector/internal/remote/auth.go
luhe19001: The terminal runner now monitors the persisted attempt state and cancels the active SSH context when cancellation is recorded. Added a regression test covering a blocked interactive process.

Resolved in `cf18f9af4e057eb1d05f3817dc4c9a7adf214b25`.

collector/cmd/coslash/api_remote.go
luhe19001: Cancellation now returns the authoritative resulting state, and the API encodes that state instead of always reporting cancelled. The ready-state race test now verifies this behavior.

Resolved in `cf18f9af4e057eb1d05f3817dc4c9a7adf214b25`.

milanshen (commented):
## Review summary

**Score: 3/5** — not reviewed (too large)

This PR exceeds the 15-minute review budget: 31 files and 2,002 changed lines, including 18 production-source files with approximately 1,290 additions/deletions. Split it into backend SSH authentication/control-master and API work, frontend guided-authentication UI, and validation/documentation changes. The 20 prior findings were not re-evaluated because the size preflight requires the PR to be split first.

**Findings:** 0 blocking (P0), 0 should-fix (P1), 0 nice-to-have (P2) — review not completed due to scope.

Review profile: Milan

luhe19001:
CI resolver made no source change for failed check(s): Frontend. No source-level cause was established. The supplied warnings are pre-existing, while the only failure annotation is a generic exit code. Full reproduction was blocked by unavailable dependencies and an incompatible local Node version, so the worktree remains unchanged.

milanshen:
@agent-review: milan

milanshen (commented):
**Carried over:** 0 resolved, 20 unresolved (20 prior findings).

| --- | --- | --- | --- |
| `auth.go` — status poll tears down ready master | P1 | Unresolved | Claimed fixed in `2330743`; not re-evaluated because size preflight stopped the review. |
| `settings.go:199` — schema rejects `user@host` | P2 | Unresolved | Claimed fixed in `2330743`; not re-evaluated. |
| `MachinesSettingsSection.tsx` — authentication disables Cancel | P1 | Unresolved | Claimed fixed in `6344352`; not re-evaluated. |
| `MachinesSettingsSection.tsx:80` — competing host setup during authentication | P1 | Unresolved | Claimed fixed in `0b7c875`; not re-evaluated. |
| `MachinesSettingsSection.tsx` — cancellation does not abort authentication start | P1 | Unresolved | Claimed fixed in `c584177`; not re-evaluated. |
| `api_remote.go:123` — password-only hosts rejected | P2 | Unresolved | Claimed fixed in `c584177`; not re-evaluated. |
| `auth.go` — cancelled attempt starts SSH | P2 | Unresolved | Claimed fixed in `c584177`; not re-evaluated. |
| `auth.go:202` — cancelling ready attempt closes connection | P2 | Unresolved | Claimed fixed in `c584177`; not re-evaluated. |
| `manager.go` — coordinator lock held with `manager.mu` | P0 | Unresolved | Claimed fixed in `03ed2a80`; not re-evaluated. |
| `auth.go` — flock ignores request context | P1 | Unresolved | Claimed fixed in `03ed2a80`; not re-evaluated. |
| `MachinesSettingsSection.tsx` — duplicate authentication buttons | P1 | Unresolved | Claimed fixed in `03ed2a80`; not re-evaluated. |
| `MachineActivity.tsx:35` — missing host-key confirmation copy | P2 | Unresolved | Claimed fixed in `03ed2a80`; not re-evaluated. |
| `api_remote.go:123` — recovered host dead-ends with 409 | P2 | Unresolved | Claimed fixed in `03ed2a80`; not re-evaluated. |
| `MachineActivity.tsx` — changed host key cannot be retried | P1 | Unresolved | Claimed fixed in `03ed2a80`; not re-evaluated. |
| `MachinesSettingsSection.tsx` — setup lock omits authentication-required stage | P1 | Unresolved | Claimed fixed in `03ed2a80`; not re-evaluated. |
| `manager.go` — overly broad `offending` classification | P2 | Unresolved | Claimed fixed in `03ed2a80`; not re-evaluated. |
| `MachinesSettingsSection.tsx` — UI reports Terminal opened prematurely | P2 | Unresolved | Claimed fixed in `03ed2a80`; not re-evaluated. |
| `MachinesSettingsSection.tsx` — alias editable before authentication starts | P1 | Unresolved | Claimed fixed in `03ed2a80`; not re-evaluated. |
| `auth.go` — Cancel leaves interactive SSH running | P1 | Unresolved | Claimed fixed in `cf18f9af`; not re-evaluated. |
| `api_remote.go` — cancellation returns stale state | P2 | Unresolved | Claimed fixed in `cf18f9af`; not re-evaluated. |

## Review summary

**Score: 3/5** — not reviewed (too large)

The size-only preflight found 31 changed files and 2,002 changed lines, including approximately 1,290 changed production-source lines across backend SSH coordination, APIs, and frontend guided-authentication behavior. This exceeds both the 25-file and 1,000-production-line guardrails, so the full diff and prior fixes could not be reviewed reliably within the 15-minute budget. Split the work into backend authentication/control-master and API changes, frontend guided-authentication UI, and validation/documentation updates.

**Findings:** 0 newly reported blocking (P0), 0 newly reported should-fix (P1), 0 newly reported nice-to-have (P2) — review not completed due to scope.

Review profile: Milan

Requested in [PR comment](https://github.com/centauri-ai/coslash/pull/172#issuecomment-5705286104)

collector/cmd/coslash/api_remote.go
```diff
@@ -76,20 +87,92 @@ func handleRemoteTest(w http.ResponseWriter, request *http.Request, manager *rem
...
+		return
+	}
+	ctx, cancel := context.WithTimeout(request.Context(), remote.DefaultConnectTimeout+5*time.Second)
+	defer cancel()
+	health, err := manager.TestAlias(ctx, body.SSHAlias)
+	if err == nil && health.State == remote.StateOK {
+		writeJSON(w, remoteAuthResponse{State: remote.AuthNotRequired})
+		return
+	}
+	if err != nil || !terminalAuthenticationPermitted(health) {
```
milanshen: **[P1] Auth start 409s non-auth probe failures as “not required”**

`handleRemoteAuthStart` re-runs `TestAlias` and maps every non-OK outcome that is not authentication/host-key confirmation onto 409 `authentication_not_required` with “Terminal authentication is not required for this host.”

Concrete: Add host → first test returns `authentication_failed` → user clicks Authenticate in Terminal. The second BatchMode probe then hits a timeout, `connection_failed`, or `host_key_changed`. This still 409s. The frontend sets `stage = 'error'` with that message. On the add-host path `remote` is still null, so the header Authenticate button is hidden and the banner button only renders at `authentication_required` — the user is stuck until they click Add again, and Terminal never opened.

Return a reason-specific error, or skip the re-probe once the UI already classified the host as authenticate. Do not use `authentication_not_required` for connection, timeout, or changed-key failures.

collector/internal/remote/manager.go
```diff
@@ -946,14 +967,20 @@ func classifyError(err error) Reason {
...
-		return ReasonHostKey
+	if strings.Contains(message, "remote host identification has changed") ||
+		(strings.Contains(message, "offending ") && strings.Contains(message, " key") && strings.Contains(message, "known_hosts")) {
+		return ReasonHostKeyChanged
+	}
+	if strings.Contains(message, "host key verification failed") || strings.Contains(message, "no host key is known") {
+		return ReasonHostKeyConfirmation
 	}
 	if strings.Contains(message, "permission denied (publickey") ||
+		strings.Contains(message, "permission denied (password") ||
```
milanshen: **[P1] Permission-denied method lists that do not start with publickey/password/keyboard-interactive never offer Terminal auth**

The new prefixes only match `permission denied (password` and `permission denied (keyboard-interactive`. OpenSSH prints offered methods in server order. `Permission denied (gssapi-with-mic,password).` contains neither prefix (`(password` is not at the start of the list), so `classifyError` returns `connection_failed`.

Add host then shows a generic SSH error with no Authenticate in Terminal button, and Add again still uses `BatchMode=yes`, so a GSSAPI-first password host cannot be added.

Treat any SSH `permission denied (` method list as `ReasonAuthentication`. Path permission is already classified earlier via `fs.ErrPermission` / `ErrPathDenied`.

frontend/src/pages/coslash/components/MachinesSettingsSection.tsx
```diff
@@ -94,7 +150,139 @@ export function MachinesSettingsSection({
...
+    if (run != null) {
+      run.cancelled = true;
+      run.controller.abort();
+      if (run.attemptID != null) await cancelRemoteAuthentication(run.attemptID).catch(() => undefined);
+    } else if (authAttemptID != null) {
+      await cancelRemoteAuthentication(authAttemptID).catch(() => undefined);
+    }
+    setAuthAttemptID(null);
+    setStage('error');
+    setMessage('Authentication was cancelled. Try again.');
```
milanshen, line 283: **[P1] Cancel after a ready attempt still marks Settings as failed**

After Terminal reports `ready`, `waitForAuthentication` awaits `retryRemoteRefreshAndWait` while `stage` remains `authenticating`, so Cancel stays visible. Clicking it always sets `stage = 'error'` and this copy, even when `CancelAuthAttempt` returns `ready` and leaves the master up.

A user who just completed the password prompt, then clicks Cancel while the reconnect refresh is still running, sees a failed Settings card while the board can already show the host as connected.

If cancel returns `ready` (or status is already ready), keep the success path; hide Cancel once the attempt is no longer `waiting`.

milanshen (commented):
## Review summary

**Carried over:** 19 resolved, 1 unresolved (20 prior findings).

| --- | --- | --- | --- |
| `auth.go` — status poll tears down ready master | P1 | Resolved | `RunAuthAttempt` treats an existing `ready` state as success and does not close the master. |
| `settings.go` — schema rejects `user@host` | P2 | Resolved | Schema and `ParseSSHDestination` share the bounded alias-or-`user@host` grammar. |
| Settings dialog disables Cancel during authentication | P1 | Resolved | `authenticating` is excluded from dialog-wide `busy`; Cancel stays clickable. |
| Competing host setup during authentication | P1 | Resolved | Setup controls are locked while `authenticating`. |
| Auth-start cancellation still opens Terminal / leaves a waiting file | P1 | Unresolved | The start fetch is aborted, but after `CreateAuthAttempt` the handler still calls context-free `SSHAuthentication`, and `CancelAuthAttempt(request.Context())` returns immediately when the client is already gone. Cancel/unmount during “Checking SSH…” has `attemptID == null`, so the waiting file can block refresh for the full TTL. |
| Password/keyboard-interactive classification | P2 | Resolved | `(password` and `(keyboard-interactive` method lists classify as authentication; tests cover those strings. |
| Cancelled attempt starts SSH | P2 | Resolved | `RunAuthAttempt` rejects every non-waiting state before invoking SSH. |
| Cancelling a ready attempt closes the master | P2 | Resolved | `CancelAuthAttempt` returns the existing state and only exits the master on a waiting→cancelled transition. |
| `manager.mu` held across coordinator lock | P0 | Resolved | Authentication cancellation and master teardown run after `manager.mu` is released. |
| Coordinator lock ignores context | P1 | Resolved | Non-blocking flock retries observe `ctx.Done()`. |
| Duplicate Authenticate in Terminal buttons | P1 | Resolved | The existing-host action is hidden during `authentication_required` / `authenticating`. |
| Missing host-key confirmation copy | P2 | Resolved | `machineCopy` includes first-time confirmation guidance. |
| Healthy re-test dead-ends with 409 | P2 | Resolved | A healthy auth-start returns `not_required`, which the frontend handles. |
| Changed host key cannot be retried | P1 | Resolved | `verify_host_key` remains retryable without enabling guided authentication. |
| Setup lock omits `authentication_required` | P1 | Resolved | `setupActionsLocked` includes `authentication_required`. |
| Overbroad `offending` classification | P2 | Resolved | Detection requires offending-key and `known_hosts` context; hostname `offending-box` is tested. |
| Terminal-open copy shown too early | P2 | Resolved | The message stays on “Checking SSH…” until auth-start returns. |
| Alias editable before authentication starts | P1 | Resolved | The alias field is disabled via `setupActionsLocked`. |
| Cancel leaves interactive SSH running | P1 | Resolved | The terminal runner watches the attempt file and cancels the SSH context when the attempt is cancelled. |
| Cancellation API always reported `cancelled` | P2 | Resolved | Cancellation returns the authoritative resulting state, and the handler encodes it. |

This PR adds a guided Terminal SSH-auth flow: short-lived attempt files, a destination coordinator around control-master start, reason-split host-key handling, and Settings UI that launches `coslash ssh-auth`. The control-master lifecycle and most earlier races look sound. The remaining merge blocker is still cancellation around auth-start, plus classification/start-probe gaps that prevent Terminal from opening or mark a successful reconnect as failed.

**Findings:** 0 blocking (P0), 4 should-fix (P1), 0 nice-to-have (P2) — one P1 is carried over; new issues are in inline comments.

Phase 1 used parallel angle specialists (A–I); Phase 2 deduped and verified those candidates against current HEAD before posting.

The unresolved start-cancel leak is cross-cutting: a leftover `AuthWaiting` file makes `AuthAttemptActive` skip refresh/retry and makes `ensureControlMaster` refuse BatchMode SSH for the five-minute TTL, even when a ControlPersist socket is already live.

collector/cmd/coslash/api_remote.go
luhe19001: Auth start now returns the probe’s specific reason and message for connection, timeout, and changed-key failures instead of reporting authentication_not_required. Added regression coverage for each case.

Resolved in `968ddf63dc7b0ac2fa19dc31200cbdc24710bf94`.

collector/internal/remote/manager.go
luhe19001: Authentication classification now accepts any OpenSSH permission-denied method list, including GSSAPI-first lists. Added the reported ordering as a regression case.

Resolved in `968ddf63dc7b0ac2fa19dc31200cbdc24710bf94`.

frontend/src/pages/coslash/components/MachinesSettingsSection.tsx
luhe19001: Cancel is now shown only while an attempt is waiting. If cancellation races with a ready result, the successful reconnect flow continues instead of showing an error.

Resolved in `968ddf63dc7b0ac2fa19dc31200cbdc24710bf94`.

milanshen (commented):
## Review summary

**Carried over:** 22 resolved, 1 unresolved (23 prior findings).

| --- | --- | --- | --- |
| `auth.go` — status poll tears down ready master | P1 | Resolved | `RunAuthAttempt` still treats an existing `ready` state as success and keeps the master. |
| `settings.go` — schema rejects `user@host` | P2 | Resolved | Schema and `ParseSSHDestination` still share the alias-or-`user@host` grammar. |
| Settings dialog disables Cancel during authentication | P1 | Resolved | `authenticating` remains excluded from dialog-wide `busy`. |
| Competing host setup during authentication | P1 | Resolved | Setup controls stay locked while `authenticating`. |
| Auth-start cancellation still opens Terminal / leaves a waiting file | P1 | Unresolved | `968ddf63` did not change this path. After `CreateAuthAttempt`, `SSHAuthentication` is still context-free, and `CancelAuthAttempt(request.Context())` no-ops if the client already aborted. Unmount/abort during “Checking SSH…” still has `attemptID == null`, so the waiting file is not cancelled. The banner Cancel button is now hidden until an ID exists, so only dialog dismiss aborts the probe. |
| Password/keyboard-interactive classification | P2 | Resolved | Method lists still classify as authentication; GSSAPI-first is now included (see below). |
| Cancelled attempt starts SSH | P2 | Resolved | `RunAuthAttempt` still rejects non-waiting state before SSH. |
| Cancelling a ready attempt closes the master | P2 | Resolved | `CancelAuthAttempt` still returns the existing state and only exits the master on waiting→cancelled. |
| `manager.mu` held across coordinator lock | P0 | Resolved | Teardown still runs after `manager.mu` is released. |
| Coordinator lock ignores context | P1 | Resolved | Non-blocking flock retries still observe `ctx.Done()`. |
| Duplicate Authenticate in Terminal buttons | P1 | Resolved | Header action still hidden during `authentication_required` / `authenticating`. |
| Missing host-key confirmation copy | P2 | Resolved | `machineCopy` still has first-time confirmation guidance. |
| Healthy re-test dead-ends with 409 | P2 | Resolved | StateOK still returns `not_required` before the probe-failure branch. |
| Changed host key cannot be retried | P1 | Resolved | `verify_host_key` remains retryable. |
| Setup lock omits `authentication_required` | P1 | Resolved | `setupActionsLocked` still includes it. |
| Overbroad `offending` classification | P2 | Resolved | Detection still requires offending-key and `known_hosts` context. |
| Terminal-open copy shown too early | P2 | Resolved | Copy still stays on “Checking SSH…” until auth-start returns. |
| Alias editable before authentication starts | P1 | Resolved | Alias field still disabled via `setupActionsLocked`. |
| Cancel leaves interactive SSH running | P1 | Resolved | Terminal runner still cancels the SSH context when the attempt file is cancelled. |
| Cancellation API always reported `cancelled` | P2 | Resolved | Handler still encodes the authoritative resulting state. |
| Auth start 409s non-auth probe failures as “not required” | P1 | Resolved | Probe failures now 409 with the classified reason/message (`connection_failed`, `refresh_timeout`, `host_key_changed`); covered by `TestRemoteAuthStartReportsProbeFailureReason`. |
| GSSAPI-first permission-denied lists never offer Terminal auth | P1 | Resolved | `classifyError` matches any `permission denied (`; `Permission denied (gssapi-with-mic,password).` is a regression test. |
| Cancel after a ready attempt marks Settings as failed | P1 | Resolved | Leaving `waiting` clears `attemptID` before reconnect/save; Cancel is hidden without an ID; a stale click returns if cancel reports `ready` instead of forcing `stage = 'error'`. |

`968ddf63` fixes the three P1s from the previous review: probe failures keep their reason, GSSAPI-first method lists classify as authentication, and Cancel after ready no longer forces a failed Settings card. The guided SSH-auth design remains sound. The one remaining should-fix is still auth-start cancellation: aborting during the pre-Terminal probe can leave a waiting attempt and still launch Terminal.

**Findings:** 0 blocking (P0), 1 should-fix (P1), 0 nice-to-have (P2) — the P1 is carried over; no new inline comments.

Phase 1 used parallel leaf workers to re-verify prior findings and scan the diff since `6345601`; Phase 2 checked those results against `968ddf63` before posting. Author replies claiming fixes were treated as hypotheses.

A leftover `AuthWaiting` file from the unresolved start-cancel leak still makes `AuthAttemptActive` skip refresh/retry for the five-minute TTL.

milanshen:
Not safe yet; shall be rebased after the 11 commits (beta bundle) got reverted (or cherry-picked)

luhe19001:
CI resolver made no source change for failed check(s): plan. No source-level cause was established. The failed plan job belongs to an unchanged release workflow, and its only annotation is a generic exit code at blank line 189 without event inputs or command output. The offline dry-run path passes; publishing paths require GitHub API state that cannot be inspected under the no-contact constraint.

luhe19001:
Closing in favor of a clean replacement rebased onto the current main branch. The replacement branch is `hlu/guided-ssh-auth-rebased`; a new PR will follow.

PR #181, opened 2026-09-15
Make vendor session metadata extensible

calvintvu:
## Context

This extensible vendor metadata change lets collectors add session information that does not exist in transcripts.

Previously, vendor metadata supported only two side-loaded fields:
- Session name
- Live status

Cursor stores many important fields in separate databases rather than transcripts:
- Name and summary
- IDE, CLI, or SDK entrypoint
- Working directory
- Model
- Context usage and window
- Recorded cost and SDK tokens
- File changes and commits
- Parent-child relationships
- Creation and update times

The new SessionEnrichment record groups this information by session ID. The shared collector then applies it to the unified session model.

**TLDR**: it creates one standard path for combining transcripts with vendor side stores. Claude, Codex, and OpenCode keep the same behavior, while Cursor can provide its additional metadata without adding many Cursor-specificmaps and resolver paths.

## Changes

- Replace separate name and live-status maps with one extensible metadata record per session.
- Add shared resolution for models, usage, pull requests, summaries, and entrypoints.
- Migrate local and remote Claude, Codex, and OpenCode collection to the new metadata interface.

## Test

- `git diff --check` — passed
- Ensure no regressions in UI, there should be no behavioral changes


milanshen (commented):

Review profile: Milan

milanshen:
@agent-review

milanshen:

milanshen:
@agent-review

luhe19001:

luhe19001 (commented):
## Review summary

The automated review could not run because the repository sandbox failed before the required size preflight (`bwrap: No permissions to create new namespace`). No diff was inspected, so merge safety cannot be assessed; rerun the review in a functioning checkout environment.

**Findings:** 0 blocking (P0), 0 should-fix (P1), 0 nice-to-have (P2) — review not completed.

Review profile: Helu

Requested in [PR comment](https://github.com/centauri-ai/coslash/pull/181#issuecomment-5688472665)

milanshen:
@agent-review

milanshen:

milanshen:
@agent-review

luhe19001:

collector/internal/vendors/metadata.go
```diff
@@ -1,19 +1,65 @@
...
+}
 
 // SessionMetadata carries raw liveness signals, not final status strings: the final
 // status needs parse output (InTurn, lastActivity), so resolveStatus computes it.
 // A Live value of "interactive" gets the busy/idle refinement; anything else
 // passes through. Absent means there is no external live signal; parsed data
 // may still provide a persisted or derived status hint.
+type SessionEnrichment struct {
+	Name, Live, Summary, Entrypoint, WorkingDirectory, Model string
+	StartedAt, LastActivityAt                                int64
```
milanshen, line 23: **[P1] Apply every advertised enrichment field**

The new record exposes working directory, timestamps, file edits, commit observations, and relationship data, but none of those fields is consumed anywhere outside this declaration. If Cursor supplies a child's parent ID and timestamps only from its side database (the use case motivating this PR), finalization fabricates timestamps before reading metadata, composes that child as an extra root, and leaves cwd/edits/commits empty. Add a pre-composition resolver that maps these values into `ParsedSession`/`Session` before activity fallbacks, or remove the fields until they are supported.

collector/internal/collector/collector.go
```diff
@@ -129,6 +132,7 @@ func finalizeSessionsSource(
 		removeUnresolvedSpawnRows(p.Session)
 	}
 	resolveNames(composition.roots, metadata)
+	resolveSummariesAndEntrypoints(composition.roots, metadata)
```
milanshen: **[P1] Resolve child summaries before subagent projection**

`resolveSummariesAndEntrypoints` sees only roots and runs after `enrichSubagents`. When a child summary exists only in `SessionEnrichment`, `subagentFrom` has already copied the transcript's empty or stale summary into the parent's result, and this resolver never visits the child, so the side-loaded result is lost. Resolve session-level summaries across `composition.parsed` before `enrichSubagents` (and keep only entrypoint projection root-scoped if that is intentional).

collector/internal/remotefacts/facts.go
```diff
@@ -447,13 +447,21 @@ func FromParsed(vendor, familyID, parserVersion, state, staleReason string, pars
 		sessionIDs[item.ID] = true
 	}
 	if metadata != nil {
-		for id, name := range metadata.Names {
+		for id, enrichment := range metadata.Sessions {
```
milanshen, line 453: **[P1] Carry enriched metadata through remote facts**

Only `Name` and `Live` are emitted from the new session records here; `Family.Parsed`, `mergeMetadataInto`, and `aggregateFingerprint` likewise preserve/hash only those two fields. If a remote side store provides or changes summary, model, usage, or entrypoint while the transcript is unchanged, the helper either reports the family unchanged or omits the values even on recollection, so the remote card and cached restart never receive them. Extend the remote schema, validation, parse/merge, and fingerprint paths for every supported enrichment field, or explicitly keep this contract local-only.

collector/internal/vendors/metadata.go
```diff
@@ -1,19 +1,65 @@
...
-	Live  map[string]string // id → raw status
+	Sessions map[string]*SessionEnrichment
 }
 
 func EmptySessionMetadata() *SessionMetadata {
-	return &SessionMetadata{Names: map[string]string{}, Live: map[string]string{}}
+	return &SessionMetadata{Sessions: map[string]*SessionEnrichment{}}
+}
+
+func (m *SessionMetadata) Session(id string) *SessionEnrichment {
```
milanshen, line 37: **[P2] Avoid allocating enrichment during reads**

`Session` is used as a getter throughout finalization and fingerprinting, but it inserts an empty record on every miss. With sparse metadata and many transcript sessions, a read-only pass grows `Sessions` to every session and makes later metadata loops scan those empty entries. Provide a non-mutating lookup for read paths and reserve this allocating accessor for writers.

milanshen (commented):
## Review summary

The migration preserves the existing name and liveness behavior, and the overall per-session enrichment shape is a sound direction. The new contract is not yet applied end to end, so vendor-only metadata can still be dropped during local composition or remote collection.

**Findings:** 0 blocking (P0), 3 should-fix (P1), 1 nice-to-have (P2) — see inline comments.

The collector test suite passes, but the metadata parity coverage currently exercises only names; it does not cover the newly supported fields, child projection, or remote cache round-trips.

collector/internal/collector/collector.go
calvintvu: Agreed and addressed in 6fb1ce7. Session enrichment now runs across all parsed sessions before composeSessions and enrichSubagents, so a child summary from side-loaded metadata is present when subagentFrom projects its result. The remote parity regression now covers this case.

collector/internal/vendors/metadata.go
calvintvu: Agreed and addressed in 6fb1ce7. Enrichment is now applied to every parsed session before timestamp fallbacks and composition, so working directory, timestamps, file edits, model, usage, pull requests, summary, and entrypoint reach the final session. I removed relationship and raw commit-observation fields for now because they had no defined downstream semantics; they can be added with the vendor implementation that consumes them.

collector/internal/remotefacts/facts.go
calvintvu: Agreed and addressed in 6fb1ce7. remotefacts.FromParsed applies enrichment before encoding the existing session fact fields, so cache round-trips preserve the enriched values. aggregateFingerprint now hashes the complete enrichment record, ensuring metadata-only changes trigger recollection. The parity test covers summary, model and usage, entrypoint, cwd, timestamps, file edits, pull requests, and child summary.

collector/internal/vendors/metadata.go
calvintvu: Agreed and addressed in 6fb1ce7. Added a non-mutating Lookup accessor and converted finalization, filtering, subagent status, and fingerprint read paths to use it. The allocating Session accessor remains limited to writer paths, with a regression proving sparse reads do not grow the metadata map.

milanshen (commented):
**Carried over:** 3 resolved, 1 unresolved (4 prior findings).

| --- | --- | --- | --- |
| metadata.go:23 — Apply every enrichment field | P1 | Resolved | Remaining enrichment fields are applied before fallbacks and composition; unsupported relationship and commit fields were removed. |
| collector.go — Resolve child summaries before projection | P1 | Resolved | Enrichment now precedes composition and subagent projection. |
| facts.go:453 — Preserve enrichment through remote facts | P1 | Unresolved | Helper fingerprints include enrichment, but direct SFTP still identifies unchanged families using file-only fingerprints. Fresh metadata affects the immediate card but not the durable snapshot, so restart restores stale enrichment. |
| metadata.go:37 — Avoid allocation during reads | P2 | Resolved | Non-mutating Lookup is used by finalization, filtering, status, and fingerprint reads. |

## Review summary

This PR centralizes session enrichment and correctly applies it before composition. The remaining issue is durable metadata-only updates in the direct SFTP path. No additional findings were identified; `git diff --check` passed, but Go tests could not run because the toolchain is unavailable.

**Findings:** 0 blocking (P0), 1 should-fix (P1), 0 nice-to-have (P2) — see the carried-over inline comment.

Review profile: Milan

milanshen (approved):

PR #182, opened 2026-09-15
Represent unknown session cost and usage

calvintvu:
## Context

Sessions without token or cost data currently appear to have used zero tokens at no cost. This is misleading because zero is a valid recorded value. This change represents unavailable cost separately from a known zero and displays unknown usage clearly.

## Changes

- Represent unavailable session and subagent cost as null while preserving recorded zero and token-derived estimates.
- Display an em dash when token or cost information is unavailable.
- Sum known costs without presenting missing data as zero.
- Preserve the existing snapshot v1 schema.

## Test

- `npm test` — 94 tests passed
- `npm run build` — passed
- `git diff --check` — passed

### Screenshots

- [x] Session cards and totals — unavailable tokens and cost display as an em dash
Screenshot 2026-09-14 at 17 17 11


luhe19001 (commented):

Review profile: Helu

milanshen (commented):
## Review summary

This PR establishes the right nullable model in the collector and updates most card and aggregate displays, but the unknown-value invariant is not yet preserved across every frontend consumer and the sharing boundary. The implementation and test suite are otherwise coherent; completing that propagation is the main work remaining before merge.

collector/internal/sessionexport/snapshot.go
```diff
@@ -392,8 +392,11 @@ func (b *builder) git(value *session.GitDrift) *snapshotv1.GitDrift {
 	return &snapshotv1.GitDrift{BaseBranch: b.text("/session/git/baseBranch", value.BaseBranch, maxBranchBytes), Ahead: value.Ahead, Behind: value.Behind}
 }
 
-func (b *builder) usage(tokens map[string]session.ModelTokens, cost float64, unpriced []string, path string) (snapshotv1.Usage, error) {
-	estimated, err := snapshotv1.CostMicroUSD(cost)
+func (b *builder) usage(tokens map[string]session.ModelTokens, cost *float64, unpriced []string, path string) (snapshotv1.Usage, error) {
+	if cost == nil {
+		cost = new(float64)
```
milanshen, line 397: **[P1] Preserve unknown cost at the sharing boundary**

When a session has neither recorded cost nor token rows, `AttachCost` now correctly leaves `Cost == nil`, but this branch converts that state to `0` before snapshot serialization. The same session therefore shows an em dash on the board while its preview/upload is indistinguishable from a known zero-cost session (the preview renders `estimatedCostMicroUsd: 0` as `$0.00`). If snapshot v1 must remain unchanged, either make unknown-cost sessions explicitly ineligible for this export or add a compatible availability signal/new schema version instead of silently coercing nil to zero; add a nil-cost snapshot regression test.

frontend/src/pages/coslash/components/SessionSortDropdownMenu.tsx
```diff
@@ -37,9 +37,9 @@ function sortValue(session: Session, key: SortKey): number {
     case SortKey.Status:
       return STATUS_ORDER.length - STATUS_ORDER.indexOf(boardStatusKey(session));
     case SortKey.Value:
-      return session.cost;
+      return session.cost ?? 0;
```
milanshen: **[P1] Keep unknown values distinct while sorting**

Coalescing `null` to `0` makes an unavailable cost tie with an authoritative zero; under ascending cost sort, unknown sessions are placed among the cheapest sessions instead of remaining unknown. Line 42 has the same issue for an empty token map versus a known zero-token row. Handle nullability in the comparator and place unknowns after known values for both sort directions, then compare only the known numbers.

frontend/src/pages/coslash/lib/session.ts
```diff
@@ -28,7 +28,7 @@ export type Subagent = {
   toolUses: number;
   commands: SubagentCommand[];
   tokens: Record<string, ModelTokens>;
-  cost: number;
+  cost: number | null;
```
milanshen, line 31: **[P1] Make the top-level session cost nullable too**

This updates `Subagent.cost`, but the top-level `Session.cost` at line 71 still declares `number` even though the Go API now serializes `null` for sessions without cost or token data. `decodeSession` casts the response to `Session`, so such a valid payload violates the frontend contract and future consumers receive no compiler warning before doing number-only operations. Mirror this change on `Session.cost` with `number | null`.

frontend/src/pages/coslash/lib/session.ts
```diff
@@ -448,7 +448,8 @@ export function sumTokens(
   return Object.values(tokens).reduce((sum, modelTokens) => sum + modelTokens[key], 0);
 }
 
-export function getTotalTokens(tokens: Session['tokens']): number {
+export function getTotalTokens(tokens: Session['tokens']): number | null {
+  if (Object.keys(tokens).length === 0) return null;
```
milanshen, line 462: **[P1] Propagate unknown tokens to the detailed breakdown**

The new total correctly returns `null` for `{}`, but `TokenBreakdown` still calls `sumTokens` four times, and that helper reduces the same empty map to `0`. Opening a no-token session or subagent therefore shows an unknown total (`— tok`) alongside `in 0k · out 0k · cache 0kr / 0kw`. Have the breakdown detect an empty map and render an unavailable state (while preserving zero for a nonempty all-zero token row), and cover that case with a component/helper test.

luhe19001:

frontend/src/pages/coslash/lib/session.ts
calvintvu: Agreed and fixed in f2a81a9. `TokenBreakdown` now renders an em dash for an empty token map while retaining numeric zeroes for a nonempty all-zero row. Added a component regression test against the rendered output.

collector/internal/sessionexport/snapshot.go
calvintvu: Agreed and fixed in f2a81a9. Completed sessions with an unknown cost are now marked non-shareable, and the preview/upload boundary rejects `Cost == nil` before snapshot v1 serialization, so unknown can no longer become an authoritative zero. Added a regression test for the non-approvable preview state.

frontend/src/pages/coslash/lib/session.ts
calvintvu: Agreed and fixed in f2a81a9. Top-level `Session.cost` is now `number | null`; the resulting compiler check also required the aggregate reducer to declare its numeric accumulator explicitly.

frontend/src/pages/coslash/components/SessionSortDropdownMenu.tsx
calvintvu: Agreed and fixed in f2a81a9. The comparator now preserves nullable cost/token values and places unknowns after known values for both ascending and descending sorts. Regression coverage distinguishes unknown from authoritative zero for both sort keys.

frontend/src/pages/coslash/components/SessionBoard.tsx
```diff
@@ -42,8 +42,14 @@ function StatusColumnHeader({ status, sessions }: { status: Status; sessions: Se
 
 function GroupTotals({ sessions }: { sessions: Session[] }) {
   const aggregate = sessionsForAggregates(sessions);
-  const tokens = aggregate.reduce((sum, session) => sum + getTotalTokens(session.tokens), 0);
-  const cost = aggregate.reduce((sum, session) => sum + session.cost, 0);
+  const tokenCounts = aggregate.map((session) => getTotalTokens(session.tokens));
+  const tokens = tokenCounts.some((count) => count != null)
```
luhe19001: **[P1] Do not publish partial usage as a total**

`some` selects the numeric branch as soon as one session has known usage, while `(count ?? 0)` and `(value ?? 0)` silently treat every remaining unknown as zero. With one aggregate-eligible session having empty tokens/null cost and another having 10k tokens/$1, this renders `10k tok · ≈$1.00` as the group total even though the true total is unknown; the cost reducer in `CoslashPage.tsx` has the same problem. Preserve null whenever any included value is unknown (for example, guard with `every(value => value != null)` before reducing), or explicitly label the result as partial.

luhe19001 (commented):
**Carried over:** 4 resolved, 0 unresolved (4 prior findings).

| --- | --- | --- | --- |
| collector/internal/sessionexport/snapshot.go:397 — Preserve unknown cost | P1 | Resolved | Completed unknown-cost sessions are ineligible, and preview construction rejects nil cost before serialization. |
| frontend/src/pages/coslash/components/SessionSortDropdownMenu.tsx:42 — Keep unknown values distinct while sorting | P1 | Resolved | Nullable costs and token totals sort after known values in both directions. |
| frontend/src/pages/coslash/lib/session.ts:71 — Make session cost nullable | P1 | Resolved | Both Session.cost and Subagent.cost are declared number \| null. |
| frontend/src/pages/coslash/lib/session.ts:452 — Propagate unknown tokens to breakdown | P1 | Resolved | TokenBreakdown returns an unavailable display for an empty token map. |

## Review summary

This PR consistently introduces nullable usage through the collector and frontend. Aggregate totals still lose the unknown state when known and unknown sessions are mixed. Review used the required single-pass fallback because subagents were unavailable; dynamic verification was unavailable because Go and installed frontend dependencies are absent.

Review profile: Helu

milanshen (approved):

frontend/src/pages/coslash/components/SessionBoard.tsx
calvintvu: Fixed in 4f0e601. Aggregate token and cost totals now remain unknown unless every included session has a known value, with regression coverage for mixed known/unknown values.

PR #184, opened 2026-09-15
[Cursor Support] Add Cursor transcript parsing foundation

calvintvu:
## Context

Cursor stores local agent conversations as JSONL transcripts. This change adds the transcript-only collection foundation needed to map those conversations into the shared session model. Cursor remains unregistered until later changes add side-store enrichment and frontend support.

## Changes

- Discover, validate, group, and parse Cursor transcript files.
- Extract prompts, activity, tools, commands, todos, file edits, commit attempts, pull-request URLs, and task intent.
- Apply candidate limits while keeping path-based session families together.
- Share commit-attempt, pull-request URL, and subagent-task parsing helpers with existing collector code.

## Test

- `git diff --check` — passed


milanshen (commented):
## Review summary

This PR adds Cursor transcript discovery and parsing, with a generally coherent reuse of shared session infrastructure. However, the feature is not integrated end-to-end and several unannounced user-visible data and presentation behaviors are incorrect.

**Findings:** 10 blocking (P0), 0 should-fix (P1), 0 nice-to-have (P2) — see inline comments.

Review profile: Milan

collector/internal/vendors/vendors.go
```diff
@@ -3,5 +3,6 @@ package vendors
 const (
 	AgentClaude   = "claude"
 	AgentCodex    = "codex"
+	AgentCursor   = "cursor"
```
milanshen, line 6: **[P0] Register Cursor throughout the application**

`AgentCursor` is added, but `collector/internal/collector/collector.go` still imports and registers only Claude, Codex, and OpenCode. Therefore valid Cursor transcripts never reach list, preview, facts, or health APIs. The frontend vendor catalog also omits Cursor, which would leave it without a filter or proper badge after backend registration.

Add the Cursor source functions to `vendorSources` and add Cursor to the frontend vendor catalog.

collector/internal/vendors/cursor/parse.go
```diff
@@ -0,0 +1,389 @@
...
+		Tokens: map[string]session.ModelTokens{}, UnpricedModels: []string{}, Subagents: []session.Subagent{},
+		CommitLog: commitLog, SessionDetails: details,
+	}
+	if startedAt > 0 && modified >= startedAt {
+		duration := int(modified - startedAt)
+		result.DurationMs = &duration
+	}
+	return &vendors.ParsedSession{
+		Session: result, LogPath: path, LogModifiedAtMs: modified,
+		ParentID: ParentIDFromPath(path),
```
milanshen, line 186: **[P0] Correlate child transcripts with Task spawns**

Parent Task entries use the tool block ID as their spawn key, but child `ParsedSession` values never set `SpawnKey`. For every Cursor subagent, shared composition cannot link the child to its Task, deletes the digest row, omits `SpawnedAtTurn`, and can report an active child as aborted.

Parse the Task result's child identifier and assign a matching spawn key to both the parent entry and child session.

collector/internal/vendors/cursor/parse.go
```diff
@@ -0,0 +1,389 @@
...
+		InTurn:   inTurn, Stopped: stopped,
```
milanshen, line 187: **[P0] Expose active Cursor sessions as active**

`InTurn` is populated, but Cursor supplies neither liveness metadata nor `StatusHint`. When a root transcript is currently mid-turn, `resolveStatus` consequently leaves its status nil and the frontend renders it as inactive.

Provide a derived busy status hint while `inTurn` is true, or add authoritative Cursor liveness metadata.

collector/internal/vendors/cursor/parse.go
```diff
@@ -0,0 +1,389 @@
...
+		Agent: vendors.AgentCursor, ID: id, WorkingDirectory: cwd,
+		EditedFileCount: len(edits.Edits), StartedAt: startedAt, LastActivityTime: modified,
+		Tokens: map[string]session.ModelTokens{}, UnpricedModels: []string{}, Subagents: []session.Subagent{},
+		CommitLog: commitLog, SessionDetails: details,
+	}
+	if startedAt > 0 && modified >= startedAt {
+		duration := int(modified - startedAt)
+		result.DurationMs = &duration
+	}
+	return &vendors.ParsedSession{
```
milanshen, line 184: **[P0] Populate Cursor names and summaries**

The returned session never receives a summary and the `ParsedSession` never receives a name. Thus every Cursor root appears as “Untitled session,” subagents fall back to UUID names, and completed sessions show “No summary available” even when the transcript contains final assistant text.

Use the first prompt as the name fallback and preserve final assistant output as a recap and session summary.

collector/internal/vendors/cursor/parse.go
```diff
@@ -0,0 +1,389 @@
...
+			prompt, timestamp := unwrapUserText(text)
+			if startedAt == 0 && timestamp > 0 {
+				startedAt = timestamp
+			}
+			category := session.DigestUser
+			if firstPrompt == "" {
+				firstPrompt = prompt
+				category = session.DigestFirstPrompt
+			}
+			if category == session.DigestUser && strings.HasSuffix(strings.TrimSpace(prompt), "?") {
```
milanshen: **[P0] Preserve user questions as user turns**

Any noninitial user prompt ending in `?` is reclassified as `DigestQuestion`, with the assistant's response stored as its answer. For a normal prompt such as “Can you fix this?”, the UI loses the user-turn entry and displays the inverse of the shared question semantics, which represent an agent question followed by the user's answer.

Always emit user prompts as user turns and derive question entries only from Cursor's question-tool interaction.

collector/internal/vendors/cursor/parse.go
```diff
@@ -0,0 +1,389 @@
...
+		if record.Role == "assistant" && record.Message != nil {
+			for _, block := range record.Message.Content {
+				if block.Type == "text" {
+					if text := strings.TrimSpace(block.Text); text != "" {
+						if pendingQuestion != "" {
+							digest.PushQuestion(pendingQuestionTurn, pendingQuestion, text, pendingQuestionTime)
+							pendingQuestion = ""
+						}
+					}
+					for _, url := range session.PullRequestURLs(block.Text) {
```
milanshen: **[P0] Count only created pull requests**

Every GitHub pull-request URL in assistant prose increments `PullRequests`. If Cursor merely discusses or reviews an existing PR URL, the session UI incorrectly reports that it created a pull request.

Associate URLs only with a successfully completed PR-create command or a dedicated creation event.

collector/internal/vendors/cursor/parse.go
```diff
@@ -0,0 +1,389 @@
...
+					if input.Command != "" {
+						commands.Note(input.Command, input.Description)
+						commitLog = append(commitLog, session.ParseCommitAttempts(input.Command)...)
+					}
+					if input.WorkingDirectory != "" {
+						cwd = input.WorkingDirectory
+					}
+				case "Write":
+					if input.Path != "" {
+						edits.Add(input.Path, session.CountLines(input.Contents), 0, false)
```
milanshen, line 101: **[P0] Do not count failed tool requests as edits**

File edits are recorded immediately from `tool_use`, without examining the corresponding result. For example, a Write denied by filesystem permissions or a StrReplace whose old text is absent still increases the edited-file and line counts even though no file changed.

Correlate tool uses with their results and record edits only after successful completion.

collector/internal/vendors/cursor/source.go
```diff
@@ -0,0 +1,192 @@
...
+func Health() vendors.SourceHealth {
+	root, err := Root()
+	if err != nil {
+		return vendors.SourceHealth{Agent: vendors.AgentCursor, Err: err}
+	}
+	scan, err := Scan()
+	if err != nil {
+		return vendors.SourceHealth{Agent: vendors.AgentCursor, Root: root, Err: err}
+	}
+	return vendors.FileSourceHealth(vendors.AgentCursor, root, scan, func(string) (bool, error) { return true, nil })
```
milanshen: **[P0] Exclude subagents from the health session count**

The always-true root predicate makes `FileSourceHealth` count every `subagents/*.jsonl` transcript as an independent session. A root with three children is therefore reported as four sessions in diagnostics.

Use a predicate such as `ParentIDFromPath(path) == ""`.

collector/internal/vendors/cursor/source.go
```diff
@@ -0,0 +1,192 @@
...
+
+func GetSessionFamily(id string) ([]*vendors.ParsedSession, *vendors.SessionMetadata, error) {
+	if id == "" {
+		return nil, vendors.EmptySessionMetadata(), nil
+	}
+	files, err := Files()
+	if err != nil {
+		return nil, vendors.EmptySessionMetadata(), err
+	}
+	parsed := parseTranscriptFilesSource(vendors.LocalReadSource, files)
```
milanshen: **[P0] Load only the requested session family**

Opening one Cursor preview parses and JSON-decodes every transcript in the user's history before discarding unrelated families. On installations with many or large transcripts, preview latency and memory use scale with the entire history.

Resolve the requested root, ancestors, and descendants from paths first, then parse only those files.

collector/internal/vendors/cursor/source.go
```diff
@@ -0,0 +1,192 @@
...
+			eligibleFamilies[familyID] = true
+		}
+	}
+	eligible := make([]string, 0, len(files))
+	for _, path := range files {
+		if eligibleFamilies[union.find(IDFromPath(path))] {
+			eligible = append(eligible, path)
+		}
+	}
+	selected, _ := vendors.LimitNewestSourceFileFamilies(source, eligible, vendors.MaxCandidateFilesPerAgent,
```
milanshen, line 127: **[P0] Do not silently truncate local history**

The 2,000-file family cap is applied even when `Collect(0)` requests all local history. Users exceeding that threshold silently lose older Cursor families from the all-history view, unlike the existing local collectors.

Avoid this cap for local all-history collection or expose an explicit truncation state.

collector/internal/vendors/vendors.go
calvintvu: Push back. This PR is intentionally transcript-only and explicitly leaves Cursor unregistered until frontend support. PR #187 registers Cursor after side-store enrichment and family reconstruction land. Registering it here would expose an incomplete source. No change.

collector/internal/vendors/cursor/parse.go
calvintvu: Push back. Cursor JSONL records Task intent but not the child and tool-call relationship. I verified local transcript tool blocks contain only tool_use fields. PR #186 reconstructs ParentID and SpawnKey from the IDE and CLI side stores. Guessing that correlation here would be incorrect. No change.

collector/internal/vendors/cursor/parse.go
calvintvu: Push back. This PR intentionally does not activate Cursor. PR #189 adds authoritative IDE and CLI liveness. A StatusHint derived only from the last transcript record would conflate stale or interrupted sessions with a currently live process. No change.

collector/internal/vendors/cursor/parse.go
calvintvu: Push back. PR #185 loads names and summaries from Cursor side stores, and PR #197 adds conversation recap and compaction-summary support. This parser already retains FirstPrompt, which shared name resolution uses after registration. Treating arbitrary final assistant text as the root summary here would be unreliable. No change.

collector/internal/vendors/cursor/parse.go
calvintvu: Agree. The punctuation heuristic inverted the shared question semantics. Commit ad0062c now preserves every Cursor user prompt as a user turn. Cursor JSONL has no question-tool result records, so the parser no longer fabricates question entries. Covered by TestParseTranscriptPreservesUserQuestionsAsUserTurns.

collector/internal/vendors/cursor/parse.go
calvintvu: Partially agree. Counting any URL in assistant prose was a real false positive, but transcript-only JSONL has no command-result records to prove creation. Commit ad0062c removes that unreliable count. PR #185 supplies the reliable path by counting URLs only from completed side-store tool results. Covered by TestParseTranscriptDoesNotCountPullRequestURLsFromAssistantProse.

collector/internal/vendors/cursor/parse.go
calvintvu: Push back on the requested correlation. I verified Cursor JSONL contains tool_use blocks and turn_ended status, but no per-tool result ID or success field. There is no corresponding transcript result to match. PR #185 adds authoritative IDE checkpoint edits where available, while transcript edit attempts remain the only available signal for CLI and SDK sessions. No change.

collector/internal/vendors/cursor/source.go
calvintvu: Agree. The health predicate counted every transcript file. Commit ad0062c now counts only paths with no ParentID, so subagent files remain entries but not sessions. Covered by TestCursorSourceHealthCountsOnlyRootTranscripts.

collector/internal/vendors/cursor/source.go
calvintvu: Agree. GetSessionFamily parsed the complete history before filtering. Commit ad0062c resolves the requested session root and family membership from paths first, then parses only that family. Covered for root and child lookup by TestCursorFamilyFilesSelectsOnlyRequestedFamily.

collector/internal/vendors/cursor/source.go
calvintvu: Agree. Commit ad0062c bypasses the candidate cap when since is zero, matching the all-history behavior of the other local collectors. Windowed collection keeps the cap. Covered by TestSelectCursorFilesDoesNotCapAllHistory.

milanshen:
@agent-review

collector/internal/vendors/vendors.go
luhe19001: **[P0 — CLOSED]** This PR intentionally lands the transcript parser without activating Cursor. Registration is credibly deferred until the side-store enrichment, family reconstruction, and frontend support are present, avoiding exposure of an incomplete source.

collector/internal/vendors/cursor/parse.go
luhe19001: **[P0 — CLOSED]** Cursor transcript JSONL provides the Task intent but not the child/tool-call relationship. Guessing a spawn key here would be incorrect; deferring correlation to the authoritative IDE and CLI side stores is acceptable.

collector/internal/vendors/cursor/parse.go
luhe19001: **[P0 — CLOSED]** A trailing in-progress transcript record is not authoritative liveness and may be stale after interruption. Deferring status until Cursor process and side-store liveness are available avoids incorrectly presenting stale sessions as active.

collector/internal/vendors/cursor/parse.go
luhe19001: **[P0 — CLOSED]** The parser retains `FirstPrompt`, which shared resolution can use as the root-name fallback. Authoritative names and summaries are reasonably deferred to Cursor side stores; arbitrary final assistant prose is not a reliable session summary.

collector/internal/vendors/cursor/parse.go
luhe19001: **[P0 — RESOLVED]** Fixed in `ad0062c0d46bce32d0d239b897533ada8f724b0a`: every nonempty Cursor user prompt is now emitted as a first-prompt or user digest entry, and the punctuation-based question conversion is gone.

collector/internal/vendors/cursor/parse.go
luhe19001: **[P0 — RESOLVED]** Fixed in `ad0062c0d46bce32d0d239b897533ada8f724b0a`: assistant prose is no longer scanned for pull-request URLs, so merely discussing an existing PR does not increment the count.

collector/internal/vendors/cursor/parse.go
luhe19001: **[P0 — CLOSED]** Cursor transcript JSONL provides tool requests but no correlatable per-tool outcome. Retaining edit attempts as the available CLI/SDK signal while authoritative IDE checkpoint edits land through the dedicated enrichment follow-up is an acceptable boundary for this parser-only PR.

collector/internal/vendors/cursor/source.go
luhe19001: **[P0 — RESOLVED]** Fixed in `ad0062c0d46bce32d0d239b897533ada8f724b0a`: Cursor health now counts only transcript paths whose `ParentIDFromPath` is empty, excluding subagent files.

collector/internal/vendors/cursor/source.go
luhe19001: **[P0 — RESOLVED]** Fixed in `ad0062c0d46bce32d0d239b897533ada8f724b0a`: `GetSessionFamily` now resolves family membership from paths before parsing and decodes only the selected files.

collector/internal/vendors/cursor/source.go
luhe19001: **[P0 — RESOLVED]** Fixed in `ad0062c0d46bce32d0d239b897533ada8f724b0a`: all-history collection now returns every eligible file without applying the 2,000-file cap.

collector/internal/vendors/cursor/source.go
```diff
@@ -0,0 +1,223 @@
...
+	for _, item := range parsed {
+		if item != nil && item.Session != nil {
+			byID[item.Session.ID] = item
+		}
+	}
+	root, ok := byID[id]
+	if !ok {
+		return nil
+	}
+	for root.ParentID != "" {
```
luhe19001, line 176: **[P1] Guard family traversal against parent cycles**

An accepted root and subagent transcript can share an ID. If the child fragment is newer, parsing assigns that session its own ID as `ParentID`; this loop then repeatedly resolves the same map entry and `GetSessionFamily` never returns.

Reject self-parent relationships and track visited IDs while walking ancestors so malformed or colliding transcript paths cannot hang requests.

collector/internal/vendors/cursor/discovery.go
```diff
@@ -0,0 +1,107 @@
...
+		return false
+	}
+	filenameID := strings.TrimSuffix(filepath.Base(path), filepath.Ext(path))
+	if !strings.EqualFold(filenameID, id) {
+		return false
+	}
+
+	directory := filepath.Dir(path)
+	if filepath.Base(filepath.Dir(directory)) == "agent-transcripts" {
+		return strings.EqualFold(filepath.Base(directory), id) &&
```
luhe19001, line 84: **[P1] Canonicalize accepted Cursor IDs**

Validation accepts UUID directory and filename casing differences with `EqualFold`, but `IDFromPath`, `ParentIDFromPath`, grouping, and family maps compare the preserved strings case-sensitively. An uppercase parent directory with a lowercase root filename is therefore accepted, yet its children are excluded from the family or dropped during composition.

Normalize IDs consistently, such as to lowercase, in both path extraction functions before using them as keys.

collector/internal/vendors/cursor/parse.go
```diff
@@ -0,0 +1,363 @@
...
+				if block.Type != "tool_use" {
+					continue
+				}
+				toolUses++
+				input, rawString := decodeToolInput(block.Input)
+				switch block.Name {
+				case "Shell", "run_terminal_cmd":
+					if input.Command != "" {
+						commands.Note(input.Command, input.Description)
+						commitLog = append(commitLog, session.ParseCommitAttempts(input.Command)...)
```
luhe19001: **[P1] Do not report unverified commit attempts**

Every `git commit` request is appended to `CommitLog` without an outcome. A failed commit is consequently reported as created when the working directory cannot be probed, or when an older repository commit has the same subject and reconciliation matches it.

Keep the command in `Commands`, but populate created-commit observations only from an authoritative successful result or side-store record.

collector/internal/vendors/cursor/parse.go
```diff
@@ -0,0 +1,363 @@
...
+		addPath, add := strings.CutPrefix(line, "*** Add File: ")
+		if update || add {
+			flush()
+			if add {
+				path = addPath
+			}
+			current = &patchFile{path: strings.TrimSpace(path), isNew: add}
+			continue
+		}
+		if strings.HasPrefix(line, "*** ") {
```
luhe19001, line 316: **[P1] Preserve ApplyPatch move sections**

A valid `*** Move to: new.go` directive is treated as an unknown section boundary. The current file is flushed before its hunks, the destination is discarded, and all following patch lines are ignored, yielding a zero-line edit against the old path.

Parse move directives as part of the current file section and retain both the destination and subsequent hunks.

collector/internal/vendors/cursor/parse.go
```diff
@@ -0,0 +1,363 @@
...
+	}
+	flush()
+	return files
+}
+
+func patchLineCounts(patch string) (int, int) {
+	adds, dels := 0, 0
+	for _, line := range strings.Split(patch, "\n") {
+		switch {
+		case strings.HasPrefix(line, "+") && !strings.HasPrefix(line, "+++"):
```
luhe19001: **[P1] Count sign-prefixed ApplyPatch content**

For extracted ApplyPatch sections there are no `+++` or `---` unified-diff headers, but these conditions still exclude them. Adding a literal line beginning `++` or deleting one beginning `--` therefore contributes zero to the displayed addition or deletion count.

Distinguish raw unified diffs from extracted ApplyPatch bodies, and count every `+` or `-` content line in the latter.

collector/internal/vendors/cursor/source.go
```diff
@@ -0,0 +1,223 @@
...
+	if err != nil {
+		return vendors.SourceHealth{Agent: vendors.AgentCursor, Root: root, Err: err}
+	}
+	return cursorSourceHealth(root, scan)
+}
+
+func cursorSourceHealth(root string, scan *vendors.SourceScan) vendors.SourceHealth {
+	return vendors.FileSourceHealth(vendors.AgentCursor, root, scan,
+		func(path string) (bool, error) { return ParentIDFromPath(path) == "", nil })
+}
```
luhe19001, line 239: **[P2] Deduplicate root transcript fragments in health counts**

The parser intentionally groups multiple root paths with the same session ID, such as a resumed session stored under two project directories, into one session. This predicate instead counts each root path, so diagnostics can report two sessions while collection returns one.

Count unique canonical root IDs rather than root files.

luhe19001 (commented):
## Review summary

This PR adds Cursor transcript discovery and parsing through the shared session infrastructure. The latest fixes improve family selection and metrics, but six concrete correctness edge cases remain.

**Findings:** 0 blocking (P0), 5 should-fix (P1), 1 nice-to-have (P2) — see inline comments.

**Prior findings:** 5 resolved, 5 closed, 0 open.
**This review:** 5 resolved, 1 closed, 0 reopened, 6 new.

Review profile: Helu

Requested in [PR comment](https://github.com/centauri-ai/coslash/pull/184#issuecomment-5724077368)

collector/internal/vendors/cursor/parse.go
calvintvu: Agreed. The generic section-boundary branch flushed the file at Move to and discarded its hunks. a728648 recognizes the move directive, updates the destination path, preserves the following patch body, and adds an end-to-end parser regression.

collector/internal/vendors/cursor/source.go
calvintvu: Agreed. The ancestor walk had no cycle guard, and a grouped root/child ID collision could leave ParentID equal to the session ID. a728648 rejects self/cyclic ancestry and adds a regression proving the lookup terminates.

collector/internal/vendors/cursor/parse.go
calvintvu: Agreed. A tool-use record proves only that the command was requested, not that it succeeded. a728648 keeps the command in Commands but removes transcript-derived commit observations; authoritative side-store evidence remains responsible for created commits.

collector/internal/vendors/cursor/discovery.go
calvintvu: Agreed. Validation was case-insensitive while map keys retained path casing. a728648 canonicalizes both IDFromPath and ParentIDFromPath to lowercase and covers mixed-case accepted paths.

collector/internal/vendors/cursor/source.go
calvintvu: Agreed. Health was counting root files while collection groups canonical session IDs. a728648 deduplicates canonical root IDs and adds a duplicate-fragment regression.

collector/internal/vendors/cursor/parse.go
calvintvu: Agreed. Extracted ApplyPatch bodies are now distinguished from raw unified diffs, so every leading + or - content line is counted while raw +++/--- headers remain excluded. Covered by a parser regression in a728648.

calvintvu:
@agent-review

collector/internal/vendors/cursor/source.go
luhe19001: **[P1 — RESOLVED]** Fixed in `a728648b86df495666fa141443edb5eb5aac6b76`: ancestor traversal now tracks visited IDs and rejects self-parent links, preventing cycles from hanging family lookup.

collector/internal/vendors/cursor/discovery.go
luhe19001: **[P1 — RESOLVED]** Fixed in `a728648b86df495666fa141443edb5eb5aac6b76`: both path extraction functions now lowercase accepted IDs before grouping and family-map lookups.

collector/internal/vendors/cursor/parse.go
luhe19001: **[P1 — RESOLVED]** Fixed in `a728648b86df495666fa141443edb5eb5aac6b76`: shell commands remain in `Commands`, but transcript tool requests no longer populate `CommitLog`.

collector/internal/vendors/cursor/parse.go
luhe19001: **[P1 — RESOLVED]** Fixed in `a728648b86df495666fa141443edb5eb5aac6b76`: move directives now update the destination path without flushing the section, preserving subsequent hunks.

collector/internal/vendors/cursor/parse.go
luhe19001: **[P1 — RESOLVED]** Fixed in `a728648b86df495666fa141443edb5eb5aac6b76`: native ApplyPatch bodies now count every sign-prefixed content line while unified-diff headers remain excluded.

collector/internal/vendors/cursor/source.go
luhe19001: **[P2 — RESOLVED]** Fixed in `a728648b86df495666fa141443edb5eb5aac6b76`: health counting now excludes subagents and deduplicates canonical root session IDs.

luhe19001 (commented):
## Review summary

This PR adds Cursor transcript discovery and parsing with shared parsing helpers. The remaining edge cases are fixed, and no new actionable defects were confirmed. Retained closed blocking-priority dispositions still determine the score.

**Prior findings:** 11 resolved, 5 closed, 0 open.
**This review:** 6 resolved, 0 closed, 0 reopened, 0 new.

Review profile: Helu

Requested in [PR comment](https://github.com/centauri-ai/coslash/pull/184#issuecomment-5732675993)

milanshen:
## Review summary

This PR adds Cursor transcript discovery and parsing with shared parsing helpers. The remaining edge cases have been fixed, all review threads are resolved, and no new actionable defects were confirmed at the current head.

**Findings:** 0 blocking (P0), 0 should-fix (P1), 0 nice-to-have (P2).

**Prior findings:** 11 resolved, 5 closed, 0 open.  
**This review:** 6 resolved, 0 closed, 0 reopened, 0 new.

milanshen (approved):

PR #185, opened 2026-09-15
[Cursor Support] Enrich Cursor sessions from local side stores

calvintvu:
## Context

Cursor transcripts omit important session fields such as names, models, working directories, and usage data. Cursor stores these fields across separate IDE, CLI, and SDK databases. This change joins those side stores with transcript sessions while preserving unknown values when Cursor does not record them.

## Changes

- Load Cursor IDE, CLI, and SDK session metadata from their local SQLite stores.
- Enrich sessions with entrypoint, name, summary, working directory, model, timestamps, context usage, recorded cost, SDK tokens, and IDE change metadata.
- Normalize Cursor model labels and add pricing data for supported Composer models.
- Keep CLI token and cost values unknown when Cursor does not store them.

## Test

- `jq empty collector/internal/session/models.json` — passed
- `git diff --check` — passed


luhe19001 (commented):
## Review summary

This PR adds Cursor side-store metadata extraction and Composer pricing using a sensible read-only SQLite approach. Enrichment ownership remains inconsistent across collection and single-session paths and should be unified before exposure.

**Findings:** 0 blocking (P0), 2 should-fix (P1), 1 nice-to-have (P2) — see inline comments.

Review profile: Helu

collector/internal/vendors/cursor/source.go
```diff
@@ -49,7 +60,73 @@ func GetSessionFamily(id string) ([]*vendors.ParsedSession, *vendors.SessionMeta
...
+	for _, item := range parsed {
+		if item == nil || item.Session == nil {
+			continue
+		}
+		enrichment := metadata.Session(item.Session.ID)
+		applyMetadataTimes(item.Session, enrichment.StartedAt, enrichment.LastActivityAt)
+		if enrichment.WorkingDirectory != "" {
+			item.Session.WorkingDirectory = enrichment.WorkingDirectory
+		}
+		mergeIDEFileEdits(item.Session, enrichment.FileEdits)
```
luhe19001, line 93: **[P1] Preserve merged edits through finalization**

This merges transcript and checkpoint edits, but `Collect` and `GetSessionFamily` return the original metadata. Shared finalization subsequently calls `vendors.ApplySessionEnrichment`, which replaces `Session.FileEdits` with the checkpoint-only `enrichment.FileEdits`. For example, if the transcript records `a.go` and the checkpoint records `b.go`, list and preview output ultimately contain only `b.go`, also losing transcript change snippets.

Make one enrichment layer own the merge, or write the merged slice back to metadata so shared finalization cannot discard it.

collector/internal/vendors/cursor/source.go
```diff
@@ -37,7 +43,12 @@ func GetSessionFacts(id string) (*vendors.ParsedSession, error) {
 	if len(fragments) == 0 {
 		return nil, nil
 	}
-	return parseTranscriptFragmentsSource(vendors.LocalReadSource, fragments)
+	parsed, err := parseTranscriptFragmentsSource(vendors.LocalReadSource, fragments)
+	if err != nil || parsed == nil {
+		return parsed, err
+	}
+	applyCursorEnrichment([]*vendors.ParsedSession{parsed}, vendors.BestEffortMetadata(vendors.AgentCursor, LoadMetadata))
```
luhe19001: **[P1] Apply usage metadata in the facts path**

`GetSessionFacts` loads metadata but only applies times, working directory, edits, and commits before discarding it. Unlike list/family finalization, this leaves model, tokens, context usage/window, recorded cost, entrypoint, and pull-request data unset. A Cursor IDE session with five turns and `contextTokensUsed` above 100,000 is consequently eligible for synthesis in list data but rejected when `/api/synthesis` reloads it through this path.

Run the shared metadata enrichment and required cost derivation before returning the session.

collector/internal/vendors/cursor/source.go
```diff
@@ -49,7 +60,73 @@ func GetSessionFamily(id string) ([]*vendors.ParsedSession, *vendors.SessionMeta
...
+		mergeIDEFileEdits(item.Session, enrichment.FileEdits)
+		if len(enrichment.CommitObservations) > 0 {
+			item.Session.CommitLog = append(item.Session.CommitLog, enrichment.CommitObservations...)
+		}
+	}
+}
+
+func applyMetadataTimes(value *session.Session, startedAt, lastActivityAt int64) {
+	if startedAt > 0 {
+		value.StartedAt = startedAt
```
luhe19001, line 109: **[P2] Recompute duration after replacing timestamps**

Cursor parsing computes `DurationMs` from the transcript start and log modification time. Replacing `StartedAt` and `LastActivityTime` here leaves that old duration intact: parsed times 100–500 produce duration 400, while metadata times 200–300 yield a final session reporting duration 400 for a 100-unit interval.

Recompute or clear `DurationMs` whenever metadata replaces the timestamps used to derive it.

collector/internal/vendors/cursor/source.go
calvintvu: Agree. The custom merge was overwritten by shared finalization. Commit c1906e6 writes the merged transcript and checkpoint slice back to the session metadata, so the later shared application preserves both paths and transcript snippets. Covered by TestCursorEnrichmentPreservesMergedEditsDuringSharedFinalization.

collector/internal/vendors/cursor/source.go
calvintvu: Agree. The facts path used only the Cursor-specific subset of metadata. Commit c1906e6 now applies the shared enrichment at the Cursor boundary, fills a model context window when needed, and attaches recorded or estimated cost before returning. Covered by TestCursorEnrichmentAppliesSharedFactsMetadataAndCost.

collector/internal/vendors/cursor/source.go
calvintvu: Agree. Commit c1906e6 now clears and recomputes DurationMs whenever metadata replaces either timestamp, using the final StartedAt and LastActivityTime values and leaving duration unknown for an invalid interval. Covered by TestCursorEnrichmentRecomputesDurationFromMetadataTimes.

milanshen:
@agent-review

luhe19001:

collector/internal/vendors/cursor/source.go
luhe19001: **[P1 — RESOLVED]** Fixed in `c1906e6`: the merged session file edits are copied back into the enrichment before shared reapplication, preserving transcript paths and snippets.

collector/internal/vendors/cursor/source.go
luhe19001: **[P1 — RESOLVED]** Fixed in `c1906e6`: the facts path now applies shared enrichment, fills the model context window when available, and attaches recorded or estimated cost.

collector/internal/vendors/cursor/source.go
luhe19001: **[P2 — RESOLVED]** Fixed in `c1906e6`: timestamp replacement now clears and recomputes `DurationMs` from the final start and activity times.

collector/internal/vendors/cursor/source.go
```diff
@@ -16,8 +19,11 @@ func Collect(since int64) ([]*vendors.ParsedSession, *vendors.SessionMetadata, e
 	if err != nil {
 		return nil, nil, err
 	}
+	metadata := vendors.BestEffortMetadata(vendors.AgentCursor, LoadMetadata)
```
luhe19001, line 22: **[P1] Keep side-store-recent sessions eligible**

Metadata is loaded here, but the following selector considers only transcript file mtimes. If `composerHeaders.lastUpdatedAt` is inside the requested window while the JSONL is older, the family is discarded before enrichment and the list omits a session whose final activity time would be recent.

Pass metadata into family selection and consider its start/activity timestamps for eligibility and ordering.

collector/internal/vendors/cursor/metadata.go
```diff
@@ -0,0 +1,590 @@
...
+		Before checkpoint `json:"gitCheckpoint"`
+		After  checkpoint `json:"afterGitCheckpoint"`
+	}
+	if json.Unmarshal([]byte(value), &bubble) != nil {
+		return nil
+	}
+	observations := []session.CommitObservation{}
+	for workspace, after := range bubble.After.CommitHashesByGitWorkspace {
+		before, ok := bubble.Before.CommitHashesByGitWorkspace[workspace]
+		if ok && before.CommitHash != after.CommitHash && fullCommitHash.MatchString(after.CommitHash) {
```
luhe19001: **[P2] Distinguish commits from other HEAD changes**

Any before/after HEAD change is recorded as a created commit. A bubble that runs `git checkout`, `reset`, or a fast-forwarding `pull` can therefore make the destination commit appear among commits created during the session once it resolves against repository history.

Require evidence of a commit operation or verify an appropriate commit-creation transition before emitting the observation.

collector/internal/vendors/cursor/source.go
luhe19001: **[P2] Scope exact-session metadata reads**

Loading facts for one session invokes unrestricted `LoadMetadata`, which opens every CLI and SDK store and scans the entire IDE side store. Thus synthesis and exact-session requests scale with the user's complete Cursor history instead of the requested ID; `GetSessionFamily` has the same behavior.

Add an ID-scoped metadata loader or reuse already-loaded metadata for exact requests.

luhe19001 (commented):
## Review summary

This PR adds read-only Cursor side-store enrichment and Composer pricing, and the prior enrichment regressions are corrected. The approach is generally sound, but incremental collection can still discard sessions before authoritative side-store activity is applied.

**Findings:** 0 blocking (P0), 1 should-fix (P1), 2 nice-to-have (P2) — see inline comments.

**Prior findings:** 3 resolved, 0 closed, 0 open.
**This review:** 3 resolved, 0 closed, 0 reopened, 3 new.

Review profile: Helu

Requested in [PR comment](https://github.com/centauri-ai/coslash/pull/185#issuecomment-5724574191)

collector/internal/vendors/cursor/source.go
calvintvu: Agreed. 8951b71 adds an exact-ID metadata path used by GetSessionFacts and GetSessionFamily. It constrains IDE and auxiliary SQL queries to the requested IDs, opens only matching CLI stores, and restricts SDK reads to stores beside the requested transcript paths. A real SQLite regression verifies unrelated IDs are not loaded.

collector/internal/vendors/cursor/source.go
calvintvu: Agreed. Metadata was loaded before selection but ignored by both the since filter and capped-family ordering. 8951b71 uses the maximum transcript or side-store activity time for both decisions, with regressions for eligibility and ordering.

collector/internal/vendors/cursor/metadata.go
calvintvu: Agreed. A HEAD delta alone can come from checkout, reset, pull, or edit checkpoints. 8951b71 now requires a completed Cursor terminal tool whose command contains an actual git commit invocation before emitting the changed hash, with positive and negative regressions.

calvintvu:
@agent-review

luhe19001:

collector/internal/vendors/cursor/source.go
luhe19001: **[P1 — RESOLVED]** Fixed in `8951b71b7312d7b00c97e58b8f5ef42d7d2fe7ac`: family eligibility and capped ordering now include side-store start and activity timestamps.

collector/internal/vendors/cursor/metadata.go
luhe19001: **[P2 — RESOLVED]** Fixed in `8951b71b7312d7b00c97e58b8f5ef42d7d2fe7ac`: a changed HEAD is emitted only for a completed terminal command containing a parsed `git commit` invocation.

collector/internal/vendors/cursor/source.go
luhe19001: **[P2 — RESOLVED]** Fixed in `8951b71b7312d7b00c97e58b8f5ef42d7d2fe7ac`: exact facts and family requests now constrain SQL queries and auxiliary-store discovery to the requested session IDs and transcript paths.

collector/internal/vendors/cursor/metadata.go
```diff
@@ -0,0 +1,753 @@
...
+		return nil
+	}
+	var rawArgs string
+	if json.Unmarshal(bubble.Tool.RawArgs, &rawArgs) != nil {
+		rawArgs = string(bubble.Tool.RawArgs)
+	}
+	var args struct {
+		Command string `json:"command"`
+	}
+	if json.Unmarshal([]byte(rawArgs), &args) != nil || len(session.ParseCommitAttempts(args.Command)) == 0 {
```
luhe19001: **[P2] Tie checkpoint hashes to the commit command**

This only verifies that the terminal command contains some `git commit`, then treats the final checkpoint HEAD as that commit. For `git commit -m temp && git checkout other`, the pre-existing destination commit is consequently reported as session-created. Verify the commit hash from command output or reject subsequent HEAD-changing operations.

collector/internal/vendors/cursor/metadata.go
```diff
@@ -0,0 +1,753 @@
...
+			ContextTokensUsed *int `json:"contextTokensUsed"`
+			ContextTokenLimit *int `json:"contextTokenLimit"`
+		}
+		if json.Unmarshal([]byte(value), &item) != nil {
+			continue
+		}
+		if parts := strings.SplitN(key, ":", 3); len(parts) == 3 && parts[0] == "bubbleId" && transcriptIDPattern.MatchString(parts[1]) {
+			id := parts[1]
+			model := strings.TrimSpace(item.ModelInfo.ModelName)
+			createdAt, _ := item.CreatedAt.(string)
```
luhe19001: **[P2] Compare numeric bubble timestamps correctly**

A numeric JSON `createdAt` decodes as `float64`, so this string assertion produces `""`. With multiple model-bearing bubbles, every timestamp then ties and unordered SQL iteration determines the reported model. Normalize string and numeric timestamps to one comparable representation with a deterministic tie-breaker.

collector/internal/vendors/cursor/metadata.go
```diff
@@ -0,0 +1,753 @@
...
+	if ids == nil {
+		return query, nil
+	}
+	if len(ids) == 0 {
+		return query + " WHERE 0", nil
+	}
+	placeholders := strings.TrimSuffix(strings.Repeat("?,", len(ids)), ",")
+	args := make([]any, len(ids))
+	for i, id := range ids {
+		args[i] = id
```
luhe19001, line 178: **[P2] Canonicalize metadata IDs consistently**

Requested IDs are lowercased before this case-sensitive query, while IDs decoded from side stores remain verbatim. A valid uppercase UUID can therefore miss exact queries or be stored under an uppercase map key while transcript enrichment looks up the lowercase ID, silently dropping metadata. Canonicalize decoded IDs and use compatible case-insensitive matching.

luhe19001 (commented):
## Review summary

This PR adds Cursor side-store enrichment and Composer pricing. The prior regressions are fixed; remaining findings are minor metadata edge cases, chiefly compound-command commit attribution.

**Findings:** 0 blocking (P0), 0 should-fix (P1), 3 nice-to-have (P2) — see inline comments.

**Prior findings:** 6 resolved, 0 closed, 0 deferred, 0 open.
**This review:** 3 resolved, 0 closed, 0 deferred, 0 reopened, 3 new.

Review profile: Helu

Requested in [PR comment](https://github.com/centauri-ai/coslash/pull/185#issuecomment-5733103949)

collector/internal/vendors/cursor/metadata.go
calvintvu: Agreed. Commit 015af52 now parses the completed command output and accepts a changed checkpoint HEAD only when it matches the commit hash emitted by git. The regression covers `git commit ... && git checkout other` and verifies that the destination HEAD is not reported as session-created.

collector/internal/vendors/cursor/metadata.go
calvintvu: Agreed. Commit 015af52 normalizes numeric, numeric-string, and Cursor timestamp values before comparing bubbles, then uses the bubble key as a deterministic tie-breaker. The regression includes out-of-order numeric timestamps and equal timestamps.

collector/internal/vendors/cursor/metadata.go
calvintvu: Agreed. Commit 015af52 canonicalizes Cursor IDs at query and decode boundaries, makes scoped SQLite ID and key predicates case-insensitive, and keeps metadata map keys lowercase. The regression loads uppercase side-store IDs through a lowercase exact request.

calvintvu: Follow-up: commit 8c24a05 also canonicalizes the IDE header ID before its working-directory side effect, and the regression now covers that field so all decoded header metadata stays on the lowercase key.

luhe19001 (approved):

collector/internal/vendors/cursor/metadata.go
calvintvu: **[P2 - RESOLVED]** Fixed in `015af52e493d3e896e58fa2e01f53ff6df99988b`: numeric, numeric-string, and Cursor timestamps are normalized before comparison, with a deterministic key tie-breaker.

collector/internal/vendors/cursor/metadata.go
calvintvu: **[P2 - RESOLVED]** Fixed in `8c24a0548350021604c03e2977588400fb82a5b0`: IDs are canonicalized at query and decode boundaries, including the IDE header path, and exact lookups are case-insensitive.

collector/internal/vendors/cursor/metadata.go
calvintvu: **[P2 - RESOLVED]** Fixed in `015af52e493d3e896e58fa2e01f53ff6df99988b`: checkpoint HEAD changes now count only when completed command output identifies the same commit, so a later checkout cannot be misattributed.

collector/internal/vendors/cursor/metadata.go
```diff
@@ -0,0 +1,798 @@
...
+			createdAt := cursorBubbleTime(item.CreatedAt)
+			previous := bubbles[id]
+			if model != "" && (previous.model == "" || createdAt > previous.time || createdAt == previous.time && key > previous.key) {
+				bubbles[id] = observed{model: model, time: createdAt, key: key}
+			}
+			if item.ToolFormerData.Status == "completed" {
+				if pullRequests[id] == nil {
+					pullRequests[id] = map[string]struct{}{}
+				}
+				for _, url := range session.PullRequestURLs(item.ToolFormerData.Result) {
```
calvintvu, line 567: **[P1] Verify that the tool actually created the PR**

Any completed Cursor IDE tool result containing a GitHub PR URL increments `PullRequests`. For example, `gh pr view --json url` or a file-read result containing a PR link reports a created PR even though no creation occurred, contrary to the requirement not to present heuristic PR data as authoritative.

Decode the tool command fields and require a successful PR-creation command, as the existing Codex path does with `session.IsPullRequestCreate`, before counting result URLs.

collector/internal/vendors/cursor/metadata.go
```diff
@@ -0,0 +1,798 @@
...
+			LatestCheckpointID string `json:"latestCheckpointId"`
+		}
+		if ok && transcriptIDPattern.MatchString(id) && json.Unmarshal([]byte(value), &item) == nil && item.LatestCheckpointID != "" {
+			checkpoints[id] = item.LatestCheckpointID
+		}
+	}
+	rows.Close()
+	for id, checkpointID := range checkpoints {
+		var value string
+		if db.QueryRow(`SELECT value FROM cursorDiskKV WHERE LOWER(key) = ?`, strings.ToLower("checkpointId:"+id+":"+checkpointID)).Scan(&value) != nil {
```
calvintvu, line 398: **[P2] Batch checkpoint lookups instead of scanning the table**

This executes one `LOWER(key) = ?` query per selected checkpoint. The expression prevents the normal `key` index from serving the lookup, so 100 selected IDE sessions against a 100,000-row `cursorDiskKV` table perform 100 full scans on every refresh.

Fetch the requested checkpoint keys in a bounded batch, or otherwise use an indexed lookup without wrapping the indexed column.

calvintvu (commented):
## Review summary

**Score: 4/5** - not safe to merge

The side-store enrichment is generally well integrated and the nine prior findings are resolved. PR attribution still accepts unrelated completed tool output, and checkpoint enrichment repeats an unindexed full-table scan per selected session.

**Findings:** 0 blocking (P0), 1 should-fix (P1), 1 nice-to-have (P2) - see inline comments.

**Prior findings:** 9 resolved, 0 closed, 0 deferred, 0 open.
**This review:** 3 resolved, 0 closed, 0 deferred, 0 reopened, 2 new.

collector/internal/vendors/cursor/metadata.go
calvintvu: **[P1 - RESOLVED]** Fixed in `d0ad420aeff9e05806cb78cb5e053d56ba2016b5`: completed Cursor IDE results now count only when the decoded terminal command passes `session.IsPullRequestCreate`. The regression verifies that `gh pr view` is ignored while `gh pr create` is counted.

PR #187, opened 2026-09-15
[Cursor Support] Show Cursor sessions in coSlash

calvintvu:
## Context

The collector can now parse and enrich local Cursor sessions, but the sessions are not registered or visible in coSlash. This change activates Cursor collection and adds the frontend presentation needed to browse Cursor sessions. Launch and live-status support remain separate follow-up changes.

## Changes

- Register Cursor collection, family lookup, session facts, and source health.
- Add Cursor filtering, session counts, labels, and black-and-gray styling.
- Display Cursor CLI, IDE, and SDK modalities.
- Show Cursor subagent parent labels and full working directories for ambiguous local repositories.
- Keep Cursor sessions view-only until launch support is available.

## Test

- `npm test` — 94 tests passed
- `npm run build` — passed
- `npm run lint` — passed with pre-existing warnings
- `git diff --check` — passed

### Screenshots

- [x] Session board — Cursor filter, session cards, colors, and modality badges
Screenshot 2026-09-15 at 10 38 46

- [x] Session inspector — Cursor details, subagents, and view-only state
Screenshot 2026-09-15 at 10 39 00


collector/internal/collector/collector.go
```diff
@@ -47,6 +48,11 @@ var vendorSources = []vendorSource{
 		loadFamily: opencode.GetSessionFamily,
 		health:     opencode.Health,
 	},
+	{
+		name: vendors.AgentCursor, collect: cursor.Collect, loadFacts: cursor.GetSessionFacts,
```
luhe19001, line 53: **[P1] Detect Cursor liveness before exposing sessions**

The Cursor collector never populates `SessionEnrichment.Live`, and parsed roots have no `StatusHint`. Consequently, a currently running Cursor IDE or CLI session reaches `resolveStatus` with `status: null` and is displayed as “Inactive”; an in-progress CLI subagent can similarly be classified as aborted. Populate Cursor liveness before registration and retain live families regardless of transcript age.

frontend/src/pages/coslash/lib/session.ts
```diff
@@ -168,8 +177,8 @@ export function environmentFact(value: string | null | undefined): string {
   return trimmed ? trimmed : '—';
 }
 
-export function sessionLocationFact(session: Pick<Session, 'repo' | 'cwd'>): string {
-  return environmentFact(session.repo?.trim() || session.cwd);
+export function sessionLocationFact(session: Pick<Session, 'repo' | 'repoLocalOnly' | 'cwd'>): string {
+  return environmentFact(session.repoLocalOnly ? session.cwd : session.repo?.trim() || session.cwd);
```
luhe19001: **[P2] Preserve repository names when remote cwd is redacted**

For an SSH session in a repository without an origin, `remoteLibrarySession` preserves the safe repository basename and `repoLocalOnly: true` but deliberately removes `cwd`. This branch selects that empty `cwd`, so cards and inspector headers now show `—` instead of the preserved repository. Prefer `cwd` only when it is nonblank, otherwise fall back to `repo`.

collector/internal/collector/collector.go
```diff
@@ -47,6 +48,11 @@ var vendorSources = []vendorSource{
...
+		loadFamily: cursor.GetSessionFamily,
+		health:     cursor.Health,
```
luhe19001, line 55: **[P2] Avoid resumable-CLI diagnostics for read-only Cursor support**

Registering Cursor health sends it through the generic CLI check. When Cursor transcripts exist but `cursor` is not on `PATH`, diagnostics reports a warning instructing users to install the CLI “to resume sessions,” although this PR explicitly disables Cursor launching and resuming. Model source launch capability separately and suppress that remediation for read-only sources.

luhe19001, line 55: **[P2] Count Cursor roots rather than every transcript file**

`cursor.Health` passes an always-true root predicate to `FileSourceHealth`, while its scan includes nested subagent transcripts. A family with one root and three children is therefore reported as four sessions, and orphan-only input can avoid the intended “no root sessions” failure. Count unique Cursor IDs whose `ParentIDFromPath` is empty.

luhe19001 (commented):
## Review summary

This PR enables local Cursor session collection and adds the corresponding frontend presentation, read-only controls, and nested-subagent ancestry. The cross-layer approach is coherent, but the source integration is missing runtime-state handling, with smaller remote-location and diagnostics regressions also needing attention.

**Findings:** 0 blocking (P0), 1 should-fix (P1), 3 nice-to-have (P2) — see inline comments.

Review profile: Helu

collector/internal/collector/collector.go
calvintvu: I am not changing this in PR 187. This PR is intentionally view-only and does not introduce Cursor liveness. PR 189 is the dedicated liveness and diagnostics layer: it adds authoritative IDE and CLI probes and retains live Cursor families even when their transcript timestamps are old. Adding a second partial liveness policy here would duplicate that logic and make the stacked behavior harder to reason about.

frontend/src/pages/coslash/lib/session.ts
calvintvu: Agreed. Commit 49c2b18 now falls back to the repository label when a remote local-only working directory is redacted, and adds a focused regression test for that case.

collector/internal/collector/collector.go
calvintvu: I am not changing this in PR 187. The generic remediation is replaced in the dedicated diagnostics PR 189, which models Cursor IDE and CLI sources separately and reports source-specific states and actions. Duplicating a temporary Cursor special case in this presentation-only PR would be removed immediately by the next stack layer.

calvintvu: This is already fixed in the rebased base commit ad0062c. Cursor health now counts only root composer records, with a focused test confirming child composer files are excluded.

milanshen:
@agent-review

collector/internal/collector/collector.go
luhe19001: **[P1 — CLOSED]** The rebuttal credibly establishes liveness as an intentional stacked-layer deferral: the dedicated follow-up provides authoritative IDE/CLI probes and live-family retention, avoiding a conflicting partial policy here.

frontend/src/pages/coslash/lib/session.ts
luhe19001: **[P2 — RESOLVED]** Fixed in `49c2b18a46881eb9f5cc223c2957dafab4961a2e`: `sessionLocationFact` now falls back to the preserved repository label when a remote local-only working directory is blank.

collector/internal/collector/collector.go
luhe19001: **[P2 — CLOSED]** The source-specific diagnostics layer is intentionally deferred to the dedicated stacked follow-up, which separates Cursor IDE and CLI detection and replaces the generic resume remediation.

luhe19001: **[P2 — RESOLVED]** Fixed in `ad0062c0d46bce32d0d239b897533ada8f724b0a`: Cursor health now counts only transcript paths whose `ParentIDFromPath` is empty, with a focused root-versus-child regression test.

frontend/src/pages/coslash/components/SessionInspector.tsx
```diff
@@ -331,7 +335,8 @@ function CacheWindowMark({ within, label }: { within: boolean; label: string })
 // Cache TTL refreshes on every request, so warmth keys off the transcript's
 // last write: within 5 min both windows hold, within 1 hr only the 1-hr one.
 function PromptCacheCell({ lastAccessAt }: { lastAccessAt: number }) {
-  const ageMin = (Date.now() - lastAccessAt) / MINUTE;
+  const [now] = useState(Date.now);
```
luhe19001: **[P1] Keep prompt-cache age moving**

`now` is captured once for the inspector’s lifetime. If the drawer opens just before the five-minute boundary, the cache remains displayed as warm indefinitely despite subsequent polling and rerenders. Update the clock with a timer or when each cache boundary is crossed.

frontend/src/pages/coslash/features/sharing/ShareToHubDialog.tsx
```diff
@@ -155,13 +156,7 @@ export function ShareToHubDialog({
     setPairingRefreshRequired(false);
     setRetryReadyAt(0);
   }, [open]);
-
-  useEffect(() => {
-    setSelected((current) => {
```
luhe19001, line 160: **[P2] Drop selections that leave the candidate set**

Removing this reconciliation leaves unavailable candidate IDs in `selected`. If a selected session becomes running, private, or offline and later becomes eligible again while the dialog remains mounted, it silently reappears selected without a new selection action. Continue pruning stored selections when `candidates` changes.

frontend/src/pages/coslash/CoslashPage.tsx
```diff
@@ -411,12 +408,7 @@ export function CoslashPage() {
       setSettingsDialogMode((current) => current ?? 'synthesis-consent');
     }
   }, [selectedSession, settingsState.response]);
-
-  useEffect(() => {
```
luhe19001, line 415: **[P2] Clear selections for missing sessions**

Without this cleanup, changing to a time window that excludes the selected session closes the inspector but retains its key. Switching back to a window containing that session then reopens the inspector without user action. Clear the key once the selected identity is absent.

frontend/src/pages/coslash/components/SnapshotPreviewDialog.tsx
```diff
@@ -134,25 +134,29 @@ export function SnapshotPreviewDialog({
...
 }: SnapshotPreviewDialogProps) {
   const [load, setLoad] = useState<LoadState>({ status: 'idle' });
+  const loadKey = `${detail.sourceId}\0${detail.agent}\0${detail.id}\0${detail.mtime}`;
 
   useEffect(() => {
-    if (!open) {
-      setLoad({ status: 'idle' });
-      return;
-    }
+    if (!open) return;
```
luhe19001, line 145: **[P2] Show loading state during preview retries**

Closing and reopening a failed preview for the same session and revision starts another request but continues rendering the matching cached error. A slow successful retry therefore still appears to have failed until it completes. Reset the state or track the active request generation as loading.

frontend/src/pages/coslash/hooks/use-diagnostics.ts
```diff
@@ -4,15 +4,13 @@ import type { Diagnostics } from '@/pages/coslash/lib/diagnostics';
...
 export function useDiagnostics(enabled: boolean) {
   const [diagnostics, setDiagnostics] = useState<Diagnostics | null>(null);
-  const [isLoading, setIsLoading] = useState(false);
   const [loadFailed, setLoadFailed] = useState(false);
   const [requestID, setRequestID] = useState(0);
+  const [completedRequestID, setCompletedRequestID] = useState(-1);
 
   useEffect(() => {
     if (!enabled) return;
     const controller = new AbortController();
```
luhe19001, line 13: **[P2] Clear diagnostics failure when retry starts**

After a diagnostics request fails, clicking “Re-run checks” starts a new request but leaves `loadFailed` true. The dialog continues showing the previous failure throughout the retry even though `isLoading` is true. Clear or key the failure state when starting each request.

frontend/src/pages/coslash/components/MachinesSettingsSection.tsx
```diff
@@ -34,19 +34,19 @@ export function MachinesSettingsSection({
...
 
   useEffect(() => onBusyChange(busy), [busy, onBusyChange]);
   useEffect(() => () => onBusyChange(false), [onBusyChange]);
 
   useEffect(() => {
-    if (!remote?.sshAlias) {
-      setMachine(null);
-      return;
-    }
+    if (!remote?.sshAlias) return;
```
luhe19001, line 50: **[P2] Discard status when the remote host changes**

The old `MachineFact` is now retained when a host is removed, and `currentMachine` only checks whether any SSH alias exists. Removing host A and adding host B without remounting can therefore display A’s offline or setup-failure state for B until the new probe returns. Clear the status or associate it with the alias that produced it.

luhe19001 (commented):
## Review summary

This PR coherently adds read-only Cursor session presentation across the collector and frontend. The dedicated follow-up makes the two Cursor deferrals acceptable, but the lint cleanup introduces a new should-fix regression and several stale-state behaviors.

**Findings:** 0 blocking (P0), 1 should-fix (P1), 5 nice-to-have (P2) — see inline comments.

**Prior findings:** 2 resolved, 2 closed, 0 open.
**This review:** 2 resolved, 2 closed, 0 reopened, 6 new.

Review profile: Helu

Requested in [PR comment](https://github.com/centauri-ai/coslash/pull/187#issuecomment-5724581185)

frontend/src/pages/coslash/components/MachinesSettingsSection.tsx
calvintvu: Agreed. The MachineFact was not associated with the alias that produced it. dcaebe5 stores the alias with each probe/setup result and displays it only for the current host; successful removal also clears it.

frontend/src/pages/coslash/CoslashPage.tsx
calvintvu: Agreed. Although the lookup intentionally ignores client-side filters, useSessions can replace the loaded library when the source time window changes. dcaebe5 restores missing-identity cleanup, so an excluded session cannot reopen when the prior window returns.

frontend/src/pages/coslash/hooks/use-diagnostics.ts
calvintvu: Agreed. dcaebe5 suppresses loadFailed whenever the current diagnostics request is still loading, so Re-run checks immediately replaces the previous failure with the loading state.

frontend/src/pages/coslash/components/SessionInspector.tsx
calvintvu: Agreed. The component captured Date.now only once, so neither cache boundary could trigger a rerender. dcaebe5 schedules refreshes exactly after the five-minute and one-hour boundaries, with a focused timing regression.

frontend/src/pages/coslash/features/sharing/ShareToHubDialog.tsx
calvintvu: Agreed. Search and window handlers pruned selections, but candidate prop changes did not. dcaebe5 reconciles selected IDs whenever the eligible candidate set changes, so a removed session cannot silently return selected.

frontend/src/pages/coslash/components/SnapshotPreviewDialog.tsx
calvintvu: Agreed. The matching cached error survived close and was rendered on the next open. dcaebe5 resets preview load state when the dialog closes, so reopening the same revision displays loading while the retry runs.

calvintvu:
@agent-review

luhe19001:

collector/internal/collector/collector.go
luhe19001: **[P1 — DEFERRED]** Cursor liveness remains absent at this head, but the maintainer’s stacked-PR rationale credibly assigns authoritative IDE/CLI probes and live-family retention to the dedicated follow-up.

collector/internal/collector/collector.go
luhe19001: **[P2 — DEFERRED]** The generic resume remediation remains inaccurate for read-only Cursor sessions, but the dedicated diagnostics follow-up will replace it with source-specific IDE and CLI states.

frontend/src/pages/coslash/components/SessionInspector.tsx
luhe19001: **[P1 — RESOLVED]** Fixed in `33d9b15e03073df1f9432815dd3254047d79dba7`: prompt-cache state now schedules rerenders immediately after both cache boundaries.

frontend/src/pages/coslash/features/sharing/ShareToHubDialog.tsx
luhe19001: **[P2 — RESOLVED]** Fixed in `33d9b15e03073df1f9432815dd3254047d79dba7`: selected IDs are pruned whenever the eligible candidate set changes.

frontend/src/pages/coslash/CoslashPage.tsx
luhe19001: **[P2 — RESOLVED]** Fixed in `33d9b15e03073df1f9432815dd3254047d79dba7`: the selected session key is cleared when its identity leaves the loaded library.

frontend/src/pages/coslash/components/SnapshotPreviewDialog.tsx
luhe19001: **[P2 — RESOLVED]** Fixed in `33d9b15e03073df1f9432815dd3254047d79dba7`: closing the preview resets its load state, so reopening shows loading during the retry.

frontend/src/pages/coslash/hooks/use-diagnostics.ts
luhe19001: **[P2 — RESOLVED]** Fixed in `33d9b15e03073df1f9432815dd3254047d79dba7`: diagnostics failures are suppressed while the newer request ID is still loading.

frontend/src/pages/coslash/components/MachinesSettingsSection.tsx
luhe19001: **[P2 — RESOLVED]** Fixed in `33d9b15e03073df1f9432815dd3254047d79dba7`: probe and setup results are keyed by SSH alias and cleared after successful removal.

collector/internal/collector/collector.go
```diff
@@ -48,6 +49,11 @@ var vendorSources = []vendorSource{
...
```
luhe19001, line 55: **[P2] Count metadata-linked Cursor children as subagents**

`cursor.Health` treats every transcript without a filesystem `subagents` parent as a root. Cursor also supports standalone child transcripts whose parent exists only in side-store metadata; collection composes those correctly, but diagnostics counts the root and child as two sessions. Apply Cursor relationship metadata when counting health roots.

collector/internal/collector/collector.go
```diff
@@ -48,6 +49,11 @@ var vendorSources = []vendorSource{
 		loadFamily: opencode.GetSessionFamily,
 		health:     opencode.Health,
 	},
+	{
+		name: vendors.AgentCursor, collect: cursor.Collect, loadFacts: cursor.GetSessionFacts,
```
luhe19001, line 53: **[P2] Avoid full-history Cursor scans on every refresh**

Activating `cursor.Collect` makes each sessions refresh call `LoadMetadata` before window selection. With no IDs supplied, that scans the global state database and every Cursor chat and SDK store, even for narrow windows; active reviews refresh every three seconds. Select relevant transcript IDs first or cache/incrementally load Cursor metadata.

frontend/src/pages/coslash/features/sharing/ShareToHubDialog.tsx
```diff
@@ -155,13 +156,13 @@ export function ShareToHubDialog({
...
   }, [open]);
+  /* oxlint-enable react/set-state-in-effect */
 
+  /* oxlint-disable react/set-state-in-effect -- revoke selections whose source is no longer eligible */
   useEffect(() => {
-    setSelected((current) => {
-      const next = reconcileVisibleSelection(current, candidates);
-      return next.size === current.size ? current : next;
-    });
+    setSelected((current) => reconcileVisibleSelection(current, candidates));
```
luhe19001: **[P2] Reconcile selections against visible candidates**

This reconciles against all eligible `candidates`, not the filtered `visible` set. If refreshed metadata makes a selected session stop matching the active search, it disappears while remaining in `selectedCandidates` and the preview batch. Reconcile whenever `visible` changes so hidden rows are removed.

frontend/src/pages/coslash/components/SessionInspector.tsx
```diff
@@ -1094,14 +1107,14 @@ export function SessionInspector({
 }) {
   const { detail, loadError } = useSessionDetail(session, sessionsVersion, synthesisSettingsKey);
   const contentRef = useRef<HTMLDivElement>(null);
-  const [selectedDiff, setSelectedDiff] = useState<FileSelection | null>(null);
+  const [selectedDiffState, setSelectedDiff] = useState<FileSelection | null>(null);
+  const selectedDiff = filePanelOpen(selectedDiffState, session) ? selectedDiffState : null;
```
luhe19001, line 1396: **[P2] Clear hidden file selections**

Deriving `selectedDiff` hides a mismatched selection without clearing `selectedDiffState`. If the parent programmatically closes the inspector by passing `session=null`, reopening the same session automatically restores its old file panel and cached result. Clear the stored selection when the session identity disappears or changes.

luhe19001 (commented):
## Review summary

This PR exposes read-only Cursor sessions and refreshes frontend transient-state handling. The approach is sound, with Cursor liveness and CLI diagnostics credibly assigned to a dedicated follow-up; remaining issues concern diagnostic accuracy, refresh cost, and stale selections.

**Findings:** 0 blocking (P0), 0 should-fix (P1), 4 nice-to-have (P2) — see inline comments.

**Prior findings:** 8 resolved, 0 closed, 2 deferred, 0 open.
**This review:** 6 resolved, 0 closed, 2 deferred, 0 reopened, 4 new.

Review profile: Helu

Requested in [PR comment](https://github.com/centauri-ai/coslash/pull/187#issuecomment-5734079005)

milanshen (approved):

collector/internal/collector/collector.go
calvintvu: Agreed. The inherited PR 186 follow-up afc3fc5 makes Cursor health relationship-aware and covers standalone metadata-linked children while preserving nested-path handling.

collector/internal/collector/collector.go
calvintvu: Agreed. Commit 6621068 replaces refresh-time full metadata loading with a lightweight IDE relationship/activity selection pass, then loads complete metadata only for selected transcript IDs and paths. This avoids opening and scanning every Cursor chat and SDK store on each refresh.

frontend/src/pages/coslash/features/sharing/ShareToHubDialog.tsx
calvintvu: Agreed. Commit 47a193b reconciles selection against the filtered `visible` candidates whenever that set changes, so hidden rows cannot remain in the preview batch.

frontend/src/pages/coslash/components/SessionInspector.tsx
calvintvu: Agreed. Commit 47a193b clears stored file selection when the controlled session disappears or its source, agent, or ID changes. The focused helper regression verifies retention only for the same identity.

collector/internal/collector/collector.go
calvintvu: **[P2 - RESOLVED]** Fixed in `afc3fc5b9c6e82bf5a91c8d35e604a0c2fe1cb65`: Cursor health counts side-store-linked standalone children as subagents.

collector/internal/collector/collector.go
calvintvu: **[P2 - RESOLVED]** Fixed in `66210687bc904afaf413b81468c70988481edfdb`: refresh selection uses lightweight relationship and activity metadata, then loads full metadata only for selected IDs and paths.

frontend/src/pages/coslash/components/SessionInspector.tsx
calvintvu: **[P2 - RESOLVED]** Fixed in `47a193bddff4df4ee4d9449b4b6bf655550700b1`: stored file selection is cleared when the controlled session identity changes or disappears.

frontend/src/pages/coslash/features/sharing/ShareToHubDialog.tsx
calvintvu: **[P2 - RESOLVED]** Fixed in `47a193bddff4df4ee4d9449b4b6bf655550700b1`: selection is reconciled against the filtered visible candidates.

collector/internal/collector/collector.go
calvintvu, line 53: **[P0] Allow Cursor through exact-detail validation**

Registering Cursor makes listed roots receive a `detailRevision`, so opening any Cursor card uses the exact-detail endpoint. `validAgent` in `collector/cmd/coslash/api_detail.go` still permits only Claude, Codex, and OpenCode, so every `agent=cursor` request returns HTTP 400 before `collector.GetSessionDetail` runs.

Add Cursor to that allowlist and cover a Cursor exact-detail request through the HTTP handler.

calvintvu (commented):
## Review summary

**Score: 2/5** - not safe to merge

Cursor is registered end to end, but the existing exact-detail allowlist rejects every Cursor inspector request. Large histories also exceed SQLite's expression-depth limit and can leave list and exact-detail revisions permanently inconsistent.

**Findings:** 1 blocking (P0), 1 should-fix (P1), 2 nice-to-have (P2) - see inline comments and prior deferred threads.

**Prior findings:** 12 resolved, 0 closed, 2 deferred, 0 open.
**This review:** 4 resolved, 0 closed, 0 deferred, 0 reopened, 2 new.

collector/internal/vendors/cursor/source.go
```diff
@@ -20,8 +20,15 @@ func Collect(since int64) ([]*vendors.ParsedSession, *vendors.SessionMetadata, e
...
-	metadata := vendors.BestEffortMetadata(vendors.AgentCursor, LoadMetadata)
-	files = selectCursorFilesSourceWithMetadata(vendors.LocalReadSource, files, since, metadata)
+	selectionMetadata := vendors.BestEffortMetadata(vendors.AgentCursor, LoadSelectionMetadata)
+	files = selectCursorFilesSourceWithMetadata(vendors.LocalReadSource, files, since, selectionMetadata)
+	ids := make([]string, 0, len(files))
+	for _, path := range files {
+		ids = append(ids, IDFromPath(path))
+	}
+	metadata := vendors.BestEffortMetadata(vendors.AgentCursor, func() (*vendors.SessionMetadata, error) {
+		return LoadMetadataForSessions(canonicalCursorIDs(ids), files)
```
calvintvu, line 30: **[P1] Keep scoped metadata queries below SQLite's expression limit**

A selected batch of 501 session IDs makes `loadIDEModelsDB` build 1,002 OR terms. The bundled SQLite dependency uses `MAX_EXPR_DEPTH=1000`, so the query fails and `BestEffortMetadata` silently returns sessions without model, context, cost, or PR metadata. Exact family lookup then succeeds with a small ID set, computes a different detail revision, and the inspector can remain stuck on `session_detail_stale`.

Batch the IDs below SQLite's expression limit or replace the OR chain with a bounded table-driven query.

collector/internal/collector/collector.go
calvintvu: Fixed in 910c3f18. validAgent now permits vendors.AgentCursor, and TestExactLocalDetailAcceptsCursorSessions exercises the real HTTP handler through a 200 response. Verified with the full collector test suite and go vet.

calvintvu: **[P0 - RESOLVED]** Fixed in `910c3f18`: Cursor is accepted by the shared exact-detail validator, and the handler regression reaches HTTP 200 with a Cursor identity. The combined stack collector suite and `go vet` pass.\n\n

collector/internal/vendors/cursor/source.go
calvintvu: **[P1 - RESOLVED]** Fixed in `1435a80dff77c6c955805e2f41b6d7fd6e8a3d7f`: scoped IDE model queries are split into batches of 400 IDs, keeping both OR terms per ID below the SQLite expression-depth limit. A 501-ID regression verifies metadata from the final batch.

PR #189, opened 2026-09-16
[Cursor Support] Detect live Cursor sessions and diagnose installations

calvintvu:
## Context

Cursor IDE and CLI store sessions separately and expose different executables. coSlash needs lane-specific liveness detection so active Cursor sessions remain visible, plus separate diagnostics for the IDE and CLI installations.

## Changes

- Detect live Cursor IDE and CLI sessions from processes holding their session databases.
- Keep live session families eligible for collection while allowing terminal transcript errors to override stale database state.
- Probe and display Cursor IDE and CLI installations separately in terminal, JSON, copied, and frontend diagnostics.

## Test

- `go test ./internal/vendors/cursor ./internal/diagnostics ./cmd/coslash` — passed
- `npm --prefix frontend run build` — passed
- `git diff --check` — passed

### Screenshots

- [x] Diagnostics dialog — Cursor source showing separate IDE and CLI availability
 
Screenshot 2026-09-15 at 18 08 46


collector/internal/vendors/cursor/source.go
```diff
@@ -260,6 +266,10 @@ func selectCursorFilesSourceWithMetadata(source vendors.ReadSource, files []stri
 			eligibleFamilies[familyID] = true
 			continue
 		}
+		if metadata != nil && metadata.Session(IDFromPath(path)).Live != "" {
```
milanshen: **[P1] Preserve live families through the second age filter**

When `since > 0`, this branch admits a family based on `Live`, but `Collect` subsequently calls `familiesSince`, which retains members solely by `LogModifiedAtMs`. A Cursor session whose store remains open but whose transcript predates the cutoff—for example, one idle longer than the 24-hour collection buffer—is parsed and then discarded, so `collector.List` never sees its live metadata and the dashboard omits it.

Make `familiesSince` preserve a family when any member is live, or consolidate the two filters into one liveness-aware filter.

collector/internal/vendors/cursor/source.go
```diff
@@ -23,6 +23,12 @@ func Collect(since int64) ([]*vendors.ParsedSession, *vendors.SessionMetadata, e
 	metadata := vendors.BestEffortMetadata(vendors.AgentCursor, LoadMetadata)
 	files = selectCursorFilesSourceWithMetadata(vendors.LocalReadSource, files, since, metadata)
 	parsed := parseTranscriptFilesSource(vendors.LocalReadSource, files)
+	for _, item := range parsed {
+		if item != nil && item.Session != nil && item.Stopped {
```
milanshen: **[P1] Apply stopped-session reconciliation to previews**

This terminal-error override is confined to `Collect`. `GetSessionFamily`, which backs `GetSessionForPreview`, parses the same records and loads the same live metadata without clearing it. If a transcript ends with `turn_ended` and `status: "error"` while Cursor still holds the SQLite store open, the board treats it as stopped but the preview's `resolveStatus` gives `Live` priority and displays it as idle or busy.

Extract this reconciliation into a helper and apply it to both `Collect` and `GetSessionFamily` before returning metadata.

collector/internal/vendors/cursor/source.go
milanshen: **[P2] Avoid creating metadata while checking liveness**

`metadata.Session(...)` inserts an empty enrichment for every transcript examined during a windowed refresh, including stale files that are rejected. A large Cursor history therefore creates a map entry per discovered transcript and makes the later metadata relationship loops traverse those empty entries on every refresh.

Use `metadata.Lookup(...)` and check the returned value for non-empty `Live` instead.

collector/internal/diagnostics/snapshot.go
```diff
@@ -259,11 +264,28 @@ func collectSource(
 			Error: displayError(userHome, skipped.Error),
 		})
 	}
-	if path, err := exec.LookPath(health.Agent); err == nil {
-		source.CLI.Found = true
-		source.CLI.Path = displayPath(userHome, path)
+	probeName := health.Agent
```
milanshen: **[P2] Remove the redundant probe-name branch**

`probeName` is always `health.Agent`: non-Cursor sources initialize `CLI.Name` to that same value, while Cursor always has a non-nil `IDE` and therefore skips the assignment. The conditional suggests a distinction that cannot occur and makes the two-installation probing harder to follow.

Call `exec.LookPath(health.Agent)` directly for the first probe.

milanshen (commented):
## Review summary

This PR adds Cursor IDE/CLI installation diagnostics and infers live Cursor sessions from open stores. The overall direction is reasonable, but liveness is handled inconsistently across filtering stages and callers, allowing live sessions to disappear and stopped previews to appear active. The timestamp-only second filter is the most important issue to fix before merge.

**Findings:** 0 blocking (P0), 2 should-fix (P1), 2 nice-to-have (P2) — see inline comments.

Review profile: Milan

collector/internal/vendors/cursor/source.go
calvintvu: Agreed. Commit db78b00 makes the second family filter liveness-aware and preserves the full family when any member has live metadata. The regression test covers an old root and child where only the child is live.

collector/internal/vendors/cursor/source.go
calvintvu: Agreed. Commit db78b00 moves terminal-stop reconciliation into the shared Cursor enrichment path, so collection, session facts, and preview-family loading all clear stale database liveness consistently. A focused test covers the stopped-session override.

collector/internal/vendors/cursor/source.go
calvintvu: Agreed. Commit db78b00 uses the non-mutating metadata lookup during window selection. The regression test verifies that rejected historical transcripts do not add empty metadata entries.

collector/internal/diagnostics/snapshot.go
calvintvu: Agreed. Commit db78b00 removes the redundant probe-name branch and probes the primary installation directly with the agent name.

calvintvu:
@agent-review

collector/internal/vendors/cursor/source.go
luhe19001: **[P1 — RESOLVED]** Fixed in `db78b00389f1cb7ead613f440a6abc36547aecff`: the original second filter became liveness-aware; the rebased head removes that filter entirely and retains the complete family when any member has live metadata.

collector/internal/vendors/cursor/source.go
luhe19001: **[P1 — RESOLVED]** Fixed in `db78b00389f1cb7ead613f440a6abc36547aecff`: terminal-stop reconciliation now runs inside shared Cursor enrichment, which is used by collection, session facts, and preview-family loading.

collector/internal/vendors/cursor/source.go
luhe19001: **[P2 — RESOLVED]** Fixed in `db78b00389f1cb7ead613f440a6abc36547aecff`: window selection uses the non-mutating metadata lookup, and rejected historical transcripts no longer create enrichment entries.

collector/internal/diagnostics/snapshot.go
luhe19001: **[P2 — RESOLVED]** Fixed in `db78b00389f1cb7ead613f440a6abc36547aecff`: the redundant probe-name conditional is gone and the primary installation is probed directly with `health.Agent`.

collector/internal/vendors/cursor/source.go
```diff
@@ -320,6 +324,10 @@ func selectCursorFilesSourceWithMetadata(source vendors.ReadSource, files []stri
 			eligibleFamilies[familyID] = true
 			continue
 		}
+		if enrichment := metadata.Lookup(IDFromPath(path)); enrichment != nil && enrichment.Live != "" {
```
luhe19001, line 327: **[P1] Prioritize live families before capping candidates**

This branch only makes a live family eligible. If more than 2,000 eligible files have newer timestamps than a currently open but long-idle family, `LimitNewestFileFamilies` subsequently ranks that live family by its old timestamps and discards it, so the collector never parses or displays the active session.

Reserve live families before applying the cap, or rank any family containing non-empty `Live` metadata ahead of non-live families.

collector/internal/vendors/cursor/live.go
```diff
@@ -0,0 +1,92 @@
...
+		}
+		if len(parts) <= idIndex+1 || !transcriptIDPattern.MatchString(parts[idIndex]) {
+			continue
+		}
+		if child != "" && parts[idIndex+1] != child {
+			continue
+		}
+		name := parts[len(parts)-1]
+		if name == database || name == database+"-wal" || name == database+"-shm" {
+			live[parts[idIndex]] = true
```
luhe19001, line 68: **[P1] Canonicalize lsof session IDs**

`transcriptIDPattern` accepts uppercase UUIDs, but this stores the directory name verbatim while transcript and side-store metadata IDs are normalized to lowercase. For an uppercase or mixed-case store directory, the resulting metadata lookup uses a different key, its entrypoint never matches, and an old active session loses its live status.

Apply `canonicalCursorID` before inserting and reconciling lsof-derived IDs.

collector/internal/diagnostics/snapshot.go
```diff
@@ -260,10 +265,23 @@ func collectSource(
 		})
 	}
 	if path, err := exec.LookPath(health.Agent); err == nil {
```
luhe19001, line 267: **[P2] Distinguish the IDE from its shell launcher**

`exec.LookPath("cursor")` only detects the optional shell launcher. A normal macOS installation can have `Cursor.app` installed without that command on `PATH`, causing every diagnostic surface to report `IDE: not found` even though the IDE is present.

Probe the application bundle separately, or label this field as launcher availability rather than IDE installation.

collector/internal/diagnostics/snapshot.go
```diff
@@ -237,6 +238,10 @@ func collectSource(
 		SkippedTotal: health.SkippedTotal,
 		CLI:          CLI{Name: health.Agent},
 	}
+	if health.Agent == "cursor" {
+		source.CLI.Name = "agent"
```
luhe19001, line 242: **[P2] Make Cursor installation checks lane-aware**

Changing `CLI.Name` to `agent` feeds into the unchanged generic checks. A machine containing only Cursor IDE sessions and the `cursor` launcher now receives a warning to install `agent` to resume those sessions, while an empty Cursor source always instructs the user to run `agent`; neither message reflects the IDE lane.

Handle Cursor's IDE and CLI availability separately in `derive` and `sourceCheck`, or suppress the generic CLI remediation for this combined source.

collector/internal/vendors/cursor/metadata.go
```diff
@@ -148,6 +148,11 @@ func loadMetadataForSessions(home string, ids, transcriptPaths []string) (*vendo
 			metadata.Session(id).Entrypoint = lane
 		}
 	}
+	for id, lane := range loadLiveSessions() {
```
luhe19001, line 151: **[P2] Avoid repeating global process probes for previews**

`GetSessionFamily` calls `LoadMetadataForSessions` twice, so this global probe executes two `ps` commands and up to four `lsof` commands for one preview. The first metadata load is used only to discover relationships, making its liveness probe unnecessary and adding avoidable preview latency.

Probe liveness only during the final load, or obtain it once and reuse the result across both metadata passes.

luhe19001 (commented):
## Review summary

This PR adds Cursor IDE/CLI diagnostics and derives session liveness from open Cursor stores. The prior issues are resolved, but authoritative liveness can still be lost during identity matching and candidate limiting; those correctness gaps should be addressed before merge.

**Findings:** 0 blocking (P0), 2 should-fix (P1), 3 nice-to-have (P2) — see inline comments.

**Prior findings:** 4 resolved, 0 closed, 0 deferred, 0 open.
**This review:** 4 resolved, 0 closed, 0 deferred, 0 reopened, 5 new.

Review profile: Helu

Requested in [PR comment](https://github.com/centauri-ai/coslash/pull/189#issuecomment-5734081215)

collector/internal/vendors/cursor/source.go
calvintvu: Agreed. The session passed the live eligibility check but could still lose the final 2,000-family ranking to newer persisted timestamps. Live members now give their whole family the highest ranking, with a regression covering 2,001 eligible families.

collector/internal/vendors/cursor/live.go
calvintvu: Agreed. IDs parsed from lsof paths are now canonicalized through the same lowercase normalization used by transcript and metadata discovery. The regression uses an uppercase UUID and verifies that only the lowercase key is returned.

collector/internal/diagnostics/snapshot.go
calvintvu: Agreed. Diagnostics now probe the standard user and system Cursor.app executable paths before falling back to the optional cursor shell launcher. A test covers an installed application bundle with no launcher on PATH.

collector/internal/diagnostics/snapshot.go
calvintvu: Agreed. The generic CLI warning is now limited to single-lane sources, so Cursor is not told that agent is required when its IDE lane is available. Empty-source remediation now explicitly names both the Cursor IDE chat and agent CLI paths.

collector/internal/vendors/cursor/metadata.go
calvintvu: Agreed. The first family lookup now uses a relationship-only state.vscdb load. Full metadata, including process-backed liveness, is loaded only once after the family files are known. The focused regression verifies that the relationship pass does not populate unrelated enrichment.

calvintvu:
@agent-review

collector/internal/vendors/cursor/source.go
milanshen: **[P1 — RESOLVED]** Fixed in `025cbe728b7ecabf0873d7bc85d90d9dc766919f`: a supplied live enrichment now gives its complete family the highest candidate-limit rank. The separate production metadata-ordering gap is reported as a new finding.

collector/internal/vendors/cursor/live.go
milanshen: **[P1 — RESOLVED]** Fixed in `025cbe728b7ecabf0873d7bc85d90d9dc766919f`: every lsof-derived ID is canonicalized before insertion and reconciliation.

collector/internal/diagnostics/snapshot.go
milanshen: **[P2 — RESOLVED]** Fixed in `025cbe728b7ecabf0873d7bc85d90d9dc766919f`: diagnostics probe the standard user and system Cursor application bundles before falling back to the optional launcher.

collector/internal/diagnostics/snapshot.go
milanshen: **[P2 — RESOLVED]** Fixed in `025cbe728b7ecabf0873d7bc85d90d9dc766919f`: the single-lane CLI warning no longer fires for a combined Cursor source, and empty-source remediation names both IDE chat and `agent` CLI paths.

collector/internal/vendors/cursor/metadata.go
milanshen: **[P2 — RESOLVED]** Fixed in `025cbe728b7ecabf0873d7bc85d90d9dc766919f`: the first preview-family pass uses relationship-only metadata, leaving process-backed liveness probing to the final full load.

collector/internal/vendors/cursor/source.go
```diff
@@ -328,6 +332,10 @@ func selectCursorFilesSourceWithMetadata(source vendors.ReadSource, files []stri
...
```
milanshen, line 335: **[P1] Load liveness before filtering Cursor files**

`Collect` filters using `LoadSelectionMetadata`, which loads relationships and timestamps but never calls `loadLiveSessions`. Liveness is discovered only by the subsequent full metadata load, after an old open session has already failed this eligibility check or the 2,000-family limit. Consequently, the new live-priority branches are exercised only by tests that inject `Live` manually.

Load lane-aware liveness into the selection metadata, or obtain it separately before filtering and ranking families.

collector/internal/vendors/cursor/source.go
```diff
@@ -79,7 +79,7 @@ func GetSessionFamily(id string) ([]*vendors.ParsedSession, *vendors.SessionMeta
 		}
 	}
 	metadata := vendors.BestEffortMetadata(vendors.AgentCursor, func() (*vendors.SessionMetadata, error) {
-		return LoadMetadataForSessions([]string{id}, requestedFiles)
+		return LoadRelationshipMetadataForSessions([]string{id})
```
milanshen, line 82: **[P1] Keep CLI relationships in the preview pass**

The replacement loader reads relationships only from the IDE `state.vscdb`, while Cursor CLI parent relationships are stored in each chat store's `subagentInfo.parentAgentId`. For a CLI child represented by a standalone transcript, the first pass now selects only that child; the final metadata pass assigns its missing parent, causing `selectFamily` to return no preview because the parent was never parsed.

Keep the relationship-only optimization, but also load the requested CLI chat-store relationships without running the global liveness probes.

collector/internal/diagnostics/checks.go
```diff
@@ -23,7 +23,7 @@ func derive(snapshot *Snapshot) []Check {
 		if source.State == SourceUnreadable {
 			scanFailed = true
 		}
-		if source.Entries > 0 && !source.CLI.Found {
+		if source.Entries > 0 && !source.CLI.Found && source.IDE == nil {
```
milanshen, line 26: **[P1] Warn when both Cursor lanes are unavailable**

Every Cursor source receives a non-nil `IDE` descriptor, even when `IDE.Found` is false. A machine retaining Cursor session history after both the IDE and `agent` CLI are removed therefore gets no missing-tool warning and can present an otherwise healthy source check despite being unable to resume sessions.

Check both lanes' `Found` values and emit a combined Cursor remediation when neither is available.

collector/internal/vendors/cursor/metadata.go
```diff
@@ -172,6 +195,11 @@ func loadMetadataForSessions(home string, ids, transcriptPaths []string) (*vendo
...
+		if lane != "" && metadata.Session(id).Entrypoint == lane {
```
milanshen: **[P2] Avoid inserting unrelated live-session metadata**

`metadata.Session(id)` mutates the map before checking whether the live ID matches an entrypoint. A scoped metadata load for one preview therefore gains an empty entry for every unrelated open Cursor session, violating the loader's requested-family scope and making later metadata loops process phantom entries.

Use `Lookup(id)` and set `Live` only when an existing enrichment has the matching entrypoint.

collector/internal/diagnostics/snapshot.go
```diff
@@ -259,16 +265,46 @@ func collectSource(
...
+	}
 	return source
 }
 
+func cursorIDEExecutable(home string) string {
+	for _, path := range []string{
+		filepath.Join(home, "Applications", "Cursor.app", "Contents", "MacOS", "Cursor"),
+		"/Applications/Cursor.app/Contents/MacOS/Cursor",
+	} {
+		if info, err := os.Stat(path); err == nil && !info.IsDir() {
```
milanshen: **[P2] Require the application binary to be executable**

This accepts any non-directory at the Cursor application path. If the bundle's binary has lost execute permissions, diagnostics still report the IDE as found while version probing silently fails, masking the installation problem the diagnostic is intended to identify.

Require a regular executable file, including an execute-mode check, before returning the bundle path.

milanshen (commented):
## Review summary

The overall approach is sound, but production integration gaps can still omit user-visible Cursor sessions and previews. Diagnostics also mishandles several realistic installation states.

**Findings:** 0 blocking (P0), 3 should-fix (P1), 2 nice-to-have (P2) — see inline comments.

**Prior findings:** 9 resolved, 0 closed, 0 deferred, 0 open.
**This review:** 5 resolved, 0 closed, 0 deferred, 0 reopened, 5 new.

Review profile: Milan

Requested in [PR comment](https://github.com/centauri-ai/coslash/pull/189#issuecomment-5736092043)

collector/internal/vendors/cursor/source.go
calvintvu: Agreed. Selection metadata now applies lane-aware process liveness before file filtering and ranking. Live CLI IDs are included even without IDE state, while IDE-derived IDs retain IDE provenance so a process from the wrong lane cannot mark them live.

collector/internal/vendors/cursor/source.go
calvintvu: Agreed. The relationship-only preview pass now reads the requested Cursor CLI chat store and applies its parentAgentId relationship without invoking the global process probes or loading unrelated enrichment.

collector/internal/diagnostics/checks.go
calvintvu: Agreed. Diagnostics now emit a combined Cursor launch-tools warning when session history exists but neither Cursor IDE nor the agent CLI is available, while still accepting either lane independently.

collector/internal/vendors/cursor/metadata.go
calvintvu: Agreed. Scoped full metadata now uses Lookup for process-derived IDs and only marks an existing matching-lane entry live, so unrelated open sessions no longer create phantom metadata entries.

collector/internal/diagnostics/snapshot.go
calvintvu: Agreed. App-bundle detection now requires a regular file with at least one execute bit. A focused regression verifies that a non-executable bundle binary is rejected.

calvintvu:
@agent-review

collector/internal/vendors/cursor/source.go
milanshen: **[P1 — RESOLVED]** Fixed in `1773ff122eaa9f98df759fcfe59b9514d1d145a8`: selection metadata now applies lane-aware process liveness before Cursor families are filtered and ranked.

milanshen:

luhe19001 (approved):

collector/internal/vendors/cursor/source.go
milanshen: **[P1 — RESOLVED]** Fixed in `1773ff122eaa9f98df759fcfe59b9514d1d145a8`: the relationship-only preview pass reads the requested CLI chat store and records its parent relationship without running global liveness probes.

collector/internal/diagnostics/checks.go
milanshen: **[P1 — RESOLVED]** Fixed in `1773ff122eaa9f98df759fcfe59b9514d1d145a8`: diagnostics now emit a combined launch-tools warning when both Cursor IDE and agent CLI are unavailable.

collector/internal/vendors/cursor/metadata.go
milanshen: **[P2 — RESOLVED]** Fixed in `1773ff122eaa9f98df759fcfe59b9514d1d145a8`: scoped full metadata uses non-mutating lookup and no longer creates entries for unrelated live sessions.

collector/internal/diagnostics/snapshot.go
milanshen: **[P2 — RESOLVED]** Fixed in `1773ff122eaa9f98df759fcfe59b9514d1d145a8`: application-bundle detection now requires a regular file with an execute mode bit.

collector/internal/vendors/cursor/metadata.go
```diff
@@ -172,9 +240,23 @@ func loadMetadataForSessions(home string, ids, transcriptPaths []string) (*vendo
...
 }
 
+func applyCursorLiveness(metadata *vendors.SessionMetadata, live map[string]string, includeUnknown bool) {
+	for id, lane := range live {
+		entry := metadata.Lookup(id)
+		if entry == nil && includeUnknown && lane != "" {
+			entry = metadata.Session(id)
+			entry.Entrypoint = lane
+		}
+		if entry != nil && lane != "" && (entry.Entrypoint == "" || entry.Entrypoint == lane) {
```
milanshen: **[P1] Preserve the lane-collision guard for liveness**

When one ID matches multiple Cursor lanes, the preceding reconciliation deliberately clears lane-specific enrichment and leaves `Entrypoint` empty. This new empty-entrypoint allowance then marks that ambiguous record interactive whenever one colliding lane is live, reattaching a signal whose provenance cannot be established.

Require exact lane equality for existing entries; the `includeUnknown` branch already assigns a lane to genuinely new selection entries.

**Change evidence:** Since `025cbe728b7e`, `collector/internal/vendors/cursor/metadata.go:254` (RIGHT): The previous implementation required `Entrypoint == lane`, so records rejected by the multi-lane collision guard could not regain liveness.

milanshen (commented):
### Review 4 Summary

**Score: 2/5** — coverage incomplete

This head completes Cursor liveness and relationship handling while adding exact session details and diffs. The five prior scenarios are fixed, but a blocking supported-vendor integration regression and several correctness and performance issues remain. The 42-file endpoint delta was reviewed with parallel scoped passes; unchanged original scope remains uncertified because historical coverage is unknown.

**Findings:** 1 blocking (P0), 3 should-fix (P1), 0 nice-to-have (P2) — see inline comments.

**Prior findings:** 14 resolved, 0 closed, 0 deferred, 0 open.
**This review:** 5 resolved, 0 closed, 0 deferred, 0 reopened, 4 new.

Findings without a valid inline location:
- [collector/cmd/coslash/api_detail.go:242] **[P0] Allow Cursor through exact-detail validation**

The inspector now sends every session with a detail revision through this endpoint, and Cursor roots receive such revisions, but this allowlist omits `vendors.AgentCursor`. Opening any Cursor card therefore returns HTTP 400 before `collector.GetSessionDetail`, replacing the previously working read-only inspector with a permanent error.

Include Cursor in the supported local-agent set, or deliberately retain a summary-only path for Cursor.

**Change evidence:** Since `025cbe728b7e`, `collector/cmd/coslash/api_detail.go:242` (RIGHT): The newly added allowlist gates the exact-detail route used by the inspector; the baseline rendered Cursor data directly and had no corresponding rejection.
- [frontend/src/pages/coslash/components/SessionInspector.tsx:140] **[P1] Do not label summary-only data as a complete cache**

For a remote session with an empty `detailRevision` on a stale or offline machine, `useSessionDetail` returns `cachedOffline: false` and `summaryOnly: true`. This helper ignores that false flag whenever machine state is present, so the inspector simultaneously says it is showing the last complete cached details and that complete details are unavailable.

Require `cachedOffline` for the cache banner, or suppress it whenever `summaryOnly` is true.

**Change evidence:** Since `025cbe728b7e`, `frontend/src/pages/coslash/components/SessionInspector.tsx:140` (RIGHT): The endpoint delta introduces both status banners and derives the complete-cache claim from machine health alone; the baseline displayed neither claim.
- [collector/internal/vendors/claude/discovery.go:94] **[P1] Avoid rescanning Claude history for every selected root**

A windowed refresh now calls `familyFiles` once per eligible root. That helper rescans the complete file list, rebuilds the owner index, and—whenever any subagent exists—parses the root before `Collect` parses it again; with many recent roots this becomes O(selected roots × all files) work on every poll.

Build the owner and re-home index once, then expand all selected background families in a single traversal.

**Change evidence:** Since `025cbe728b7e`, `collector/internal/vendors/claude/discovery.go:94` (RIGHT): The baseline used the two-pass `FilesSinceSource` result directly and did not perform a complete-history scan and extra root parse for each selected family.



Review profile: Milan

Requested in [PR comment](https://github.com/centauri-ai/coslash/pull/189#issuecomment-5736605176)

collector/internal/vendors/cursor/metadata.go
calvintvu: Agreed. Commit 82ab87a0 now requires exact entrypoint equality for existing metadata. Selection-only unknown sessions still receive their detected lane before the check, while multi-lane entries with an intentionally empty entrypoint remain non-interactive. The regression reproduces and protects that collision case.

calvintvu: **[P1 - RESOLVED]** Fixed in `82ab87a00b9eb08fcacea952f15566c05cf4d819`: existing metadata now requires exact entrypoint equality before liveness is attached, preserving the lane-collision guard.

collector/internal/diagnostics/checks_test.go
```diff
@@ -0,0 +1,68 @@
...
+	t.Setenv("PATH", "")
+	home := t.TempDir()
+	path := filepath.Join(home, "Applications", "Cursor.app", "Contents", "MacOS", "Cursor")
+	if err := os.MkdirAll(filepath.Dir(path), 0o755); err != nil {
+		t.Fatal(err)
+	}
+	if err := os.WriteFile(path, nil, 0o644); err != nil {
+		t.Fatal(err)
+	}
+	if got := cursorIDEExecutable(home); got != "" {
```
calvintvu: **[P1] Isolate the executable-rejection test from system Cursor installs**

The test clears `PATH` and creates a non-executable user bundle, but `cursorIDEExecutable` also probes `/Applications/Cursor.app/Contents/MacOS/Cursor`. On a normal macOS machine with Cursor installed there, the resolver correctly returns the system executable and this test fails while expecting an empty string.

Inject or isolate candidate paths for this test, or make the assertion account for the valid system fallback.

calvintvu (commented):
## Review summary

**Score: 4/5** - not safe to merge

Cursor liveness now preserves lane provenance, and the applicable prior inline findings are resolved. One environment-dependent test still fails on a standard macOS installation where Cursor exists in the system Applications directory.

**Findings:** 0 blocking (P0), 1 should-fix (P1), 0 nice-to-have (P2) - see inline comments.

**Prior findings:** 15 resolved, 3 closed as outside the current PR diff, 0 deferred, 0 open.
**This review:** 1 resolved, 3 closed, 0 deferred, 0 reopened, 1 new.

collector/internal/diagnostics/checks_test.go
calvintvu: **[P1 - RESOLVED]** Fixed in `7cf292ec2021bf4797967360390c4a4d7d6f00d2`: the test now asserts that the non-executable user bundle is skipped while allowing the resolver to return a valid system installation fallback.

PR #194, opened 2026-09-16
Add background code review sessions

calvintvu:
## Context

Reviewing an agent’s changes currently requires manually opening another coding agent and reconstructing the relevant context. This adds a one-click, read-only review flow that runs in the background and records the result as a normal agent session.

## Changes

- Add a reviewer dialog for installed Claude Code, Codex, and OpenCode agents (Cursor deferred).
- Run reviews headlessly with read-only permissions, progress polling, failure handling, and retry support.
- Link review sessions to their origin while preserving multiple review attempts as ordinary sessions.

## Test

- `npm --prefix frontend test` — 102 tests passed
- `npm --prefix frontend run build` — passed
- `npm --prefix frontend run lint` — passed with existing warnings
- Manual: launched a background review and verified picker, spinner, session naming, and reciprocal links in the app.

### Screenshots

Reviewer picker:
reviewer-dialog

Background review progress:

review-progress-full


collector/internal/launch/launch.go
```diff
@@ -49,6 +55,77 @@ type terminalAdapter struct {
...
+		return fmt.Errorf("launch: session has no working directory")
+	}
+	spec, err := reviewCLICommand(request.Reviewer, workingDirectory, request.Name, request.Prompt)
+	if err != nil {
+		return err
+	}
+	command := exec.CommandContext(ctx, spec.bin, spec.args...)
+	command.Dir = workingDirectory
+	command.Stdout = io.Discard
+	var stderr bytes.Buffer
```
luhe19001: **[P1] Bound and sanitize reviewer stderr**

Reviewer stderr is accumulated without a size limit, returned from `Review`, persisted in manager state, and ultimately exposed as `reviewError` through `/api/sessions`. A noisy failing plugin can consume arbitrary collector memory, while ordinary authentication or configuration failures can disclose internal paths and diagnostics to the browser. This violates `collector/CLAUDE.md`: “Keep internal errors in server logs and return fixed, non-sensitive messages to clients.”

Capture only a bounded diagnostic, log it server-side, and store a fixed client-safe failure message, following the synthesis command-error pattern.

collector/internal/launch/launch.go
```diff
@@ -49,6 +55,77 @@ type terminalAdapter struct {
...
+		}
+		return err
+	}
+	return nil
+}
+
+func reviewCLICommand(reviewer, workingDirectory, name, prompt string) (reviewCommandSpec, error) {
+	switch reviewer {
+	case vendors.AgentClaude:
+		return reviewCommandSpec{bin: "claude", args: []string{"-p", "--name", name, "--permission-mode", "plan", prompt}}, nil
```
luhe19001: **[P1] Keep the review prompt out of argv**

`Prompt` includes an uncapped outcome, every changed path, and every commit, and each reviewer receives the entire result as one command-line argument. A sufficiently large session will exceed the OS argument limit and fail before the reviewer starts; the prompt is also exposed through process arguments. The synthesis runner already bounds session input and passes it through stdin.

Apply per-field/list and total-size limits, then pass review context through stdin where supported instead of embedding it in argv.

collector/internal/review/review.go
```diff
@@ -0,0 +1,140 @@
...
+
+func Name(originName, originID string) string {
+	if originName == "" {
+		originName = "Untitled session"
+	}
+	shortID := originID
+	if len(shortID) > 8 {
+		shortID = shortID[:8]
+	}
+	return fmt.Sprintf("%s%s (%s)", prefix, originName, shortID)
```
luhe19001, line 131: **[P1] Normalize multiline origin names before embedding them**

`originName` is inserted verbatim into the synthetic first line, but `NameFromPrompt` examines only text before the first newline and expects the short ID there. For a Codex session whose fallback name is a multiline initial prompt, the launched review therefore cannot be recognized or linked to its origin; the UI treats it as an ordinary session and offers to review it again.

Collapse and cap the origin name to one line—using `session.Truncate`, for example—before constructing the review name, and cover a multiline origin title in the tests.

luhe19001 (commented):
## Review summary

The review feature is coherently integrated across the collector, launch manager, and UI, and its route/auth handling follows repository conventions. Several edge cases still create unsafe error/resource handling or break review-session identity. Static checks completed, but tests could not run because Go and frontend dependencies are unavailable.

**Findings:** 0 blocking (P0), 3 should-fix (P1), 0 nice-to-have (P2) — see inline comments.

Review profile: Helu

collector/internal/launch/launch.go
calvintvu: Fixed in 7e00d43. Reviewer stderr is capped at 8 KiB, retained only in the internal error, and the review manager now logs that diagnostic while exposing a fixed client-safe failure message. Added focused coverage for the bounded diagnostic writer; full collector tests and vet pass.

collector/internal/launch/launch.go
calvintvu: Fixed in 7e00d43. All three reviewer prompts now travel over stdin instead of argv (Codex uses its explicit `-` stdin form), and prompt fields, list lengths, and total bytes are bounded. Command-spec and prompt-bound regression tests cover the behavior; full collector tests and vet pass.

collector/internal/review/review.go
calvintvu: Fixed in 7e00d43. Review names now collapse whitespace and cap the origin name before constructing the synthetic first line, preserving name-based identity for multiline fallback prompts. Added a multiline-name regression test; full collector tests and vet pass.

calvintvu:
@agent-review: he

calvintvu:
@agent-review: milan

luhe19001 (commented):
## Review summary

**Carried over:** 3 resolved, 0 unresolved (3 prior findings).

| --- | --- | --- | --- |
| `launch.go`: bound and sanitize reviewer stderr | P1 | Resolved | Stderr is capped at 8 KiB; details are logged while clients receive a fixed message. |
| `launch.go`: keep review prompt out of argv | P1 | Resolved | Prompt fields and total bytes are bounded, and all reviewers receive it through stdin. |
| `review.go`: normalize multiline origin names | P1 | Resolved | `session.Truncate` collapses whitespace before constructing the synthetic name. |

The background-review architecture and prior hardening fixes are generally sound. Two functional edge cases remain around OpenCode review capability and identifying review sessions without their origin. Go and frontend tests could not be rerun because the required toolchain/dependencies are unavailable in this checkout.

**Findings:** 0 blocking (P0), 2 should-fix (P1), 2 nice-to-have (P2) — see inline comments.

Review profile: Helu

collector/internal/launch/launch.go
```diff
@@ -49,6 +55,93 @@ type terminalAdapter struct {
...
+	switch reviewer {
+	case vendors.AgentClaude:
+		return reviewCommandSpec{bin: "claude", args: []string{"-p", "--name", name, "--permission-mode", "plan"}, stdin: prompt}, nil
+	case vendors.AgentCodex:
+		return reviewCommandSpec{bin: "codex", args: []string{"exec", "--sandbox", "read-only", "--skip-git-repo-check", "-"}, stdin: prompt}, nil
+	case vendors.AgentOpenCode:
+		return reviewCommandSpec{
+			bin:   "opencode",
+			args:  []string{"run", "--title", name, "--dir", workingDirectory},
+			env:   []string{`OPENCODE_PERMISSION={"edit":"deny","bash":"deny"}`},
```
luhe19001: **[P1] Let OpenCode obtain the working-tree diff**

The OpenCode launch denies `bash` entirely, while stdin contains only the review prompt and contextual filenames—not the actual diff. For an origin with uncommitted edits, OpenCode therefore cannot run `git diff` or inspect the changed hunks, so it can only review final file contents without knowing what changed.

Provide a bounded diff as review input or narrowly permit read-only Git commands while continuing to deny edits.

frontend/src/pages/coslash/components/SessionCard.tsx
```diff
@@ -464,7 +473,13 @@ export function SessionCard({
   onClick,
   variant = 'detailed',
   showMachineBadge = false,
+  reviewerOptions = [],
+  reviewLink,
+  reviewActive = false,
+  onReviewStarted = () => undefined,
+  onSelectRelated,
 }: SessionCardProps) {
+  const showReviewAction = isLocalSession(session) && reviewLink?.kind !== 'review';
```
luhe19001: **[P1] Recognize review sessions without requiring their origin**

`showReviewAction` treats a session as a review only when `buildReviewLinks` found its origin. Start a review for an old session while viewing “All,” then switch to “7 days”: the new review remains in the response but its old origin does not, so `reviewLink` is undefined and the review card offers “Send for review” again.

Track whether the card itself has the validated review marker separately from whether its origin is currently linkable, and suppress recursive review actions in either case.

collector/cmd/coslash/api.go
```diff
@@ -324,6 +329,58 @@ func handleLaunch(w http.ResponseWriter, r *http.Request, settingsStore *setting
...
+		http.Error(w, "session has no working directory", http.StatusConflict)
+		return
+	}
+	originName := ""
+	if found.Name != nil {
+		originName = *found.Name
+	}
+	name := reviewpkg.Name(originName, found.ID)
+	prompt := reviewpkg.Prompt(found)
+	startReview(found.ID, reviewpkg.Launch{
```
luhe19001: **[P2] Report rejected duplicate review starts**

`Manager.Start` returns `false` when the origin already has a pending review, but the handler ignores that result and always logs a launch and returns `202`. With two tabs selecting different reviewers concurrently, the second dialog closes as successful even though its requested reviewer never starts.

Return a conflict or an explicit already-running response when `startReview` returns `false`.

frontend/src/pages/coslash/lib/review.ts
```diff
@@ -0,0 +1,87 @@
...
+    }
+  }
+  return links;
+}
+
+export function activeReviewForOrigin(
+  origin: ReviewableSession,
+  sessions: readonly ReviewableSession[],
+): boolean {
+  return sessions.some((candidate) => {
```
luhe19001: **[P2] Index active reviews once per session refresh**

Every rendered card calls `activeReviewForOrigin`, which scans the complete session library; `buildReviewLinks` also filters all origins once per review. In an all-history view containing thousands of sessions, this produces millions of repeated regex and prefix comparisons on every refresh and can noticeably stall the board.

Build a memoized source/short-ID index and active-origin set once when the session payload changes, then use constant-time lookups while rendering cards.

milanshen:

collector/internal/launch/launch.go
calvintvu: Fixed in 20d165b by including bounded recorded change hunks in the review prompt. OpenCode keeps bash fully denied, so it receives the actual session diff without broadening write-capable shell access. Prompt coverage now asserts the diff content, and the total prompt remains capped.

frontend/src/pages/coslash/components/SessionCard.tsx
calvintvu: Fixed in 20d165b. The review index now records every session with a valid review-name marker independently of whether its origin is present in the current session payload. SessionCard uses that marker to suppress recursive review actions; the orphan/ambiguous-origin test covers this case.

collector/cmd/coslash/api.go
calvintvu: Fixed in 20d165b. The handler now checks Manager.Start and returns HTTP 409 with `review already running` when a concurrent request loses the deduplication race. Added a handler regression test.

frontend/src/pages/coslash/lib/review.ts
calvintvu: Fixed in 20d165b. A single memoized review index is built when the session payload changes; links, review markers, and active-origin state are then constant-time lookups for both list and board cards. The existing link, newest-review, active-review, and ambiguous-ID cases now exercise the index.

luhe19001 (commented):
**Carried over:** 7 resolved, 0 unresolved (7 prior findings).

| --- | --- | --- | --- |
| `launch.go` — bound and sanitize reviewer stderr | P1 | Resolved | Stderr is capped at 8 KiB, logged internally, and replaced with a fixed client-safe error. |
| `launch.go` — keep review prompt out of argv | P1 | Resolved | All reviewers receive a bounded 16 KiB prompt over stdin. |
| `review.go:99` — normalize multiline origin names | P1 | Resolved | `session.Truncate` collapses whitespace and caps the generated name. |
| `launch.go:124` — provide OpenCode with change hunks | P1 | Resolved | The prompt now includes bounded recorded change hunks, satisfying the requested bounded-diff option. |
| `SessionCard.tsx` — recognize reviews without their origin | P1 | Resolved | Review sessions are marked independently of whether an origin link can be resolved. |
| `api.go` — report duplicate starts | P2 | Resolved | A failed `Manager.Start` now returns HTTP 409. |
| `review.ts` — index active reviews once | P2 | Resolved | A memoized linear-time index supplies constant-time card lookups. |

## Review summary

This PR adds background review launches, review-session linking, and UI state for supported agent CLIs. The carried-over fixes are sound, but the automated launch boundary still permits repository-controlled reviewer configuration to execute, which blocks merge. Static diff checks passed; Go and frontend tests could not run because the Go toolchain and frontend dependencies are unavailable.

**Findings:** 1 blocking (P0), 1 should-fix (P1), 2 nice-to-have (P2) — see inline comments.

Review profile: Helu

collector/internal/launch/launch.go
```diff
@@ -49,6 +55,93 @@ type terminalAdapter struct {
...
+		}
+		return err
+	}
+	return nil
+}
+
+func reviewCLICommand(reviewer, workingDirectory, name, prompt string) (reviewCommandSpec, error) {
+	switch reviewer {
+	case vendors.AgentClaude:
+		return reviewCommandSpec{bin: "claude", args: []string{"-p", "--name", name, "--permission-mode", "plan"}, stdin: prompt}, nil
```
luhe19001, line 119: **[P0] Disable project code when launching reviewers**

These commands run inside the session working directory, but plan/read-only tool permissions do not disable repository hooks, plugins, MCP configuration, or other startup code. In particular, `claude -p` skips the workspace-trust dialog and this invocation omits `--safe-mode`; OpenCode also loads project plugins and configuration from its working directory despite `bash` being denied. If a Codex-origin checkout contains a malicious Claude project hook and the user selects Claude as reviewer, that hook can execute as the collector user before the review begins.

Use the hardened modes already established by the synthesis runner: add Claude safe/restricted configuration, ignore Codex user/project rules as appropriate, and run OpenCode with project configuration/plugins disabled.

collector/cmd/coslash/api.go
```diff
@@ -324,6 +329,61 @@ func handleLaunch(w http.ResponseWriter, r *http.Request, settingsStore *setting
...
+	w http.ResponseWriter,
+	r *http.Request,
+	settingsStore *settings.Store,
+	getSession func(string) (*session.Session, error),
+	reviewerAvailable func(string) bool,
+	startReview reviewStarter,
+) {
+	state := settingsStore.State()
+	if !state.Valid {
+		http.Error(w, state.Error+"; open Settings to repair it", http.StatusConflict)
```
luhe19001: **[P1] Return a fixed settings failure message**

When `settings.json` is unreadable or malformed, `State.Error` can contain an OS path or parser diagnostic, and this new endpoint returns it verbatim to the browser. This directly violates `collector/CLAUDE.md`: “Keep internal errors in server logs and return fixed, non-sensitive messages to clients.”

Log `state.Error` server-side and return a fixed message such as `settings are invalid; open Settings to repair them`.

frontend/src/pages/coslash/lib/review.ts
```diff
@@ -0,0 +1,86 @@
...
+
+function reviewerRank(reviewer: string, originAgent: string): number {
+  if (reviewer === originAgent) return 2;
+  return reviewer === 'codex' ? 0 : 1;
+}
+
+export function reviewRequestPath(origin: SessionIdentity, reviewer: VendorKey): string {
+  return `/api/reviews?${new URLSearchParams({
+    source: origin.sourceId,
+    id: origin.id,
```
luhe19001, line 46: **[P2] Preserve the origin agent in review requests**

A session identity includes `agent`, but this request sends only source and ID. The server consequently searches Claude, Codex, and OpenCode stores in fixed order and keys review state solely by ID. If copied or restored Claude and Codex histories contain the same UUID, clicking the Codex card can review the Claude session instead, while pending/error state is shared between both cards.

Include the origin agent in the request, resolve the session within that vendor store, and key manager state by the complete local session identity.

collector/internal/collector/collector.go
```diff
@@ -409,6 +410,12 @@ func probeGitEnvironment(roots []*vendors.ParsedSession) {
 func resolveNames(roots []*vendors.ParsedSession, metadata map[string]*vendors.SessionMetadata) {
 	for _, p := range roots {
 		s := p.Session
+		if s.FirstPrompt != nil {
+			if name, ok := review.NameFromPrompt(*s.FirstPrompt); ok {
```
luhe19001, line 428: **[P2] Avoid inferring review provenance from user text**

Any ordinary session whose first prompt begins with `Review — <text> (<eight characters>)` is classified as an internally generated review. Its provider title is overwritten, the review action is suppressed, and it can be linked to an unrelated session with that ID prefix. This can occur when a user pastes a review title as the first line of a normal prompt.

Track review provenance with explicit metadata or a validated opaque marker carrying the full origin identity instead of treating a broad display-name pattern as control state.

milanshen (commented):
**Carried over:** 7 resolved, 4 unresolved (11 prior findings).

| --- | --- | --- | --- |
| `launch.go` — Bound and sanitize reviewer stderr | P1 | Resolved | Stderr is capped at 8 KiB, logged internally, and replaced with a fixed client-safe failure message. |
| `launch.go` — Keep the review prompt out of argv | P1 | Resolved | All reviewers receive bounded prompts through stdin, with field, list, and total-size limits. |
| `review.go` — Normalize multiline origin names | P1 | Resolved | `session.Truncate` collapses whitespace before constructing the review name. |
| `launch.go` — Let OpenCode obtain the working-tree diff | P1 | Resolved | Bounded recorded change hunks are now included in the prompt supplied to OpenCode. |
| `SessionCard.tsx` — Recognize review sessions without their origin | P1 | Resolved | Valid review markers are indexed independently of origin lookup and suppress recursive review actions. |
| `api.go` — Report rejected duplicate review starts | P2 | Resolved | A failed `Manager.Start` now produces HTTP 409. |
| `review.ts` — Index active reviews once per refresh | P2 | Resolved | Review links and active origins are built in two linear passes and memoized per session payload. |
| `launch.go` — Disable project code when launching reviewers | P0 | Unresolved | Claude lacks safe mode, Codex lacks configuration/rule isolation, and OpenCode still loads project configuration and plugins from the checkout. |
| `api.go:344` — Return a fixed settings failure message | P1 | Unresolved | `state.Error` is still returned verbatim by the review endpoint. |
| `review.ts:45` — Preserve the origin agent | P2 | Unresolved | Requests, server lookup, manager state, and provenance still identify origins without their agent. |
| `collector.go:414` — Avoid inferring review provenance from user text | P2 | Unresolved | A spoofable display-name regex remains the sole provenance marker. |

## Review summary

This PR adds a coherent background review workflow and resolves most earlier correctness and scaling concerns. The blocking issue remains execution of repository-controlled startup code despite the advertised read-only review mode. The complete diff was reviewed statically; Go and frontend test tooling were unavailable in the checkout.

**Findings:** 1 blocking (P0), 1 should-fix (P1), 2 nice-to-have (P2) — all carried over above; no new inline findings.

Review profile: Milan

Requested in [PR comment](https://github.com/centauri-ai/coslash/pull/194#issuecomment-5703679018)

collector/internal/launch/launch.go
milanshen: not sure I fully agree -- if the permission is too strict, can the agent session still perform effective review? leave to @calvintvu for thinking this through: we want the safest permission mode that can still support an effective review

calvintvu: The product assumes users review repositories they have access to, so repository-owned startup configuration is trusted or can at least be audited. Keeping the repository as the working directory improves review quality by allowing surrounding-code inspection; reviews remain read-only, receive a bounded diff/debrief over stdin, and OpenCode keeps bash/edit denied.

collector/cmd/coslash/api.go
calvintvu: Fixed in 6647e44. The review endpoint now logs the detailed settings failure server-side and returns only the fixed message `settings are invalid; open Settings to repair them`. Added regression coverage for malformed settings.

frontend/src/pages/coslash/lib/review.ts
calvintvu: Fixed in 6647e44. Review requests now include the origin agent, lookup is restricted to that vendor store, and pending/error state is keyed by `agent:id`. Tests cover vendor-specific lookup and the complete state key.

collector/internal/collector/collector.go
calvintvu: Not changing this in the initial implementation. Name-based provenance is an explicit product constraint for this feature: no review metadata or new session category. The false-positive case requires a user-created first line matching the exact generated review-name convention and an existing matching short ID; we accept that low-probability limitation for now.

calvintvu:
@agent-review: milan

luhe19001 (commented):
**Carried over:** 8 resolved, 3 unresolved (11 prior findings).

| --- | --- | --- | --- |
| `launch.go` — Bound and sanitize reviewer stderr | P1 | Resolved | Fixed in `7e00d43`: stderr is capped at 8 KiB, logged internally, and replaced with a fixed client-safe failure message. |
| `launch.go` — Keep the review prompt out of argv | P1 | Resolved | Fixed in `7e00d43`: bounded prompts are supplied through stdin. |
| `review.go` — Normalize multiline origin names | P1 | Resolved | Fixed in `7e00d43`: `session.Truncate` collapses whitespace before constructing the review name. |
| `launch.go` — Let OpenCode obtain the working-tree diff | P1 | Reopened | The partial `20d165b` fix emits only recorded `FileEdit.Changes`; shell-command edits remain stats-only and are omitted, while OpenCode still denies bash and cannot obtain their diff. |
| `SessionCard.tsx` — Recognize review sessions without their origin | P1 | Resolved | Fixed in `20d165b`: validated review markers are indexed independently of origin lookup. |
| `api.go` — Report rejected duplicate review starts | P2 | Resolved | Fixed in `20d165b`: `Manager.Start == false` now returns HTTP 409. |
| `review.ts` — Index active reviews once per refresh | P2 | Resolved | Fixed in `20d165b`: links and active origins are built in two linear passes and memoized. |
| `launch.go` — Disable project code when launching reviewers | P0 | Unresolved | Reviewers still inherit repository-controlled startup configuration, hooks, plugins, and environment while running in the checkout. |
| `api.go` — Return a fixed settings failure message | P1 | Resolved | Fixed in `6647e44`: the detailed error is logged server-side and the endpoint returns fixed text. |
| `review.ts` — Preserve the origin agent | P2 | Resolved | Fixed in `6647e44`: requests, lookup, and manager state use the complete agent-qualified identity. |
| `collector.go` — Avoid inferring review provenance from user text | P2 | Unresolved | The backend and frontend still treat a user-controlled display-name regex as review provenance. |

## Review summary

This PR adds an end-to-end background review workflow, and most earlier correctness and scaling issues are now fixed. Repository-controlled startup code remains the blocking concern, and the OpenCode diff fix is incomplete for shell-based edits. Static review covered all changed hunks and all nine required angles.

**Findings:** 1 blocking (P0), 4 should-fix (P1), 2 nice-to-have (P2) — 3 carried over above and 4 new inline comments.

Review profile: Helu

collector/internal/launch/launch.go
```diff
@@ -49,6 +55,93 @@ type terminalAdapter struct {
...
+	if err != nil {
+		return err
+	}
+	command := exec.CommandContext(ctx, spec.bin, spec.args...)
+	command.Dir = workingDirectory
+	command.Stdin = strings.NewReader(spec.stdin)
+	command.Stdout = io.Discard
+	stderr := boundedBuffer{limit: 8 << 10}
+	command.Stderr = &stderr
+	command.Env = append(os.Environ(), spec.env...)
```
luhe19001: **[P1] Set PWD to the selected OpenCode directory**

`command.Env` is always populated from `os.Environ()`, so Go deliberately does not update the inherited `PWD` when `command.Dir` changes. If coSlash starts in repository A and launches an OpenCode review for repository B, OpenCode receives A as `PWD`; this repository's synthesis runner explicitly notes that OpenCode reads `PWD` before cwd, so it can initialize against the wrong project despite `--dir`.

Build the environment from `command.Environ()` after assigning `Dir`, or explicitly set `PWD` to `workingDirectory`.

collector/internal/review/review.go
```diff
@@ -0,0 +1,190 @@
...
+		}
+		commits = append(commits, "- "+session.Truncate(commit, maxArtifactRunes))
+	}
+	if len(commits) == 0 {
+		commits = append(commits, "- —")
+	}
+	return limitBytes(strings.Join([]string{
+		Name(name, origin.ID),
+		"",
+		"Review the current working-tree changes. Use an installed code-review skill or your native review capability when available. Do not modify files.",
```
luhe19001, line 204: **[P1] Snapshot the selected session's review target**

The reviewer is instructed to inspect the mutable current working tree rather than an immutable revision belonging to the selected session. If that session committed its work and the checkout later changes branches—or another session edits the same directory—the reviewer examines a clean or unrelated tree; the bounded recorded snippets do not identify a complete base/head, and the available `CommitSHAs` are omitted.

Capture the exact base, commit SHAs, and bounded uncommitted diff when starting the review, or run against an isolated snapshot whose state is verified before launch.

collector/internal/review/review.go
```diff
@@ -0,0 +1,190 @@
...
+func (m *Manager) Start(originID string, launch Launch) bool {
+	m.mu.Lock()
+	if m.states[originID].Pending {
+		m.mu.Unlock()
+		return false
+	}
+	m.states[originID] = State{Pending: true}
+	m.mu.Unlock()
+
+	go func() {
```
luhe19001, line 77: **[P1] Bound and cancel reviewer processes**

Each distinct origin immediately starts a reviewer goroutine rooted at `context.Background()` for up to 30 minutes. Starting reviews across many cards therefore launches an unbounded number of heavyweight, billed CLI processes, and collector shutdown has no manager cancellation hook, allowing in-flight children to continue after the server exits.

Give the manager bounded worker slots plus a cancellable root context and waitable `Shutdown`, then register that shutdown with the server.

frontend/src/pages/coslash/hooks/use-sessions.ts
```diff
@@ -232,7 +233,11 @@ export function useSessions({ localWindow, remoteWindow }: SessionsQuery) {
           setSessionsVersion((version) => version + 1);
           setIsLoading(false);
           setLoadError(null);
-          scheduleRefresh();
+          scheduleRefresh(
```
luhe19001, line 236: **[P2] Poll lightweight review state instead of all sessions**

While any review is pending, this refreshes the complete session payload every three seconds. In the all-history view each request reparses vendor histories and performs filesystem/Git probes through `collector.List`; a review reaching the 30-minute timeout can cause roughly 600 full scans.

Expose a lightweight review-status endpoint or push transition, then refresh the complete library once when the review finishes.

milanshen:

collector/internal/launch/launch.go
milanshen: **[P1 — STILL OPEN]** `20d165b` only supplies recorded `FileChange` text. Shell-command edits can have counts but no retained change text, and `FileEdit.Changes()` filters those entries while OpenCode still has bash denied, so it cannot obtain their actual diff.

collector/internal/launch/launch.go
milanshen: **[P0 — STILL OPEN]** Repository access does not make repository startup code safe to execute. The codebase's own OpenCode model probe explicitly avoids an untrusted checkout because project plugins/config can run code, but these reviewer launches omit the same safe mode, project-config disabling, and isolated environment.

collector/internal/collector/collector.go
milanshen: **[P2 — STILL OPEN]** Accepting the collision as unlikely does not remove the observable behavior: a matching ordinary first prompt still overrides its provider title, suppresses the review action, and may link to an unrelated session. Explicit provenance remains necessary to distinguish product state from user text.

collector/internal/review/review.go
```diff
@@ -0,0 +1,190 @@
...
+		"Review the current working-tree changes. Use an installed code-review skill or your native review capability when available. Do not modify files.",
+		"Report findings by severity with file locations, then give a concise conclusion.",
+		"",
+		"Branch: " + branch,
+		"",
+		"Debrief:",
+		outcome,
+		"",
+		"Changed files:",
+		strings.Join(files, "\n"),
```
milanshen, line 215: **[P0] Treat recorded review context as untrusted data**

The prompt appends branch names, commit subjects, synthesis text, paths, and raw changed-file snippets directly after the reviewer instructions. A repository can put instructions such as reading a same-user credential file into a changed snippet; the selected CLI retains read and model access, so following that injected text can disclose secrets to the provider or persisted review session despite edit restrictions.

Place recorded fields in an explicitly delimited, untrusted-data block, instruct the reviewer never to follow instructions from it, and restrict filesystem and credential access to an isolated review snapshot.

frontend/src/pages/coslash/components/SessionCard.tsx
```diff
@@ -464,7 +474,14 @@ export function SessionCard({
...
   variant = 'detailed',
   showMachineBadge = false,
+  reviewerOptions = [],
+  reviewLink,
+  isReview = false,
+  reviewActive = false,
+  onReviewStarted = () => undefined,
+  onSelectRelated,
 }: SessionCardProps) {
+  const showReviewAction = isLocalSession(session) && !isReview;
```
milanshen: **[P2] Hide review actions for sessions without a directory**

This offers review for every local non-review session, although locally collected sessions with a prompt but no `cwd` can remain visible and the backend rejects them with `session has no working directory`. The user can therefore complete the reviewer picker only to encounter an avoidable conflict.

Include a nonempty working directory in the eligibility check, or pass the existing `disabled` prop with an appropriate explanation.

frontend/src/pages/coslash/components/ReviewDialog.tsx
```diff
@@ -0,0 +1,157 @@
...
+              value={reviewer.id}
+              checked={selected === reviewer.id}
+              onChange={() => onSelect(reviewer.id)}
+              disabled={launching}
+            />
+            <span className="text-sm font-semibold">{reviewer.label}</span>
+          </label>
+        ))}
+      </div>
+      {error != null && <div className="text-destructive text-xs">{error}</div>}
```
milanshen: **[P2] Announce review launch errors**

Immediate POST failures are inserted as an ordinary `div`. After a keyboard or screen-reader user activates “Start review,” focus remains on the button and the new error is not announced; the separate `role="alert"` only covers background `reviewError`, not this local launch error.

Give this dynamic error an alert or appropriate live-region role.

milanshen (commented):
**Carried over:** 8 resolved, 0 closed, 7 unresolved (15 prior findings).

| --- | --- | --- | --- |
| `launch.go` — Bound and sanitize reviewer stderr | P1 | Resolved | Fixed in `7e00d43`: stderr is capped, logged internally, and replaced with a fixed client message. |
| `launch.go` — Keep the review prompt out of argv | P1 | Resolved | Fixed in `7e00d43`: bounded prompts are supplied through stdin. |
| `review.go` — Normalize multiline origin names | P1 | Resolved | Fixed in `7e00d43`: `session.Truncate` collapses whitespace before constructing the name. |
| `launch.go` — Let OpenCode obtain the working-tree diff | P1 | Reopened | `20d165b` includes recorded `FileChange` snippets, but shell-command edits remain stats-only while OpenCode cannot run Git. |
| `SessionCard.tsx` — Recognize reviews without their origin | P1 | Resolved | Fixed in `20d165b`: review markers are indexed independently of origin lookup. |
| `api.go` — Report rejected duplicate starts | P2 | Resolved | Fixed in `20d165b`: a rejected duplicate returns HTTP 409. |
| `review.ts` — Index active reviews once per refresh | P2 | Resolved | Fixed in `20d165b`: review links and active origins are built once and use constant-time lookups. |
| `launch.go` — Disable project code when launching reviewers | P0 | Unresolved | All reviewers still start in the checkout with inherited environment and repository-controlled startup configuration enabled. |
| `api.go` — Return a fixed settings failure message | P1 | Resolved | Fixed in `6647e44`: detailed settings errors are logged and fixed text is returned. |
| `review.ts` — Preserve the origin agent | P2 | Resolved | Fixed in `6647e44`: requests, lookup, and manager state use agent-qualified identity. |
| `collector.go` — Avoid inferring provenance from user text | P2 | Unresolved | Backend and frontend still treat a user-controlled display-name pattern as review provenance. |
| `launch.go` — Set PWD to the selected OpenCode directory | P1 | Unresolved | `command.Env` is built from `os.Environ()`, retaining the collector's stale `PWD`. |
| `review.go` — Snapshot the selected review target | P1 | Unresolved | Launches still inspect a mutable pathname without a pinned revision or isolated snapshot. |
| `review.go` — Bound and cancel reviewer processes | P1 | Unresolved | Distinct origins launch unlimited processes from `context.Background`, with no manager shutdown hook. |
| `use-sessions.ts` — Poll lightweight review state | P2 | Unresolved | A pending review still triggers a complete session collection every three seconds. |

## Review summary

This PR adds an end-to-end background review workflow, and its frontend/backend wiring is coherent. It is not ready to merge because reviewer execution remains exposed to repository-controlled behavior, while target and lifecycle gaps can review incorrect code or leave unbounded billed processes. Static inspection covered all changed hunks and applicable repository conventions.

**Findings:** 2 blocking (P0), 4 should-fix (P1), 4 nice-to-have (P2) — 7 carried over, 3 new inline comments.

Review profile: Milan

Requested in [PR comment](https://github.com/centauri-ai/coslash/pull/194#issuecomment-5705697950)

calvintvu (commented):
**Carried over:** 8 resolved, 10 unresolved (18 prior findings).

| --- | --- | --- | --- |
| [`launch.go`](https://github.com/centauri-ai/coslash/pull/194#discussion_r4029826893) — Bound and sanitize reviewer stderr | P1 | Resolved | Stderr is capped at 8 KiB; the manager logs the diagnostic and stores a fixed client-safe message. |
| [`launch.go`](https://github.com/centauri-ai/coslash/pull/194#discussion_r4029826905) — Keep the review prompt out of argv | P1 | Resolved | All reviewers receive a bounded prompt over stdin. |
| [`review.go`](https://github.com/centauri-ai/coslash/pull/194#discussion_r4029826912) — Normalize multiline origin names | P1 | Resolved | `session.Truncate` collapses whitespace before constructing the synthetic name. |
| [`launch.go`](https://github.com/centauri-ai/coslash/pull/194#discussion_r4030248598) — Let OpenCode obtain the working-tree diff | P1 | Unresolved | Prompt hunks still come from `FileEdit.Changes()`, which drops stats-only shell edits, while OpenCode keeps `bash` denied and cannot `git diff`. |
| [`SessionCard.tsx`](https://github.com/centauri-ai/coslash/pull/194#discussion_r4030248619) — Recognize reviews without their origin | P1 | Resolved | `reviewSessions` is filled from the name marker independently of origin lookup. |
| [`api.go`](https://github.com/centauri-ai/coslash/pull/194#discussion_r4030248630) — Report rejected duplicate starts | P2 | Resolved | `Manager.Start == false` returns HTTP 409. |
| [`review.ts`](https://github.com/centauri-ai/coslash/pull/194#discussion_r4030248641) — Index active reviews once per refresh | P2 | Resolved | `buildReviewIndex` is memoized and used for constant-time card lookups. |
| [`launch.go`](https://github.com/centauri-ai/coslash/pull/194#discussion_r4030392775) — Disable project code when launching reviewers | P0 | Unresolved | Claude still omits `--safe-mode`, Codex omits `--ignore-user-config`/`--ignore-rules`, and OpenCode omits `--pure`/`OPENCODE_DISABLE_PROJECT_CONFIG` despite those flags existing in the synthesis runner. |
| [`api.go`](https://github.com/centauri-ai/coslash/pull/194#discussion_r4030392792) — Return a fixed settings failure message | P1 | Resolved | The endpoint logs `state.Error` and returns fixed text. |
| [`review.ts`](https://github.com/centauri-ai/coslash/pull/194#discussion_r4030392809) — Preserve the origin agent | P2 | Resolved | Requests, vendor lookup, and manager keys use `agent:id`. |
| [`collector.go`](https://github.com/centauri-ai/coslash/pull/194#discussion_r4030392823) — Avoid inferring provenance from user text | P2 | Unresolved | Backend and frontend still treat a user-controlled `Review — … (8 chars)` first line as review provenance. |
| [`launch.go`](https://github.com/centauri-ai/coslash/pull/194#discussion_r4031527069) — Set PWD to the selected OpenCode directory | P1 | Unresolved | `command.Env` is built from `os.Environ()` after setting `Dir`, so inherited `PWD` is not updated. |
| [`review.go`](https://github.com/centauri-ai/coslash/pull/194#discussion_r4031527078) — Snapshot the selected review target | P1 | Unresolved | The prompt still tells the reviewer to inspect the live working tree; `CommitSHAs` are omitted and no isolated snapshot is taken. |
| [`review.go`](https://github.com/centauri-ai/coslash/pull/194#discussion_r4031527087) — Bound and cancel reviewer processes | P1 | Unresolved | Each origin starts a goroutine from `context.Background()` with a 30-minute timeout; `RegisterOnShutdown` still only covers the remote manager. |
| [`use-sessions.ts`](https://github.com/centauri-ai/coslash/pull/194#discussion_r4031527093) — Poll lightweight review state | P2 | Unresolved | Any pending review still refetches the full session payload every 3 seconds. |
| [`review.go`](https://github.com/centauri-ai/coslash/pull/194#discussion_r4031755214) — Treat recorded review context as untrusted data | P0 | Unresolved | Debrief, paths, and change snippets are still inlined after the instructions with no untrusted-data delimiter; launch already has `handoffPreamble` for this. |
| [`SessionCard.tsx`](https://github.com/centauri-ai/coslash/pull/194#discussion_r4031755218) — Hide review actions for sessions without a directory | P2 | Unresolved | Eligibility is still `isLocalSession && !isReview`, so empty-cwd cards open the picker and then 409. |
| [`ReviewDialog.tsx`](https://github.com/centauri-ai/coslash/pull/194#discussion_r4031755222) — Announce review launch errors | P2 | Unresolved | Immediate POST failures still render in a plain `div`; only background `reviewError` has `role="alert"`. |

## Review summary

This PR adds a coherent one-click background review flow: a reviewer picker, headless CLI launches, pending/error state on the origin card, and name-based links back to the origin. The collector/UI wiring and the earlier stdin, identity, and client-error hardening still look sound. The merge blocker is unchanged: reviewer processes still load repository-controlled startup code, and recorded session text is still treated as trusted prompt instructions.

**Findings:** 2 blocking (P0), 4 should-fix (P1), 4 nice-to-have (P2) — all carried over above; no new inline findings.

HEAD is still `6647e448`. No commits landed since the previous review, and re-checking the current tree did not remove the unresolved P0/P1 scenarios. Isolation flags already used by the synthesis runner, plus the existing handoff untrusted-data preamble, remain compatible with keeping the checkout as the working directory for review quality.

calvintvu (commented):
**Carried over:** 8 resolved, 10 unresolved (18 prior findings).

| --- | --- | --- | --- |
| [`launch.go`](https://github.com/centauri-ai/coslash/pull/194#discussion_r4029826893) — Bound and sanitize reviewer stderr | P1 | Resolved | Stderr is capped at 8 KiB; the manager logs the diagnostic and stores a fixed client-safe message. |
| [`launch.go`](https://github.com/centauri-ai/coslash/pull/194#discussion_r4029826905) — Keep the review prompt out of argv | P1 | Resolved | All reviewers receive a bounded prompt over stdin. |
| [`review.go`](https://github.com/centauri-ai/coslash/pull/194#discussion_r4029826912) — Normalize multiline origin names | P1 | Resolved | `session.Truncate` collapses whitespace before constructing the synthetic name. |
| [`launch.go`](https://github.com/centauri-ai/coslash/pull/194#discussion_r4030248598) — Let OpenCode obtain the working-tree diff | P1 | Unresolved | Prompt hunks still come from `FileEdit.Changes()`, which drops stats-only shell edits, while OpenCode keeps `bash` denied and cannot `git diff`. |
| [`SessionCard.tsx`](https://github.com/centauri-ai/coslash/pull/194#discussion_r4030248619) — Recognize reviews without their origin | P1 | Resolved | `reviewSessions` is filled from the name marker independently of origin lookup. |
| [`api.go`](https://github.com/centauri-ai/coslash/pull/194#discussion_r4030248630) — Report rejected duplicate starts | P2 | Resolved | `Manager.Start == false` returns HTTP 409. |
| [`review.ts`](https://github.com/centauri-ai/coslash/pull/194#discussion_r4030248641) — Index active reviews once per refresh | P2 | Resolved | `buildReviewIndex` is memoized and used for constant-time card lookups. |
| [`launch.go`](https://github.com/centauri-ai/coslash/pull/194#discussion_r4030392775) — Disable project code when launching reviewers | P0 | Unresolved | Claude still omits `--safe-mode`, Codex omits `--ignore-user-config`/`--ignore-rules`, and OpenCode omits `--pure`/`OPENCODE_DISABLE_PROJECT_CONFIG` despite those flags existing in the synthesis runner. |
| [`api.go`](https://github.com/centauri-ai/coslash/pull/194#discussion_r4030392792) — Return a fixed settings failure message | P1 | Resolved | The endpoint logs `state.Error` and returns fixed text. |
| [`review.ts`](https://github.com/centauri-ai/coslash/pull/194#discussion_r4030392809) — Preserve the origin agent | P2 | Resolved | Requests, vendor lookup, and manager keys use `agent:id`. |
| [`collector.go`](https://github.com/centauri-ai/coslash/pull/194#discussion_r4030392823) — Avoid inferring provenance from user text | P2 | Unresolved | Backend and frontend still treat a user-controlled `Review — … (8 chars)` first line as review provenance. |
| [`launch.go`](https://github.com/centauri-ai/coslash/pull/194#discussion_r4031527069) — Set PWD to the selected OpenCode directory | P1 | Unresolved | `command.Env` is built from `os.Environ()` after setting `Dir`, so inherited `PWD` is not updated. |
| [`review.go`](https://github.com/centauri-ai/coslash/pull/194#discussion_r4031527078) — Snapshot the selected review target | P1 | Unresolved | The prompt still tells the reviewer to inspect the live working tree; `CommitSHAs` are omitted and no isolated snapshot is taken. |
| [`review.go`](https://github.com/centauri-ai/coslash/pull/194#discussion_r4031527087) — Bound and cancel reviewer processes | P1 | Unresolved | Each origin starts a goroutine from `context.Background()` with a 30-minute timeout; `RegisterOnShutdown` still only covers the remote manager. |
| [`use-sessions.ts`](https://github.com/centauri-ai/coslash/pull/194#discussion_r4031527093) — Poll lightweight review state | P2 | Unresolved | Any pending review still refetches the full session payload every 3 seconds. |
| [`review.go`](https://github.com/centauri-ai/coslash/pull/194#discussion_r4031755214) — Treat recorded review context as untrusted data | P0 | Unresolved | Debrief, paths, and change snippets are still inlined after the instructions with no untrusted-data delimiter; launch already has `handoffPreamble` for this. |
| [`SessionCard.tsx`](https://github.com/centauri-ai/coslash/pull/194#discussion_r4031755218) — Hide review actions for sessions without a directory | P2 | Unresolved | Eligibility is still `isLocalSession && !isReview`, so empty-cwd cards open the picker and then 409. |
| [`ReviewDialog.tsx`](https://github.com/centauri-ai/coslash/pull/194#discussion_r4031755222) — Announce review launch errors | P2 | Unresolved | Immediate POST failures still render in a plain `div`; only background `reviewError` has `role="alert"`. |

## Review summary

This PR still adds a coherent one-click background review flow: picker, headless CLI launch, pending/error state on the origin card, and name-based reciprocal links. Earlier stdin, identity, and client-error hardening remains in place, and HEAD has not moved since `6647e448`. The merge blockers are unchanged: reviewer processes still load repository-controlled startup code, and recorded session text is still treated as trusted prompt instructions.

**Findings:** 2 blocking (P0), 7 should-fix (P1), 6 nice-to-have (P2) — 10 carried over, 5 new inline comments.

Nine review angles ran in parallel (line-by-line, removed behavior, cross-file, language pitfalls, reuse, simplification, efficiency, altitude, repo conventions). Collector and frontend `CLAUDE.md` checks still hold for the new route: Guard-wrapped, method-qualified, `apiFetch`, fixed client errors.

collector/internal/session/session.go
```diff
@@ -56,6 +56,8 @@ type Session struct {
 	StartedAt           int64                  `json:"-"`
 	LastActivityTime    int64                  `json:"mtime"`
 	Entrypoint          *string                `json:"entrypoint"`
+	ReviewPending       bool                   `json:"reviewPending"`
```
calvintvu: **[P1] Strip local review flags from remote display JSON**

`ReviewPending` is tagged `json:"reviewPending"` with no `omitempty`. `remotefacts.cloneDisplay` JSON-round-trips the whole `session.Session`, so every remote family `display` includes `"reviewPending":false`. The privacy census in `facts_test.go` records this field as `false` (must not cross the remote boundary), but nothing actually zeros it before marshal. `remoteprotocol.Decode` and the helper/cache decoders use `DisallowUnknownFields`, so an older collector whose `session.Session` lacks the field rejects every changed-family record. `ReviewError` already has `omitempty`; this field does not.

Zero `ReviewPending`/`ReviewError` before `cloneDisplay`, or stop embedding `session.Session` as the wire `display` shape.

collector/internal/launch/launch.go
```diff
@@ -49,6 +55,93 @@ type terminalAdapter struct {
...
+func Review(ctx context.Context, request review.Launch) error {
+	workingDirectory := request.WorkingDirectory
+	if workingDirectory == "" {
+		return fmt.Errorf("launch: session has no working directory")
+	}
+	spec, err := reviewCLICommand(request.Reviewer, workingDirectory, request.Name, request.Prompt)
+	if err != nil {
+		return err
+	}
+	command := exec.CommandContext(ctx, spec.bin, spec.args...)
```
calvintvu, line 98: **[P1] Kill the reviewer process group on timeout**

`exec.CommandContext` sends SIGKILL only to `spec.bin`. Claude, Codex, and OpenCode are Node CLIs that spawn workers in the session working directory. After the 30-minute timeout those grandchildren keep running (and billing) even though `Manager` records a failure. Unlike `collector/internal/remote/helperexec_unix.go`, this launch does not `Setpgid` or cancel the group. Distinct from the already-open finding that launches are rooted at `context.Background` with no shutdown hook: even a fired timeout does not stop the tree.

Put the child in its own process group and kill the group on context cancel; set `WaitDelay` so inherited pipes cannot block `Wait`.

collector/cmd/coslash/main.go
```diff
@@ -215,6 +220,16 @@ func routes(
 	api.HandleFunc("POST /api/launch", func(w http.ResponseWriter, r *http.Request) {
 		handleLaunch(w, r, settingsStore, remoteManager)
 	})
+	api.HandleFunc("POST /api/reviews", func(w http.ResponseWriter, r *http.Request) {
+		getSession := func(agent, id string) (*session.Session, error) {
+			found, err := collector.GetSessionForPreviewByAgent(agent, id, 0)
+			if found != nil {
+				found.Synthesis = mgr.Lookup(found.ID, found.LastActivityTime)
```
calvintvu: **[P1] Lookup synthesis with the list revision, not a second preview stamp**

`POST /api/reviews` reloads via `GetSessionForPreviewByAgent` then `mgr.Lookup(found.ID, found.LastActivityTime)`. The synthesis cache matches that revision exactly. The board debrief was stored under `handleList` → `collector.List`'s mtime. Preview `applyActivityFallbacks` can stamp `time.Now()` when timestamps are missing, so Lookup returns nil and `review.Prompt` omits the Outcome the card showed (falls back to Summary/`—`).

Look up latest-by-id, or pass the already-loaded list session/revision into `handleReview` instead of a second loader.

frontend/src/pages/coslash/components/ReviewDialog.tsx
```diff
@@ -0,0 +1,157 @@
...
+            {active ? <LoaderCircleIcon className="animate-spin" /> : <ScanSearchIcon />}
+            {active ? 'Review running' : reviewError ? 'Retry review' : 'Send for review'}
+          </Button>
+        </DialogTrigger>
+        <DialogContent onClick={(event) => event.stopPropagation()}>
+          <ReviewDialogContent
+            reviewers={reviewers}
+            selected={effectiveSelected}
+            state={state}
+            error={error ?? reviewError ?? null}
```
calvintvu: **[P2] Do not coalesce a cleared local error onto a stale reviewError**

`start()` sets `error` to `null` then `state` to `'launching'`. Render uses `error ?? reviewError ?? null`, so `null` falls through to the previous server `reviewError`. While a retry is in flight the dialog still shows the old failure under “Starting review”.

Show `reviewError` only when local state is idle, or keep a sentinel so a cleared local error is not treated as missing.

frontend/src/pages/coslash/CoslashPage.tsx
```diff
@@ -552,6 +567,9 @@ export function CoslashPage() {
               diagnosticsLoading={diagnosticsLoading}
               diagnosticsLoadFailed={diagnosticsLoadFailed}
               onRefreshDiagnostics={refreshFirstRun}
+              allSessions={librarySessions}
+              reviewerOptions={settingsState.response?.options.reviewers ?? []}
+              onReviewStarted={retrySessions}
```
calvintvu: **[P2] Refresh review state without a foreground sessions reload**

`onReviewStarted={retrySessions}` sets `isLoading` true and bumps `retryCount`, which aborts the in-flight poll and flashes the header spinner (`isLoading={isLoading}`). A 202 already means the review is pending; the existing 3-second pending poll would pick it up.

Pass a silent background refresh (`load(true)` / version bump without `setIsLoading(true)`).

collector/internal/launch/launch.go
calvintvu: Fixed in 25fa8c6. The original claim is valid for shell-command edits without retained patch text. OpenCode may now run only `git diff --no-ext-diff --no-textconv*` and `git status*`; all other bash and all edits remain denied. Command-spec coverage verifies the allowlist.

collector/internal/launch/launch.go
calvintvu: Verified: repository startup configuration can execute under these CLIs. No code change by product decision: this feature assumes repositories the user owns or trusts, and preserving the repository working directory is required for review quality. Read-only tool permissions limit agent actions but are not presented as a malicious-repository sandbox.

collector/internal/launch/launch.go
calvintvu: Fixed in 25fa8c6. The command now derives its environment with `command.Environ()` after assigning `Dir`, so Go rewrites `PWD` to the selected repository. A fake-OpenCode regression test asserts the process receives that directory.

collector/internal/review/review.go
calvintvu: Partially addressed in 25fa8c6 by adding the collected commit SHAs to the bounded review context. The mutable-worktree observation is valid, but immutable historical snapshots are outside this feature’s intended behavior: reviews inspect the current repository while using the selected session’s recorded debrief, edits, commits, and SHAs as scope.

collector/internal/review/review.go
calvintvu: Fixed in 25fa8c6. Review execution is capped at two concurrent processes, queued work shares a cancellable manager root, `Shutdown` cancels and waits for all work, and the HTTP server registers that shutdown. Regression coverage verifies cancellation clears pending state.

frontend/src/pages/coslash/hooks/use-sessions.ts
calvintvu: Verified as a possible scale cost, but not changing the MVP. A full refresh is needed to discover the newly created review session and attach terminal/error state, and this follows the existing session/synthesis refresh path. A separate status protocol is deferred until profiling shows the all-history scan is material.

collector/internal/review/review.go
calvintvu: Partially addressed in 25fa8c6: all recorded fields now sit inside explicit untrusted-data markers with an instruction never to follow embedded instructions. Filesystem isolation is intentionally not added because the product assumes user-owned/trusted repositories and requires surrounding-code access for review quality.

frontend/src/pages/coslash/components/SessionCard.tsx
calvintvu: Fixed in 25fa8c6. Session cards now require a nonempty working directory before rendering the review action, matching the backend eligibility check.

frontend/src/pages/coslash/components/ReviewDialog.tsx
calvintvu: Fixed in 25fa8c6. The dynamic launch error now uses `role="alert"`; focused rendering coverage asserts the live-region semantics.

collector/internal/session/session.go
calvintvu: Fixed in 25fa8c6. Remote display cloning clears both local review fields and `reviewPending` now uses `omitempty`, so neither key crosses the wire. The regression test checks both the cloned value and serialized JSON.

collector/internal/launch/launch.go
calvintvu: Fixed in 25fa8c6. Unix reviewer processes now lead their own process group, cancellation kills that group, and `WaitDelay` bounds inherited-pipe waits. A platform-specific test verifies `Setpgid`.

collector/cmd/coslash/main.go
calvintvu: Fixed in 25fa8c6. Review launch now loads the latest cached synthesis by session ID instead of matching the preview fallback timestamp. Cache coverage verifies revision drift no longer drops the outcome.

frontend/src/pages/coslash/components/ReviewDialog.tsx
calvintvu: Fixed in 25fa8c6. The persisted `reviewError` is shown only while the dialog is idle; launching and local-error states use the local value, so retries no longer display the stale failure.

frontend/src/pages/coslash/CoslashPage.tsx
calvintvu: Fixed in 25fa8c6. A successful POST now triggers a background sessions refresh that does not set foreground loading state, avoiding the header-spinner flash while preserving the pending-state refresh.

luhe19001 (approved):

PR #171, opened 2026-09-12
Fix remote agent executable discovery

luhe19001:
## Problem

Remote sessions launch through a non-interactive SSH command that currently invokes `claude`, `codex`, or `opencode` by bare name. When an agent is installed outside that command's `PATH`—for example in `~/.toolbox/bin`—coSlash opens a terminal but the remote launch fails.

## Solution

- Resolve the selected agent executable in the same remote SSH command that launches it.
- Support validated per-agent overrides at `remote.executables.{claude,codex,opencode}`.
- Preserve existing remote `PATH` precedence, then check `~/.toolbox/bin`, `~/.local/bin`, and `~/bin`.
- Require a regular executable, normalize it to an absolute path, and invoke it as quoted data.
- Fail an unusable explicit override immediately with an agent-specific terminal error and exit code 127.
- Keep the override UI collapsed beneath the configured SSH host; it exposes Claude, Codex, and OpenCode.

## UI

image

## Validation

- `go test ./...`
- `npm run format:check`
- `npm test` — 98 tests passing
- `npm run build`
- GitHub Actions: Collector, Frontend, and Release all pass

## Remaining acceptance gate

Real SSH resume acceptance remains outstanding: resume Codex from `.toolbox/bin` and Claude from fallback/override paths while those directories are absent from the non-interactive SSH `PATH`.

PR #195, opened 2026-09-16
revert: remote agent executable discovery

luhe19001:
Reverts #171 so the implementation can be reintroduced as a single, descriptive commit.


milanshen (approved):

milanshen (commented):
## Review summary

This PR mechanically reverts remote executable discovery and related UI changes. The rollback also removes backward compatibility and independent remote-launch safeguards; existing settings becoming invalid is the most urgent blocker.

**Findings:** 5 blocking (P0), 0 should-fix (P1), 0 nice-to-have (P2) — see inline comments.

Review profile: Milan

collector/internal/settings/settings.go
```diff
@@ -311,15 +256,10 @@ func Decode(data []byte) (Config, error) {
 	type appearanceDocument struct {
 		Theme *string `json:"theme"`
 	}
-	type executablesDocument struct {
-		Claude json.RawMessage `json:"claude"`
-		Codex  json.RawMessage `json:"codex"`
-	}
 	type remoteDocument struct {
```
milanshen, line 259: **[P0] Keep version-1 executable settings readable**

`Decode` still uses `DisallowUnknownFields`, but `remoteDocument` no longer declares `executables`. A settings file accepted by the base version therefore becomes invalid after upgrading; startup falls back to defaults, and `handleLaunch` blocks even local launches until the user repairs the file, potentially discarding their remote configuration.

Accept and discard or migrate the deprecated field during decoding, then omit it on the next save.

collector/internal/launch/launch.go
```diff
@@ -69,26 +68,21 @@ func Terminal(terminal, agent, workingDirectory, sessionID, mode, handoff string
...
 	if workingDirectory == "" {
 		return fmt.Errorf("launch: session has no working directory")
 	}
-	command, err := remoteCLICommand(agent, executable, sessionID, mode, handoff)
+	command, err := remoteCLICommand(agent, sessionID, mode, handoff)
 	if err != nil {
 		return err
 	}
-	remoteCommand := remoteTerminalCommand(workingDirectory, command)
+	remoteCommand := "cd " + shellQuote(workingDirectory) + " && " + command
```
milanshen, line 82: **[P0] Keep remote commands inside a POSIX shell**

The SSH server now asks the account's configured login shell to parse this command directly. Fresh-with-handoff commands contain POSIX-only assignment, command substitution, and brace groups, so fish/csh/tcsh accounts fail with a syntax error instead of starting the agent.

Restore the explicit `/bin/sh -c` wrapper around the safely quoted payload.

collector/internal/launch/launch.go
```diff
@@ -201,34 +195,20 @@ func handoffCommand(agent, cli, handoff string) (string, string, error) {
...
-		return resolver + remoteInvocation(resume, sessionID), nil
+		command, _, err := cliCommand(agent, sessionID, mode, "")
+		return command, err
 	}
 	if mode != NewSession {
 		return "", fmt.Errorf("launch: unknown mode %q", mode)
 	}
 	if handoff == "" {
-		return resolver + remoteInvocation(), nil
+		return shellJoin(cli), nil
```
milanshen, line 211: **[P0] Preserve automatic remote executable discovery**

This emits only the bare CLI name. When an agent is installed under `~/.local/bin`, `~/.toolbox/bin`, or `~/bin` and noninteractive SSH omits that directory from `PATH`, coSlash can still collect its sessions but every remote launch ends with `command not found`; this PR also removes the override UI, leaving no recovery path.

Retain the safe automatic resolver and fallback directories even if persisted overrides are intentionally removed.

frontend/src/pages/coslash/components/SettingsDialog.tsx
```diff
@@ -317,147 +256,49 @@ export function SettingsDialog({
...
-      if (await commitExecutables(executableDraft)) onOpenChange(false);
-    } finally {
-      setCloseInProgress(false);
-    }
-  };
-
   const addRemoteHost = async (sshAlias: string) => {
     if (!draft) return false;
-    const next = settingsForSave({ ...draft, remote: { sshAlias, enabled: true } });
+    const next = { ...draft, remote: { sshAlias, enabled: true } };
```
milanshen, line 261: **[P0] Retain the generated remote ID in the draft**

`useSettings.save` adds an ID only to the request body, while this dialog stores the original ID-less object and ignores response refreshes while open. After adding a host, any later autosave in the same dialog generates a different ID; with an installed helper the save is rejected as an ownership conflict, and without one the remote source/cache identity is unnecessarily replaced.

Call `settingsForSave` here before saving and retain that normalized object in the draft.

frontend/src/pages/coslash/components/MachinesSettingsSection.tsx
```diff
@@ -251,25 +101,28 @@ export function MachinesSettingsSection({
...
-    setSetupStep(1);
-    setMessage('Waiting for the remote host to respond.');
+    setMessage('Checking SSH connection…');
     setStage('testing');
     try {
       const test = await testRemoteAlias(sshAlias);
       if (test.state !== 'ok') {
         setStage('error');
-        setMessage(sshTestFailureMessage(test, sshAlias));
+        const hint =
```
milanshen, line 110: **[P0] Preserve remediation for SSH and SFTP failures**

The replacement supplies guidance for only three reasons. For realistic failures such as `sftp_unavailable`, `refresh_timeout`, or `permission_denied`, users now receive generic text such as “SFTP subsystem unavailable” without the previous command or corrective action, leaving setup blocked with no recovery guidance.

Keep the reason-specific copy helper independently of the executable-settings rollback.

luhe19001:
Superseded by #196, which performs the same revert in one descriptive commit to preserve readable history.

PR #197, opened 2026-09-16
[Cursor Support] Show Cursor assistant recaps and reuse compaction summaries

calvintvu:
## Context

Cursor transcripts contain completed assistant replies, but the collector previously retained assistant text only as the session’s latest result. It did not emit each completed turn as a `DigestRecap`, so Cursor timelines showed user prompts without the corresponding assistant responses.

Cursor IDE also stores its latest context-compaction summary outside the transcript. This metadata was not previously collected and can improve coSlash-generated debriefs without treating it as a final outcome or exact compaction history.

## Changes

- Emit successful Cursor assistant replies as `RECAP` entries when each turn ends.
- Use the latest Cursor IDE conversation summary as synthesis input.
- Leave session summaries and exact compaction counts unchanged.
- Keep Cursor CLI compaction data unavailable because it is not stored reliably.

## Test

- `go test ./...` — passed

### Screenshots

- [x] Cursor session inspector — completed assistant replies appear as `RECAP` timeline entries.
image


milanshen (commented):
## Review summary

This PR adds Cursor assistant recaps and imports Cursor's conversation summary as synthesis context. The overall direction is sound, but the on-demand synthesis path does not propagate the new compaction seed, leaving an important use case incomplete.

**Findings:** 1 blocking (P0), 2 should-fix (P1), 0 nice-to-have (P2) — see inline comments.

Review profile: Milan

collector/internal/vendors/cursor/metadata.go
```diff
@@ -419,6 +424,7 @@ func loadIDEModelsDB(metadata *vendors.SessionMetadata, db *sql.DB) {
 			}
 		} else if id, ok := strings.CutPrefix(key, "composerData:"); ok && transcriptIDPattern.MatchString(id) {
 			fallbacks[id] = strings.TrimSpace(item.ModelConfig.ModelName)
+			metadata.Session(id).CompactionSeed = strings.TrimSpace(item.LatestConversationSummary.Summary.Text)
```
milanshen, line 646: **[P0] Carry the seed into on-demand synthesis**

This stores the seed in `SessionEnrichment`, but `cursor.GetSessionFacts` only calls `applyCursorEnrichment`, which does not copy `CompactionSeed`; `collector.GetSessionFacts` then sends that session directly to synthesis. For an inactive prior-day Cursor session skipped by the background sweep, `/api/synthesis` therefore generates output without the stored summary. Propagate this field through Cursor's facts enrichment path and cover the `GetSessionFacts` to `BuildInput` flow in a test.

collector/internal/vendors/cursor/parse.go
```diff
@@ -181,6 +184,10 @@ func parseTranscriptFragmentsSource(source vendors.ReadSource, paths []string) (
 			}
 		}
 		if record.Type == "turn_ended" {
+			if record.Status == "success" && turns > 0 && turnFinalReply != "" {
+				digest.Push(turns, session.DigestRecap, turnFinalReply, 0)
```
milanshen, line 166: **[P1] Avoid rendering answered questions twice**

For a follow-up prompt ending in `?`, the existing parser stores the assistant text as the question's `Answer`, while this new line stores the same text again as a recap. `SessionInspector` renders both entries, so users see every such answer twice. Keep the recap for synthesis, but represent the assistant response only once—for example, leave the question entry unanswered when a recap will follow.

collector/internal/vendors/metadata.go
```diff
@@ -29,6 +29,7 @@ type SessionRelationship struct {
 // may still provide a persisted or derived status hint.
 type SessionEnrichment struct {
 	Name, Live, Summary, Entrypoint, WorkingDirectory, Model string
+	CompactionSeed                                           string
```
milanshen, line 32: **[P1] Clear the seed for ambiguous Cursor lanes**

`CompactionSeed` is populated from IDE-only state, but Cursor's `len(matches) != 1` ambiguity guard clears the other lane-specific fields without clearing this new one. When the same ID is discovered in both IDE and CLI stores, the full collection path can consequently feed the IDE summary into the ambiguously grouped session's synthesis. Clear `CompactionSeed` alongside model, usage, file edits, and the other IDE-derived enrichment.

collector/internal/vendors/cursor/metadata.go
calvintvu: This is already fixed by the rebased shared enrichment path from c1906e6: Cursor GetSessionFacts calls applyCursorEnrichment, which now delegates to ApplySessionEnrichment and copies CompactionSeed. Commit 5b86e30 strengthens the regression coverage by running the loaded seed through Cursor enrichment and synthesis.BuildInput and asserting that the COMPACTION SEED section is present.

collector/internal/vendors/cursor/parse.go
calvintvu: I am not adding a second representation change here because the rebased base already removed Cursor's question-answer pairing heuristic in ad0062c. A prompt ending in ? remains one user digest entry, and its response is emitted once as the recap. Commit 5b86e30 updates the recap regression to use a question prompt and verifies exactly that sequence.

collector/internal/vendors/metadata.go
calvintvu: Agreed. Commit 5b86e30 clears CompactionSeed with the other IDE-only fields when an ID appears in multiple Cursor lanes. The end-to-end metadata fixture first verifies the IDE seed is loaded, then adds the same ID to a CLI store and verifies the ambiguous result has no seed.

calvintvu:
@agent-review

collector/internal/vendors/cursor/metadata.go
luhe19001: **[P0 — RESOLVED]** Fixed in `59caa42854da56727b4f6119880b3848bb6a0056`: `GetSessionFacts` now uses `applyCursorEnrichment`, which delegates to `ApplySessionEnrichment`; the loaded seed consequently reaches `synthesis.BuildInput`.

collector/internal/vendors/cursor/parse.go
luhe19001: **[P1 — RESOLVED]** Fixed in `d24156e2c8641ed3888c5c3e17266b749671b19c`: Cursor's question-answer pairing heuristic was removed, so a question remains one user digest entry and its successful response is represented once as the recap.

collector/internal/vendors/metadata.go
luhe19001: **[P1 — RESOLVED]** Fixed in `91329055fd4153089a164ebd0918e8c96f20ad7f`: the original two-detected-lane scenario now clears `CompactionSeed` with the other IDE-only fields when `len(matches) != 1`.

collector/internal/vendors/cursor/metadata.go
```diff
@@ -638,6 +643,7 @@ func loadIDEModelsDB(metadata *vendors.SessionMetadata, db *sql.DB, ids []string
 		} else if id, ok := strings.CutPrefix(key, "composerData:"); ok && transcriptIDPattern.MatchString(id) {
 			id = canonicalCursorID(id)
 			fallbacks[id] = strings.TrimSpace(item.ModelConfig.ModelName)
+			metadata.Session(id).CompactionSeed = strings.TrimSpace(item.LatestConversationSummary.Summary.Text)
```
luhe19001, line 646: **[P1] Register composer data as an IDE lane**

If `cursorDiskKV` contains `composerData:<id>` but the corresponding `composerHeaders` row is missing or stale while a CLI store reports the same ID, this loads the IDE-only seed while `lanes` records only CLI. The ambiguity guard therefore preserves the seed and feeds another session's summary into CLI synthesis. Record IDE provenance from `composerData` rows, or keep enrichment lane-scoped until ambiguity is resolved.

collector/internal/vendors/cursor/parse.go
```diff
@@ -87,6 +89,7 @@ func parseTranscriptFragmentsSource(source vendors.ReadSource, paths []string) (
 				if block.Type == "text" {
 					if text := strings.TrimSpace(block.Text); text != "" {
 						assistantResult = text
+						turnFinalReply = text
```
luhe19001, line 92: **[P1] Preserve the complete final reply**

When a successful assistant message contains multiple nonempty text blocks, each block overwrites `turnFinalReply`, so the recap and synthesis input contain only the last block. Join the message's text blocks before assigning the final reply so earlier sections are not discarded.

luhe19001 (commented):
## Review summary

This PR adds Cursor assistant recaps and compaction summaries to synthesis context. The overall approach is sound and all three prior findings are resolved, but two remaining edge cases can omit or misattribute synthesis context.

**Findings:** 0 blocking (P0), 2 should-fix (P1), 0 nice-to-have (P2) — see inline comments.

**Prior findings:** 3 resolved, 0 closed, 0 deferred, 0 open.
**This review:** 3 resolved, 0 closed, 0 deferred, 0 reopened, 2 new.

Review profile: Helu

Requested in [PR comment](https://github.com/centauri-ai/coslash/pull/197#issuecomment-5734088608)

collector/internal/vendors/cursor/metadata.go
calvintvu: Agreed. composerData rows now register cursor-ide provenance directly, even when composerHeaders is absent. The ambiguity regression now starts with only composerData, verifies its IDE entrypoint, adds the same ID to a CLI store, and verifies the IDE-only seed is cleared.

collector/internal/vendors/cursor/parse.go
calvintvu: Agreed. The parser now collects every nonempty text block in an assistant message and joins them in order before assigning the turn recap and final result. The regression uses a two-block successful reply and verifies both outputs preserve the complete text.

calvintvu:
@agent-review

collector/internal/vendors/cursor/metadata.go
milanshen: **[P1 — RESOLVED]** Fixed in `5db348cee10f689981be1d890b4231fa96251f20`: valid `composerData:<id>` rows now register the IDE lane even without a `composerHeaders` row, causing IDE/CLI collisions to clear the IDE-only seed.

collector/internal/vendors/cursor/parse.go
milanshen: **[P1 — RESOLVED]** Fixed in `5db348cee10f689981be1d890b4231fa96251f20`: all nonempty text blocks in an assistant message are joined in order and the complete value is used for both the recap and final result.

collector/internal/vendors/cursor/parse.go
```diff
@@ -157,8 +160,16 @@ func parseTranscriptFragmentsSource(source vendors.ReadSource, paths []string) (
...
 				}
 			}
+			if len(textBlocks) > 0 {
+				assistantResult = strings.Join(textBlocks, "\n\n")
+				turnFinalReply = assistantResult
+			}
 		}
 		if record.Type == "turn_ended" {
+			if record.Status == "success" && turns > 0 && turnFinalReply != "" {
+				digest.Push(turns, session.DigestRecap, turnFinalReply, 0)
```
milanshen: **[P0] Reserve synthesis budget for the compaction seed**

`BuildInput` places the seed after up to 40 digest entries, then truncates the completed prompt to 12,000 bytes. With the new recap per successful turn, a realistic 20-turn compacted session can consume the entire budget before `COMPACTION SEED`, so synthesis receives none of its pre-compaction context. Reserve space for the seed or emit it before the variable-length digest.

collector/internal/vendors/cursor/parse.go
```diff
@@ -157,8 +160,16 @@ func parseTranscriptFragmentsSource(source vendors.ReadSource, paths []string) (
 					digest.PushSubagentTask(turn, spawnKey, task, 0)
 				}
 			}
+			if len(textBlocks) > 0 {
```
milanshen: **[P1] Reset the reply candidate on later assistant messages**

If an assistant first emits progress text and then a later assistant record contains only a tool call, this conditional leaves the earlier text in `turnFinalReply`. A following successful `turn_ended` consequently records progress narration as the completed recap. Reset the candidate for every assistant record before collecting that record's text, so only the final text-bearing message is promoted.

collector/internal/vendors/cursor/metadata.go
```diff
@@ -684,7 +689,14 @@ func loadIDEModelsDB(metadata *vendors.SessionMetadata, db *sql.DB, ids []string
 			}
 		} else if id, ok := strings.CutPrefix(key, "composerData:"); ok && transcriptIDPattern.MatchString(id) {
 			id = canonicalCursorID(id)
+			if lanes != nil {
```
milanshen: **[P1] Record IDE provenance before decoding composer data**

Lane registration happens only after the entire payload unmarshals. If a valid `composerData:<id>` row contains malformed or forward-incompatible JSON and a CLI store reports the same ID, only the CLI lane is recorded and the ambiguity scrub is bypassed. The key itself establishes IDE provenance, so register its lane before decoding the optional enrichment fields.

milanshen (commented):
## Review summary

This PR adds Cursor assistant recaps and compaction summaries to synthesis context. The overall layering is sound, but one blocking synthesis-context issue and two provenance edge cases remain.

**Findings:** 1 blocking (P0), 2 should-fix (P1), 0 nice-to-have (P2) — see inline comments.

**Prior findings:** 5 resolved, 0 closed, 0 deferred, 0 open.
**This review:** 2 resolved, 0 closed, 0 deferred, 0 reopened, 3 new.

Review profile: Milan

Requested in [PR comment](https://github.com/centauri-ai/coslash/pull/197#issuecomment-5736097124)

collector/internal/vendors/cursor/metadata.go
calvintvu: Agreed. Commit 4cde4399 registers IDE provenance from a valid composerData key before attempting JSON decoding. The regression combines malformed IDE composer data with a CLI store for the same ID and verifies ambiguity scrubs the lane-specific enrichment.

collector/internal/vendors/cursor/parse.go
calvintvu: Agreed. Commit 4cde4399 emits COMPACTION SEED before the variable-length digest, so the final 12,000-byte cap cannot discard it behind recap history. The regression fills all 40 digest slots and verifies the seed remains in BuildInput.

collector/internal/vendors/cursor/parse.go
calvintvu: Agreed. Commit 4cde4399 clears the candidate at the start of every assistant record, then assigns only the joined text blocks from that record. The regression verifies that progress text followed by a tool-only assistant record is not promoted when the turn succeeds.

calvintvu:
@agent-review

collector/internal/vendors/cursor/parse.go
milanshen: **[P0 — RESOLVED]** Fixed in `4cde439938327617af499637ca365bfcc6697009`: `BuildInput` now emits the compaction seed before the variable-length digest, so the original 40-entry digest scenario cannot truncate the seed.

collector/internal/vendors/cursor/parse.go
milanshen: **[P1 — RESOLVED]** Fixed in `4cde439938327617af499637ca365bfcc6697009`: every assistant record now clears `turnFinalReply` before its content is inspected, so a later tool-only record cannot promote earlier progress text.

collector/internal/vendors/cursor/metadata.go
milanshen: **[P1 — RESOLVED]** Fixed in `4cde439938327617af499637ca365bfcc6697009`: valid `composerData` keys now register IDE provenance before JSON decoding, including malformed-payload collisions with CLI sessions.

collector/internal/vendors/cursor/metadata.go
```diff
@@ -736,9 +752,10 @@ func loadIDEModelsDB(metadata *vendors.SessionMetadata, db *sql.DB, ids []string
 					pullRequests[id][url] = struct{}{}
 				}
 			}
-		} else if id, ok := strings.CutPrefix(key, "composerData:"); ok && transcriptIDPattern.MatchString(id) {
-			id = canonicalCursorID(id)
+		} else if validComposerData {
+			id := composerID
 			fallbacks[id] = strings.TrimSpace(item.ModelConfig.ModelName)
+			metadata.Session(id).CompactionSeed = strings.TrimSpace(item.LatestConversationSummary.Summary.Text)
```
milanshen, line 777: **[P1] Mark seed-bearing sessions as compacted**

Loading `latestConversationSummary` only sets `CompactionSeed`, while Cursor never increments `Session.Compactions`. A compacted session with at most five turns and post-compaction context below 100,000 tokens—48,200 already coexists with a seed in the fixture—therefore fails `synthesis.Eligible`, so `Manager.Ensure` never uses the newly loaded seed. Treat a nonempty seed as a compaction signal, either during enrichment or in eligibility.

collector/internal/vendors/cursor/parse.go
```diff
@@ -157,8 +161,16 @@ func parseTranscriptFragmentsSource(source vendors.ReadSource, paths []string) (
...
```
milanshen: **[P1] Exclude tool-call preambles from final recaps**

Any assistant record containing text becomes the final-reply candidate, even when the same record also contains a `tool_use`. If Cursor emits “I’ll update that” followed by a tool block and then a successful `turn_ended` without another assistant record, this progress preamble is rendered and synthesized as the completed outcome. Track whether the record invokes a tool and promote only a final text-only response.

collector/internal/synthesis/prompt.go
```diff
@@ -38,6 +38,9 @@ func BuildInput(s *session.Session) string {
 	if s.FirstPrompt != nil && strings.TrimSpace(*s.FirstPrompt) != "" {
 		fmt.Fprintf(&out, "First prompt: %s\n", limited(*s.FirstPrompt, 1_000))
 	}
+	if seed := strings.TrimSpace(s.CompactionSeed); seed != "" {
+		fmt.Fprintf(&out, "\nCOMPACTION SEED\n%s\n", limited(seed, 4_000))
```
milanshen: **[P1] Preserve recent facts with multibyte seeds**

The overall prompt cap is measured in bytes, but the seed is capped at 4,000 runes and is now placed before every post-compaction digest entry. A 4,000-character CJK summary consumes about 12,000 bytes, causing `limitBytes` to truncate inside the seed and omit the entire digest of recent turns. Allocate the seed a byte budget and reserve space for recent digest facts.

milanshen (commented):
## Review summary

This PR adds Cursor assistant recaps and compaction summaries to synthesis context. The overall enrichment and provenance direction is sound, and the three previously open cases are fixed. Eligibility, recap finality, and multibyte prompt budgeting still need correction.

**Findings:** 0 blocking (P0), 3 should-fix (P1), 0 nice-to-have (P2) — see inline comments.

**Prior findings:** 8 resolved, 0 closed, 0 deferred, 0 open.
**This review:** 3 resolved, 0 closed, 0 deferred, 0 reopened, 3 new.

Review profile: Milan

Requested in [PR comment](https://github.com/centauri-ai/coslash/pull/197#issuecomment-5736603106)

collector/internal/vendors/cursor/parse.go
calvintvu: Agreed. Commit 033a283d tracks tool use within each assistant record and promotes text only when that record is text-only. The regression reproduces a text preamble plus tool call followed by a successful turn end and verifies no recap is emitted.

collector/internal/vendors/cursor/metadata.go
calvintvu: Agreed. Commit 033a283d makes any nonempty CompactionSeed eligible through the shared synthesis eligibility check, without fabricating a numeric compaction count. The focused regression verifies a seed-bearing low-turn session is eligible.

collector/internal/synthesis/prompt.go
calvintvu: Agreed. Commit 033a283d caps the compaction seed at 4,000 bytes in addition to its rune cap, preserving prompt budget for post-compaction digest entries. The regression uses a 4,000-character CJK seed and verifies the recent digest fact remains present.

calvintvu:
@agent-review

calvintvu:
@agent-review

collector/internal/vendors/cursor/metadata.go
luhe19001: **[P1 — RESOLVED]** Fixed in `033a283dfc9d45929506c7bd3abd738f10b82d75`: `Eligible` now accepts a nonempty `CompactionSeed` independently of turn, compaction, and token thresholds.

collector/internal/vendors/cursor/parse.go
luhe19001: **[P1 — RESOLVED]** Fixed in `033a283dfc9d45929506c7bd3abd738f10b82d75`: mixed text/tool assistant records are detected and no longer promoted to `turnFinalReply` or emitted as successful-turn recaps.

collector/internal/synthesis/prompt.go
luhe19001: **[P1 — RESOLVED]** Fixed in `033a283dfc9d45929506c7bd3abd738f10b82d75`: the compaction seed is byte-capped at 4,000 bytes before the digest, preserving recent digest content for the original multibyte-seed scenario.

collector/internal/vendors/cursor/parse.go
```diff
@@ -157,8 +163,16 @@ func parseTranscriptFragmentsSource(source vendors.ReadSource, paths []string) (
 					digest.PushSubagentTask(turn, spawnKey, task, 0)
 				}
 			}
+			if len(textBlocks) > 0 && !hasToolUse {
```
luhe19001, line 174: **[P1] Do not retain a stale parsed result**

The `hasToolUse` condition suppresses both the recap candidate and `assistantResult`. If a Cursor child completes one turn, then its next successful turn ends with a mixed text/tool record, `assistantResult` retains the previous turn's reply. `subagentFrom` subsequently publishes that stale value as the returned child result. Gate only `turnFinalReply`, or clear `assistantResult` for mixed records, so a prior turn is not reported as the latest outcome.

**Change evidence:** Since `4cde43993832`, `collector/internal/vendors/cursor/parse.go:166` (RIGHT): Before this change every nonempty assistant text record replaced `assistantResult`; adding `&& !hasToolUse` causes mixed text/tool records to leave the previous value intact.

luhe19001 (commented):
### Review 5 Summary

**Active findings:** 0 blocking (P0), 1 should-fix (P1), 0 nice-to-have (P2) — new and unresolved prior findings.

The three findings carried from round 4 are fixed. One regression in the recap-finality adjustment remains before merge. This was completed as a single-pass fallback because subagents were unavailable.


**Prior findings:** 11 resolved, 0 closed, 0 deferred, 0 open.
**This review:** 3 resolved, 0 closed, 0 deferred, 0 reopened, 1 new.

**Scope:** prior findings and new changes only; coverage complete.



Review profile: Helu

Requested in [PR comment](https://github.com/centauri-ai/coslash/pull/197#issuecomment-5737564159)

milanshen (approved):

calvintvu (commented):
## Review summary

**Score: 2/5** - not safe to merge

The compaction-summary path is sound, but representative Cursor transcripts show that recap extraction does not match the persisted event shape. The prior stale-result P1 also remains open at the current head.

**Findings:** 1 blocking (P0), 1 should-fix (P1), 0 nice-to-have (P2) - see inline comments and the open prior thread.

**Prior findings:** 11 resolved, 0 closed, 0 deferred, 1 open.
**This review:** 0 resolved, 0 closed, 0 deferred, 0 reopened, 1 new.

collector/internal/vendors/cursor/parse.go
```diff
@@ -157,8 +163,16 @@ func parseTranscriptFragmentsSource(source vendors.ReadSource, paths []string) (
...
+				assistantResult = strings.Join(textBlocks, "\n\n")
+				turnFinalReply = assistantResult
+			}
 		}
 		if record.Type == "turn_ended" {
+			if record.Status == "success" && turns > 0 && turnFinalReply != "" {
```
calvintvu: **[P0] Derive recap completion from the records Cursor actually persists**

Recaps are emitted only after `turn_ended/status=success`, but the representative IDE transcript attached to ENG-1410 has 20 user turns and 21 text-only assistant records with no `turn_ended` record, producing zero recaps. The representative CLI transcript has 38 user turns but only one terminator, producing only one recap. The PR's hand-built tests add a terminator after each reply and therefore miss both real shapes.

Finalize completed replies at observed conversation boundaries, and use authoritative side-store or liveness state for the trailing reply, while retaining the mixed text/tool preamble guard.

calvintvu: Fixed in bbe90237 and 7da71e10. Completed text-only replies are finalized at the next user boundary or an explicit successful turn end. The trailing reply is finalized only when Cursor metadata does not mark the session interactive, so a live progress record is not promoted. Real IDE and CLI transcript smoke tests now produce recaps, and the full collector suite plus go vet pass.

collector/internal/vendors/cursor/parse.go
calvintvu: Fixed in bbe90237. assistantResult is cleared at each user boundary and before every assistant record, so mixed text/tool records cannot retain a prior turn result. TestParseTranscriptClearsResultForToolPreamble covers the regression.

collector/internal/vendors/cursor/parse.go
calvintvu: **[P0 - RESOLVED]** Fixed in `bbe90237` and `7da71e10`: reply candidates are finalized at observed user or successful-turn boundaries, while the trailing candidate is promoted only when Cursor metadata does not mark the session interactive. The real IDE and CLI samples now produce recaps without promoting mixed/tool progress records.\n\n

collector/internal/vendors/cursor/parse.go
calvintvu: **[P1 - RESOLVED]** Fixed in `bbe90237`: each user boundary and assistant record clears `assistantResult` before a new final text-only reply can replace it, so a mixed tool preamble cannot retain the prior turn result. The regression test and full collector suite pass.\n\n

PR #201, opened 2026-09-17
Preserve complete Codex records over SSH

luhe19001:
## Context

SSH collection must preserve complete Codex session revisions without allowing interrupted, malformed, incomplete, or oversized refreshes to replace the last good data. The portable record contract from #203 is now merged into `main`; this PR adds its SSH transport, validation, and durable-cache layer.

## Changes

- Transport bounded complete records for Codex root and descendant sessions through helper and SFTP collection.
- Require exactly one full record for every session in a changed family, and force recollection of legacy root-only cache generations.
- Convert aggregate family-size failures into structured skips that retain the last complete generation.
- Size helper requests using their exact framed wire encoding.
- Require complete protocol responses and authoritative inventories before publishing replacements or tombstones.
- Add source-bound cache validation, atomic publication, previous-generation fallback, and exact revision reads.

## Stack

This is 2 of 2 and now targets `main`. The contract PR #203 has merged.

Downstream PRs #202 and #205 currently target this branch and must be reconciled separately after #201 merges.

## Test

- `cd collector && make check` — passed
- `cd collector && make test` — passed
- Focused race coverage for the corrected remote, helper, protocol, and cache regressions — passed
- GitHub Frontend, Collector, and Release checks — passed

PR #203, opened 2026-09-17
Add portable full-session record contract

luhe19001:
## Context

Remote collection and downstream storage need one stable, lossless representation of a parsed session revision. This PR defines that versioned contract while excluding machine-dependent filesystem enrichment and raw transcript data.

## Changes

- Add canonical `full-session-record/v1` records with exact identity, parent relationships, ordered file-change bodies, and nullable cost provenance.
- Enforce deterministic encoding and bounded validation for input size, strings, arrays, timestamps, counts, hashes, and costs.
- Add portable collector composition, generated boundary fixtures, and content-free measurement tooling.

## Stack

This is 1 of 2 and targets `main`.

1. #203 — portable record contract and composition
2. #201 — SSH transport, durable cache, and exact reads

Merge this PR first, then retarget #201 to `main`.

## Test

- `cd collector && make check && make test` — passed
- `cd frontend && npm run lint && npm run format:check && npm test && npm run build` — passed
- Release build and smoke tests — passed


collector/internal/fullsessionrecord/adapter.go
```diff
@@ -0,0 +1,230 @@
...
+func ToSession(record fullsessionv1.Record) (*session.Session, error) {
+	if err := fullsessionv1.Validate(record); err != nil {
+		return nil, err
+	}
+	value := &session.Session{
+		Agent: record.Agent, ID: record.SessionID, Name: cloneString(record.Session.Name), Summary: cloneString(record.Session.Summary),
+		Status: cloneString(record.Session.Status), WorkingDirectory: record.Session.WorkingDirectory,
+		Branch: cloneString(record.Session.Branch), EditedFileCount: record.Session.EditedFileCount,
+		DurationMs: cloneInt(record.Session.DurationMs), Tokens: map[string]session.ModelTokens{},
+		Cost: optionalDollars(record.Session.CostMicroUSD), UnpricedModels: append([]string{}, record.Session.UnpricedModels...),
```
milanshen: **[P0] Preserve unknown costs across round trips**

For a session with no recorded cost and no token rows, `AttachCost` intentionally leaves `Cost == nil`. `FromSession` turns that into `costMicroUsd: 0`, and this line always restores a non-nil zero; `eligibilityFor` consequently changes from `failed` to `eligible`, while the UI changes from `—` to `≈$0.00`. Make session and subagent cost fields nullable (`*int64` or an equivalent presence flag) and preserve nil in both directions.

collector/internal/collector/collector.go
```diff
@@ -299,6 +302,25 @@ func ListRemote(
 	return sessions
 }
 
+func finalizePortableSessionsSource(
+	parsed []*vendors.ParsedSession,
+	metadata map[string]*vendors.SessionMetadata,
+	source vendors.ReadSource,
+) []*vendors.ParsedSession {
+	applySessionEnrichment(parsed, metadata)
+	applyActivityFallbacks(parsed)
```
milanshen: **[P1] Keep portable revision timestamps deterministic**

When independently parsed copies have no transcript timestamps and source metadata is unavailable, this fallback ultimately uses `time.Now()`. Because `Freeze` hashes those timestamps, reparsing the unchanged transcript after a restart produces a different revision ID. Portable composition should use a stable source timestamp or reject missing timing data instead of applying the UI-oriented wall-clock fallback.

collector/internal/collector/collector.go
```diff
@@ -299,6 +302,25 @@ func ListRemote(
 	return sessions
 }
 
+func finalizePortableSessionsSource(
```
milanshen: **[P2] Share the session-finalization pipeline**

This function duplicates `finalizeSessionsSource` line-for-line except for the final liveness flag. A future enrichment or ordering fix applied to only one copy would silently break the stated local/helper/SFTP parity. Extract a common finalizer with an explicit `livenessAuthoritative` parameter.

collector/fullsession/v1/record.go
```diff
@@ -0,0 +1,406 @@
...
+	if err := validate(record, false); err != nil {
+		return Record{}, err
+	}
+	preimage, err := json.Marshal(record)
+	if err != nil {
+		return Record{}, err
+	}
+	digest := sha256.Sum256(preimage)
+	record.RevisionID = hex.EncodeToString(digest[:])
+	if err := Validate(record); err != nil {
```
milanshen: **[P2] Avoid recomputing the entire frozen record**

A record near the supported 64 MiB limit currently incurs three body-hash passes and three full JSON encodes: initial derivation, validation, and final sizing. Reuse the already-validated preimage and derived revision, then marshal the final record once to enforce the aggregate limit.

milanshen (commented):
## Review summary

This PR introduces a portable immutable full-session record and conversion layer. The overall contract boundary is reasonable, but round-trip conversion still loses source semantics and therefore is not safe for cache-backed use.

**Findings:** 1 blocking (P0), 1 should-fix (P1), 2 nice-to-have (P2) — see inline comments.

Review profile: Milan

collector/internal/collector/collector.go
luhe19001: Consolidated both paths into `finalizeSessionsSource`, with explicit liveness and local-activity-fallback flags.

Resolved in `44adfb9da3353a34e2e0a6d9483fd0269088b67b`.

collector/fullsession/v1/record.go
luhe19001: Removed redundant post-derivation validation. `Freeze` now validates the preimage, derives the revision, and marshals the final record once for size enforcement.

Resolved in `44adfb9da3353a34e2e0a6d9483fd0269088b67b`.

collector/internal/fullsessionrecord/adapter.go
luhe19001: Made session and subagent aggregate costs nullable and preserved nil through marshal/decode and adapter round trips.

Resolved in `44adfb9da3353a34e2e0a6d9483fd0269088b67b`.

collector/internal/collector/collector.go
luhe19001: Portable finalization now uses only source-derived timing. Missing timestamps remain invalid and are rejected by `Freeze` instead of using wall-clock time.

Resolved in `44adfb9da3353a34e2e0a6d9483fd0269088b67b`.

luhe19001:
@agent-review

Resolver pushed `44adfb9da3353a34e2e0a6d9483fd0269088b67b` and resolved the addressed review feedback. Please review the updated head.

collector/internal/fullsessionrecord/adapter.go
```diff
@@ -0,0 +1,234 @@
...
+	fullsessionv1 "github.com/centauri-ai/coslash/collector/fullsession/v1"
+	"github.com/centauri-ai/coslash/collector/internal/collector"
+	"github.com/centauri-ai/coslash/collector/internal/session"
+	"github.com/centauri-ai/coslash/collector/internal/vendors"
+)
+
+// FromParsedFamily runs the same shared composition used by local and remote
+// library sessions before freezing the rooted complete records.
+func FromParsedFamily(sourceID, vendor string, source vendors.ReadSource, parsed []*vendors.ParsedSession, metadata *vendors.SessionMetadata) ([]fullsessionv1.Record, error) {
+	roots := collector.ListRemote(source, map[string]vendors.RemoteCollection{
```
milanshen: **[P0] Preserve complete subagent text before display projection**

`ListRemote` calls `enrichSubagents`, whose `subagentFrom` applies `session.Truncate` to every child prompt and summary. Even a short multiline prompt has its whitespace collapsed, while text over 280 runes is cut off; because child sessions are folded into the root rather than emitted separately, the original data is permanently absent from the supposedly complete record.

Build records directly from the parsed family before the UI-oriented subagent projection, or add lossless child task/result fields to the composition model.

collector/internal/collector/collector.go
```diff
@@ -283,7 +289,10 @@ func ListRemote(
 		parsed = append(parsed, collection.Sessions...)
 		metadata[agent] = collection.Metadata
 	}
-	roots := finalizeSessionsSource(parsed, metadata, source)
+	// Portable composition intentionally does not apply local liveness or
+	// filesystem-derived enrichment. The same semantics are used when source is
+	// LocalReadSource so canonical local/helper/SFTP records remain comparable.
+	roots := finalizeSessionsSource(parsed, metadata, source, false, false)
```
milanshen: **[P0] Exclude volatile liveness from portable revisions**

Passing `livenessAuthoritative=false` does not stop `resolveStatus` or `subagentStatus` from consuming `Metadata.Live`. The same Codex transcript therefore produces `busy`/`running` locally when `lsof` supplies `interactive`, but nil/`aborted` through helper or SFTP metadata; `resolveStatus` can also switch from busy to idle solely when the two-minute wall-clock threshold passes. These user-visible statuses change the canonical bytes and revision ID despite the claimed local/helper/SFTP parity.

Add an explicit portable mode that ignores live signals for both roots and subagents and derives only stable transcript-backed statuses.

collector/fullsession/v1/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+The record includes the portable parsed session envelope, detail, subagents,
+and every ordered file-change body. Portable means values obtained from the
+transcript or allowlisted vendor metadata before local filesystem probes run.
+It excludes raw vendor transcript rows, SSH configuration, credentials owned
+by coSlash, sockets, and transport commands. File-change IDs are opaque
+identifiers, never filesystem paths.
+
+`sourceId`, `agent`, `sessionId`, and the content-derived `revisionId` form the
+exact read identity. `revisionId` is SHA-256 over canonical JSON with an empty
```
milanshen, line 15: **[P1] Define canonical JSON for non-Go consumers**

The contract hashes “canonical JSON” without specifying member ordering or escaping. For example, a file-change body containing `<` is emitted as `\u003c` by Go's encoder, while another valid JSON or RFC-canonical encoder may emit `<`; that consumer computes a different revision and its record is rejected by `Decode`.

Specify the exact language-neutral canonicalization rules and add escaping fixtures, or adopt a published canonical JSON standard.

milanshen (commented):
**Carried over:** 4 resolved, 0 closed, 0 unresolved (4 prior findings).

| --- | --- | --- | --- |
| `adapter.go:128` Preserve unknown costs | P0 | Resolved | Fixed in `44adfb9da3353a34e2e0a6d9483fd0269088b67b`: nullable aggregate costs remain nil through marshal, decode, and adapter round trips. |
| `collector.go` Deterministic portable timestamps | P1 | Resolved | Fixed in `44adfb9da3353a34e2e0a6d9483fd0269088b67b`: portable composition uses source-derived timing and rejects missing timestamps. |
| `collector.go` Shared finalization pipeline | P2 | Resolved | Fixed in `44adfb9da3353a34e2e0a6d9483fd0269088b67b`: both modes now use one parameterized finalizer. |
| `record.go` Redundant record recomputation | P2 | Resolved | Fixed in `44adfb9da3353a34e2e0a6d9483fd0269088b67b`: the redundant post-derivation validation pass was removed. |

## Review summary

This PR defines a portable immutable full-session record, validation, fixtures, measurement, and collector adapters. The typed-contract direction is reasonable, and the updated commit demonstrably resolves all four earlier findings. The producer still cannot guarantee lossless, source-independent record bytes, which undermines the core complete-record invariant.

**Findings:** 2 blocking (P0), 2 should-fix (P1), 0 nice-to-have (P2) — see inline comments.

Review profile: Milan

Requested in [PR comment](https://github.com/centauri-ai/coslash/pull/203#issuecomment-5719343541)

collector/fullsession/v1/record.go
```diff
@@ -0,0 +1,407 @@
...
+}
+
+// Freeze computes all body hashes/counts and the immutable revision identity.
+// The revision is the SHA-256 of the canonical record with revisionId empty.
+func Freeze(record Record) (Record, error) {
+	record.SchemaVersion = SchemaVersion
+	record.RevisionID = ""
+	for editIndex := range record.Session.FileEdits {
+		for changeIndex := range record.Session.FileEdits[editIndex].Changes {
+			change := &record.Session.FileEdits[editIndex].Changes[changeIndex]
```
milanshen, line 171: **[P1] Avoid mutating caller-owned storage while freezing**

`Record` is copied shallowly, so filling change IDs, counts, and hashes mutates the input's backing slices—even when later validation fails. After success, a caller modifying its retained input slice also silently changes the returned frozen record while leaving its revision and body hashes stale.

Deep-clone the record before normalization and hashing so the frozen result does not alias caller-owned slices or pointers.

collector/internal/fullsessionrecord/adapter.go
luhe19001: Added a lossless portable composition path for full records, preserving multiline and long subagent task/result text. The display path retains truncation behavior, with regression coverage.

Resolved in `f6af405ef2e7df0494707228a77cce49f8ab1e26`.

collector/internal/collector/collector.go
luhe19001: Portable composition now ignores live metadata and wall-clock status refinement for roots and subagents, deriving stable transcript-backed statuses instead. Added parity coverage with and without live signals.

Resolved in `f6af405ef2e7df0494707228a77cce49f8ab1e26`.

collector/fullsession/v1/README.md
luhe19001: Documented exact member ordering, encoding, escaping, and hashing rules, and published an escaping fixture covering HTML characters, separators, controls, quotes, and backslashes.

Resolved in `f6af405ef2e7df0494707228a77cce49f8ab1e26`.

collector/fullsession/v1/record.go
luhe19001: Freeze now deep-clones all nested pointers and slices before normalization. Tests verify failed freezes do not mutate inputs and successful results remain independent of caller mutations.

Resolved in `f6af405ef2e7df0494707228a77cce49f8ab1e26`.

luhe19001:
@agent-review

Resolver pushed `f6af405ef2e7df0494707228a77cce49f8ab1e26` and resolved the addressed review feedback. Please review the updated head.

collector/fullsession/v1/record.go
```diff
@@ -0,0 +1,468 @@
...
+}
+
+func Decode(data []byte) (Record, error) {
+	if len(data) > MaxRecordBytes {
+		return Record{}, ErrOversized
+	}
+	decoder := json.NewDecoder(bytes.NewReader(data))
+	decoder.DisallowUnknownFields()
+	var record Record
+	if err := decoder.Decode(&record); err != nil {
```
milanshen, line 288: **[P0] Bound collection sizes before decoding them**

`Decode` materializes every slice before `Validate` enforces `MaxItems`. A payload under the permitted 64 MiB limit can contain roughly twenty million empty `usage` objects, causing over a gigabyte of slice allocation and potentially terminating the collector before it returns `ErrInvalid`.

Use bounded custom unmarshalling or token-level array counting so each collection is rejected once it reaches `MaxItems`, before appending further elements.

milanshen (commented):
**Carried over:** 8 resolved, 0 closed, 0 unresolved (8 prior findings).

| --- | --- | --- | --- |
| `adapter.go:128` Preserve unknown costs | P0 | Resolved | Fixed in `44adfb9da3353a34e2e0a6d9483fd0269088b67b`: nullable session and subagent costs survive record round trips. |
| `collector.go` Deterministic portable timestamps | P1 | Resolved | Fixed in `44adfb9da3353a34e2e0a6d9483fd0269088b67b`: portable composition uses source timing and rejects missing timestamps. |
| `collector.go` Shared finalization pipeline | P2 | Resolved | Fixed in `44adfb9da3353a34e2e0a6d9483fd0269088b67b`: both paths use the parameterized shared finalizer. |
| `record.go` Avoid redundant recomputation | P2 | Resolved | Fixed in `44adfb9da3353a34e2e0a6d9483fd0269088b67b`: redundant post-derivation validation was removed. |
| `adapter.go` Preserve complete subagent text | P0 | Resolved | Fixed in `f6af405ef2e7df0494707228a77cce49f8ab1e26`: portable composition retains full task and result text. |
| `collector.go` Exclude volatile portable liveness | P0 | Resolved | Fixed in `f6af405ef2e7df0494707228a77cce49f8ab1e26`: portable root and Codex-child statuses ignore live metadata and wall-clock refinement. |
| `README.md:15` Define canonical JSON | P1 | Resolved | Fixed in `f6af405ef2e7df0494707228a77cce49f8ab1e26`: ordering, escaping, encoding, and hashing rules are specified with an escaping fixture. |
| `record.go:159` Avoid caller-storage mutation | P1 | Resolved | Fixed in `f6af405ef2e7df0494707228a77cce49f8ab1e26`: Freeze deep-clones nested pointers and slices before normalization. |

## Review summary

This PR introduces a typed canonical full-session record, adapters, fixtures, and measurement tooling. The contract direction and all eight prior fixes are sound, but record construction still violates its lossless and deterministic guarantees, while the shared display path and decoder introduce blocking regressions. These boundaries need correction before merge.

**Findings:** 4 blocking (P0), 1 should-fix (P1), 0 nice-to-have (P2) — see inline comments.

Review profile: Milan

Requested in [PR comment](https://github.com/centauri-ai/coslash/pull/203#issuecomment-5719577996)

collector/internal/collector/collector.go
```diff
@@ -283,7 +312,10 @@ func ListRemote(
 		parsed = append(parsed, collection.Sessions...)
 		metadata[agent] = collection.Metadata
 	}
-	roots := finalizeSessionsSource(parsed, metadata, source)
+	// Portable composition intentionally does not apply local liveness or
+	// filesystem-derived enrichment. The same semantics are used when source is
+	// LocalReadSource so canonical local/helper/SFTP records remain comparable.
+	roots := finalizeSessionsSource(parsed, metadata, source, false, false, preserveSubagentText)
```
milanshen: **[P0] Keep live status on the remote display path**

This shared call disables live status for both `ComposePortable` and the existing `ListRemote` display path. When `composeFromGeneration` supplies fresh `Metadata.Live`, an active root now becomes nil/its stale hint and an in-turn Codex child becomes `aborted`; the UI shows unknown liveness and can enable Resume, while `LaunchSession` likewise no longer blocks resuming that active remote session.

Pass `useLiveStatus=true` for `ListRemote` and disable it only for canonical portable composition.

collector/fullsession/v1/record.go
```diff
@@ -0,0 +1,468 @@
...
+	CacheReadInputTokens       int    `json:"cacheReadInputTokens"`
+	CostMicroUSD               int64  `json:"costMicroUsd"`
+}
+
+type SubagentCommand struct {
+	Label   string `json:"label"`
+	Command string `json:"command"`
+}
+
+type Subagent struct {
```
milanshen, line 99: **[P0] Preserve complete child-session data**

The portable producer still reduces children to this display-shaped `Subagent`, which has no file edits, digest, todos, synthesis, or descendants. If a Codex subagent edits a file or spawns another subagent, those change bodies and the grandchild disappear because `FromParsedFamily` freezes only roots and emits no separate child records, violating the complete-record invariant.

Represent the full family recursively or as flattened sessions with parent identities before applying the UI projection.

collector/internal/fullsessionrecord/adapter.go
```diff
@@ -0,0 +1,234 @@
...
+	fullsessionv1 "github.com/centauri-ai/coslash/collector/fullsession/v1"
+	"github.com/centauri-ai/coslash/collector/internal/collector"
+	"github.com/centauri-ai/coslash/collector/internal/session"
+	"github.com/centauri-ai/coslash/collector/internal/vendors"
+)
+
+// FromParsedFamily runs portable composition without the display projection
+// before freezing the rooted complete records.
+func FromParsedFamily(sourceID, vendor string, source vendors.ReadSource, parsed []*vendors.ParsedSession, metadata *vendors.SessionMetadata) ([]fullsessionv1.Record, error) {
+	roots := collector.ComposePortable(source, map[string]vendors.RemoteCollection{
```
milanshen: **[P0] Compose from a clone of the parsed family**

`ComposePortable` mutates the supplied session pointers: it appends projected subagents and rewrites or inserts spawn digest rows. Calling `FromParsedFamily` twice with the same unchanged parsed family therefore produces one child on the first call and duplicate children/digest rows on the second, yielding a different revision for identical source data.

Deep-clone the parsed family before finalization or make portable composition side-effect-free.

collector/internal/fullsessionrecord/adapter.go
```diff
@@ -0,0 +1,234 @@
...
+	for _, item := range record.Session.Digest {
+		value.Digest = append(value.Digest, session.DigestEntry{
+			Turn: item.Turn, Category: item.Category, Description: item.Description, Answer: item.Answer,
+			SubagentID: item.SubagentID, Time: item.TimeMs,
+		})
+	}
+	for _, item := range record.Session.FileEdits {
+		changes := make([]session.FileChange, 0, len(item.Changes))
+		for _, change := range item.Changes {
+			changes = append(changes, session.FileChange{
```
milanshen, line 179: **[P1] Preserve opaque file-change IDs through the adapter**

This conversion drops `FileChange.ID`. A valid record using an opaque ID such as `patch-abcd` becomes a private session without that identity, and converting it back regenerates a positional `change-...` ID, changing the revision and breaking any body association keyed by the original ID.

Carry the ID in `session.FileChange` and `FileEditWithChanges`, or keep record-backed identities outside this lossy adapter.

collector/fullsession/v1/record.go
luhe19001: Emitted a complete record for every family member with parentSessionId linkage, preserving child edits, todos, synthesis, digest, and descendants.

Resolved in `5d17b48cda1b821bc68878e0ab28d279dbb764da`.

collector/internal/fullsessionrecord/adapter.go
luhe19001: Portable family composition now operates on a deep clone. Repeated conversion is deterministic and leaves parsed sessions unchanged.

Resolved in `5d17b48cda1b821bc68878e0ab28d279dbb764da`.

collector/fullsession/v1/record.go
luhe19001: Added token-level collection counting before typed decoding, rejecting arrays once they exceed MaxItems without materializing oversized slices.

Resolved in `5d17b48cda1b821bc68878e0ab28d279dbb764da`.

collector/internal/collector/collector.go
luhe19001: Restored live-status handling for ListRemote while keeping canonical portable composition independent of live metadata.

Resolved in `5d17b48cda1b821bc68878e0ab28d279dbb764da`.

collector/internal/fullsessionrecord/adapter.go
luhe19001: Added FileChange.ID to the private model and preserved opaque IDs through both adapter directions.

Resolved in `5d17b48cda1b821bc68878e0ab28d279dbb764da`.

luhe19001:
@agent-review

Resolver pushed `5d17b48cda1b821bc68878e0ab28d279dbb764da` and resolved the addressed review feedback. Please review the updated head.

collector/internal/collector/collector.go
```diff
@@ -273,17 +284,49 @@ func ListRemote(
 	collections map[string]vendors.RemoteCollection,
 	since int64,
 ) []*session.Session {
-	parsed := []*vendors.ParsedSession{}
-	metadata := map[string]*vendors.SessionMetadata{}
-	for _, agent := range []string{vendors.AgentClaude, vendors.AgentCodex} {
-		collection, ok := collections[agent]
-		if !ok {
-			continue
+	return listRemote(source, collections, since, false, true)
```
milanshen, line 287: **[P0] Preserve waiting status when liveness is unavailable**

`ListRemote` now enables authoritative live-status handling for every source. During SFTP collection or cache loading through `nullReadSource`, no live signal may exist; a transcript parsed while awaiting approval therefore enters `resolveStatus` as `waiting` but is cleared to nil. The remote UI moves that session from Waiting to Unknown and drops it from the machine's waiting count. Previously, only `LocalReadSource` treated absent liveness as authoritative. Keep separate flags for consuming available live signals and for whether their absence disproves `waiting`, with the latter derived from the source.

milanshen (commented):
**Carried over:** 13 resolved, 0 closed, 0 unresolved (13 prior findings).

| --- | --- | --- | --- |
| `adapter.go` Preserve unknown costs | P0 | Resolved | Fixed in `44adfb9da335`: nullable session and subagent costs survive adapter round trips. |
| `collector.go` Deterministic portable timestamps | P1 | Resolved | Fixed in `44adfb9da335`: portable composition uses source timing and rejects missing timestamps. |
| `collector.go` Shared finalization pipeline | P2 | Resolved | Fixed in `44adfb9da335`: display and portable paths share the parameterized finalizer. |
| `record.go` Avoid redundant recomputation | P2 | Resolved | Fixed in `44adfb9da335`: Freeze validates the preimage and performs only the required final encoding. |
| `adapter.go` Preserve complete subagent text | P0 | Resolved | Fixed in `f6af405ef2e7`: portable composition retains complete task and result text. |
| `collector.go` Exclude volatile portable liveness | P0 | Resolved | Fixed in `f6af405ef2e7`: portable root and child statuses ignore live metadata and wall-clock refinement. |
| `README.md` Define canonical JSON | P1 | Resolved | Fixed in `f6af405ef2e7`: ordering and escaping rules are specified with a matching fixture. |
| `record.go` Avoid caller-storage mutation | P1 | Resolved | Fixed in `f6af405ef2e7`: Freeze deep-clones nested pointers and slices. |
| `record.go` Bound collections before decoding | P0 | Resolved | Fixed in `5d17b48cda1b`: token-level counting rejects oversized arrays before typed decoding. |
| `collector.go` Keep live status on remote display | P0 | Resolved | Fixed in `5d17b48cda1b`: ListRemote again consumes available live metadata. |
| `record.go` Preserve complete child-session data | P0 | Resolved | Fixed in `5d17b48cda1b`: every family member receives a complete record with parent linkage. |
| `adapter.go` Compose from a cloned family | P0 | Resolved | Fixed in `5d17b48cda1b`: portable composition deep-clones parsed input before mutation. |
| `adapter.go` Preserve opaque file-change IDs | P1 | Resolved | Fixed in `5d17b48cda1b`: change IDs survive both adapter directions. |

## Review summary

The record contract and all prior corrections are sound, but restoring live display status introduced a blocking regression for SFTP sessions without authoritative liveness. The measured scope exceeded the default review budget, so prior findings and production code were prioritized; tests, fixtures, and documentation were checked for adequacy rather than line-by-line bug hunting. Subagents were unavailable, making this a single-pass fallback across the required review angles.

**Findings:** 1 blocking (P0), 0 should-fix (P1), 0 nice-to-have (P2) — see inline comments.

Review profile: Milan

Requested in [PR comment](https://github.com/centauri-ai/coslash/pull/203#issuecomment-5719721468)

collector/internal/collector/collector.go
luhe19001: Separated live-status consumption from authoritative absence. Non-local sources now preserve parsed waiting status when no live signal is available, while local collection still clears stale waiting status. Added regression coverage.

Resolved in `55a3ef9db18ef2522697c2f8ccba5c33327ac8a0`.

luhe19001:
@agent-review

Resolver pushed `55a3ef9db18ef2522697c2f8ccba5c33327ac8a0` and resolved the addressed review feedback. Please review the updated head.

milanshen (commented):
**Carried over:** 14 resolved, 0 closed, 0 unresolved (14 prior findings).

| --- | --- | --- | --- |
| `adapter.go` Preserve unknown costs | P0 | Resolved | Fixed in `44adfb9da335`: nullable aggregate costs survive adapter round trips. |
| `collector.go` Deterministic portable timestamps | P1 | Resolved | Fixed in `44adfb9da335`: portable composition uses only source-derived timing. |
| `collector.go` Shared finalization pipeline | P2 | Resolved | Fixed in `44adfb9da335`: both paths use the parameterized shared finalizer. |
| `record.go` Avoid redundant recomputation | P2 | Resolved | Fixed in `44adfb9da335`: Freeze performs one final size-enforcement encoding. |
| `adapter.go` Preserve complete subagent text | P0 | Resolved | Fixed in `f6af405ef2e7`: portable Codex projection retains full task/result text. |
| `collector.go` Exclude volatile portable liveness | P0 | Resolved | Fixed in `f6af405ef2e7`: portable statuses ignore live metadata and wall-clock refinement. |
| `README.md` Define canonical JSON | P1 | Resolved | Fixed in `f6af405ef2e7`: ordering and escaping rules have a matching fixture. |
| `record.go` Avoid caller-storage mutation | P1 | Resolved | Fixed in `f6af405ef2e7`: Freeze deep-clones nested storage. |
| `record.go` Bound collections before decoding | P0 | Resolved | Fixed in `5d17b48cda1b`: a single oversized array is rejected before typed decoding. |
| `collector.go` Keep live status on remote display | P0 | Resolved | Fixed in `5d17b48cda1b`: ListRemote consumes available live metadata. |
| `record.go` Preserve complete child-session data | P0 | Resolved | Fixed in `5d17b48cda1b`: every family member receives a complete linked record. |
| `adapter.go` Compose from a cloned family | P0 | Resolved | Fixed in `5d17b48cda1b`: portable composition no longer mutates parsed input. |
| `adapter.go` Preserve opaque file-change IDs | P1 | Resolved | Fixed in `5d17b48cda1b`: opaque IDs survive both adapter directions. |
| `collector.go` Preserve waiting without authoritative liveness | P0 | Resolved | Fixed in `55a3ef9db18e`: non-local/cache composition retains transcript-derived waiting status when live data is unavailable. |

## Review summary

The updated head fixes the reported waiting-status regression, and all prior findings remain resolved. A remaining resource-boundary flaw blocks merge; the adapter and corpus tool also need correctness hardening. At 26 files and over 1,000 production-source changed lines, production code and prior findings were prioritized while tests, fixtures, and documentation were checked for adequacy rather than line-by-line bug hunting.

**Findings:** 1 blocking (P0), 3 should-fix (P1), 0 nice-to-have (P2) — see inline comments.

Review profile: Milan

Requested in [PR comment](https://github.com/centauri-ai/coslash/pull/203#issuecomment-5719797283)

collector/fullsession/v1/record.go
```diff
@@ -0,0 +1,507 @@
...
+		delim, isDelim := token.(json.Delim)
+		if isDelim && (delim == ']' || delim == '}') {
+			if len(stack) > 0 {
+				stack = stack[:len(stack)-1]
+			}
+			continue
+		}
+		if len(stack) > 0 && stack[len(stack)-1].kind == '[' {
+			stack[len(stack)-1].items++
+			if stack[len(stack)-1].items > MaxItems {
```
milanshen: **[P0] Bound aggregate nested items before typed decoding**

The token pass limits each array independently, so a payload below 64 MiB can contain a few hundred `fileEdits`, each with 100,000 empty `changes` objects. Every array passes this check, after which typed decoding materializes roughly twenty million `FileChange` values—multiple gigabytes—before validation rejects them. Enforce a global/schema-aware nested-item allocation budget during the token pass so invalid remote records cannot exhaust the collector.

collector/internal/fullsessionrecord/adapter.go
```diff
@@ -0,0 +1,266 @@
...
+func ToSession(record fullsessionv1.Record) (*session.Session, error) {
+	if err := fullsessionv1.Validate(record); err != nil {
+		return nil, err
+	}
+	value := &session.Session{
+		Agent: record.Agent, ID: record.SessionID, Name: cloneString(record.Session.Name), Summary: cloneString(record.Session.Summary),
+		Status: cloneString(record.Session.Status), WorkingDirectory: record.Session.WorkingDirectory,
+		Branch: cloneString(record.Session.Branch), EditedFileCount: record.Session.EditedFileCount,
+		DurationMs: cloneInt(record.Session.DurationMs), Tokens: map[string]session.ModelTokens{},
+		Cost: optionalDollars(record.Session.CostMicroUSD), UnpricedModels: append([]string{}, record.Session.UnpricedModels...),
```
milanshen: **[P1] Preserve null arrays through adapter round trips**

The contract permits both null and empty arrays, but `append([]string{}, value...)` converts null `unpricedModels`, `commands`, `commits`, `commitShas`, and synthesis lists into non-null empty slices. A valid decoded record containing any such null array therefore gets different canonical bytes and a different revision after `ToSession` → `FromSession`. Use nil-preserving clones, or make non-null arrays a documented and validated canonical requirement.

collector/internal/fullsessionrecord/adapter.go
```diff
@@ -0,0 +1,266 @@
...
+			Model: cloneString(record.Session.Model), ContextTokens: cloneInt(record.Session.ContextTokens),
+			ContextWindow: cloneInt(record.Session.ContextWindow), Turns: record.Session.Turns,
+			ToolUses: record.Session.ToolUses, Errors: record.Session.Errors, Compactions: record.Session.Compactions,
+			FirstPrompt: cloneString(record.Session.FirstPrompt), Commands: append([]string{}, record.Session.Commands...),
+			Commits: append([]string{}, record.Session.Commits...), CommitSHAs: append([]string{}, record.Session.CommitSHAs...),
+			PullRequests:     record.Session.PullRequests,
+			SynthesisPending: record.Session.SynthesisPending, DeclaredGoal: cloneString(record.Session.DeclaredGoal),
+		},
+	}
+	for _, usage := range record.Session.Usage {
```
milanshen: **[P1] Preserve canonical usage ordering through the adapter**

`ToSession` collapses the ordered `Usage` array into a map, while `FromSession` later sorts map keys. Because validation accepts any unique model order and the contract says arrays retain their declared order, a valid record with usage ordered `[z, a]` round-trips as `[a, z]` and receives a new revision. Either require sorted usage in the contract and validator or retain the original order through the private representation.

collector/fullsession/v1/cmd/measure/main.go
```diff
@@ -0,0 +1,46 @@
...
+	flags := flag.NewFlagSet("measure", flag.ExitOnError)
+	version := flags.String("collector-version", "", "collector revision measured")
+	flags.Parse(os.Args[1:])
+	if *version == "" || flags.NArg() == 0 {
+		fmt.Fprintln(os.Stderr, "--collector-version and one or more record files are required")
+		os.Exit(2)
+	}
+	records := make([]fullsessionv1.Record, 0, flags.NArg())
+	for _, path := range flags.Args() {
+		data, err := os.ReadFile(path)
```
milanshen: **[P1] Bound memory before decoding measurement inputs**

`os.ReadFile` allocates the complete input before `Decode` can apply `MaxRecordBytes`, and the CLI retains every accepted record—including change bodies—until the entire corpus is measured. An accidentally multi-gigabyte file or a large corpus of individually valid records can therefore terminate the measurement process instead of producing a report. Read each file through a `MaxRecordBytes+1` limit and aggregate sizes/counters incrementally rather than retaining all decoded bodies.

collector/internal/fullsessionrecord/adapter.go
luhe19001: Documented and validated ascending model order for session and subagent usage arrays, matching adapter emission so accepted records round-trip canonically.

Resolved in `7d6f527da4b6b3e389ad2053e89debc98e5188ff`.

collector/fullsession/v1/cmd/measure/main.go
luhe19001: Added byte-limited reader decoding and changed measurement to incrementally retain only record sizes and counters, not decoded records or change bodies.

Resolved in `7d6f527da4b6b3e389ad2053e89debc98e5188ff`.

collector/fullsession/v1/record.go
luhe19001: The token prepass now bounds aggregate array items and nesting before typed decoding, preventing independently valid nested arrays from causing unbounded allocations.

Resolved in `7d6f527da4b6b3e389ad2053e89debc98e5188ff`.

collector/internal/fullsessionrecord/adapter.go
luhe19001: Replaced eager empty-slice copies with nil-preserving clones in both adapter directions, including synthesis lists, and added revision-stability coverage.

Resolved in `7d6f527da4b6b3e389ad2053e89debc98e5188ff`.

luhe19001:
@agent-review

Resolver pushed `7d6f527da4b6b3e389ad2053e89debc98e5188ff` and resolved the addressed review feedback. Please review the updated head.

collector/fullsession/v1/record.go
```diff
@@ -0,0 +1,522 @@
...
+		if isDelim && (delim == ']' || delim == '}') {
+			if len(stack) > 0 {
+				stack = stack[:len(stack)-1]
+			}
+			continue
+		}
+		if len(stack) > 0 && stack[len(stack)-1].kind == '[' {
+			stack[len(stack)-1].items++
+			totalItems++
+			if stack[len(stack)-1].items > MaxItems || totalItems > MaxItems {
```
milanshen: **[P1] Enforce the aggregate item limit while freezing**

The new aggregate limit applies only during Decode. A record containing 50,001 file edits with one valid change each remains well below 64 MiB and passes Freeze/Marshal because every individual slice is below MaxItems, but Decode counts 100,002 array elements and rejects the package's own output.

Apply the same aggregate budget during typed validation/Freeze, ideally through one shared limit policy.

collector/internal/fullsessionrecord/adapter.go
```diff
@@ -0,0 +1,265 @@
...
+		models = append(models, model)
+	}
+	sort.Strings(models)
+	for _, model := range models {
+		tokens := value.Tokens[model]
+		record.Session.Usage = append(record.Session.Usage, fullsessionv1.ModelUsage{
+			Model: model, InputTokens: tokens.InputTokens, OutputTokens: tokens.OutputTokens,
+			CacheCreationInputTokens:   tokens.CacheCreationInputTokens,
+			CacheCreation1hInputTokens: tokens.CacheCreation1hInputTokens,
+			CacheReadInputTokens:       tokens.CacheReadInputTokens, CostMicroUSD: micros(tokens.Cost),
```
milanshen: **[P1] Populate per-model costs before freezing usage**

Normal Codex and Claude token buckets populate counts but leave `ModelTokens.Cost` at zero; `AttachCost` estimates only the session aggregate. Consequently, a normal priced-model session can have a positive aggregate cost while every emitted `ModelUsage.costMicroUsd` is falsely zero. The fixture masks this by manually setting the model cost.

Calculate each priced model's cost during enrichment or when constructing these usage rows, while retaining `unpricedModels` for genuinely unknown prices.

collector/internal/collector/collector.go
```diff
@@ -273,17 +284,49 @@ func ListRemote(
...
+	for _, root := range roots {
+		rootKeys[sessionKey{agent: root.Session.Agent, id: root.Session.ID}] = true
+	}
+	for _, item := range parsed {
+		byKey[sessionKey{agent: item.Session.Agent, id: item.Session.ID}] = item
+	}
+	portable := make([]PortableSession, 0, len(parsed))
+	for _, item := range parsed {
+		if belongsToRoot(item, byKey, rootKeys) {
+			portable = append(portable, PortableSession{ParentSessionID: item.ParentID, Session: item.Session})
```
milanshen, line 315: **[P1] Resolve names for standalone descendant records**

Portable composition now emits every parsed child, but `resolveNames` still processes roots only. For a child whose `ParsedSession.Name` is set while `Session.Name` is nil, the parent subagent projection has the name but the child's complete standalone record does not.

Resolve names across all composed family members before emitting portable records.

collector/internal/fullsessionrecord/adapter.go
```diff
@@ -0,0 +1,265 @@
...
+		},
+	}
+	models := make([]string, 0, len(value.Tokens))
+	for model := range value.Tokens {
+		models = append(models, model)
+	}
+	sort.Strings(models)
+	for _, model := range models {
+		tokens := value.Tokens[model]
+		record.Session.Usage = append(record.Session.Usage, fullsessionv1.ModelUsage{
```
milanshen: **[P1] Preserve non-null empty structural arrays**

The contract distinguishes `[]` from `null`, but structured collections are rebuilt solely through append loops. A valid empty `usage`, `subagents`, `todos`, `digest`, `fileEdits`, or nested command/usage/change array therefore becomes null after `ToSession` → `FromSession`, changing canonical bytes and the revision. The published fixture already contains an empty subagent usage array.

Retain collection-presence information in the private representation, or require and normalize one canonical array representation.

milanshen (commented):
**Carried over:** 18 resolved, 0 closed, 0 unresolved (18 prior findings).

| --- | --- | --- | --- |
| `adapter.go` — preserve unknown costs | P0 | Resolved | Fixed in `44adfb9da335`: nullable aggregate costs survive adapter round trips. |
| `collector.go` — deterministic portable timestamps | P1 | Resolved | Fixed in `44adfb9da335`: portable composition rejects missing source timing instead of using wall-clock time. |
| `collector.go` — shared finalization pipeline | P2 | Resolved | Fixed in `44adfb9da335`: both paths use the parameterized common finalizer. |
| `record.go` — avoid redundant recomputation | P2 | Resolved | Fixed in `44adfb9da335`: Freeze validates one preimage and performs one final size encoding. |
| `adapter.go` — preserve complete subagent text | P0 | Resolved | Fixed in `f6af405ef2e7`: portable projection retains full task and result text. |
| `collector.go` — exclude volatile portable liveness | P0 | Resolved | Fixed in `f6af405ef2e7`: portable statuses ignore live metadata and wall-clock refinement. |
| `README.md` — define canonical JSON | P1 | Resolved | Fixed in `f6af405ef2e7`: ordering and escaping rules are documented with a matching fixture. |
| `record.go` — avoid caller-storage mutation | P1 | Resolved | Fixed in `f6af405ef2e7`: Freeze deep-clones nested storage before normalization. |
| `record.go` — bound collections before decoding | P0 | Resolved | Fixed in `5d17b48cda1b`: individual oversized arrays are rejected before typed decoding. |
| `collector.go` — retain remote live status | P0 | Resolved | Fixed in `5d17b48cda1b`: ListRemote consumes available live metadata. |
| `record.go` — preserve complete child sessions | P0 | Resolved | Fixed in `5d17b48cda1b`: every family member receives a complete linked record. |
| `adapter.go` — avoid mutating parsed families | P0 | Resolved | Fixed in `5d17b48cda1b`: portable composition operates on a deep clone. |
| `adapter.go` — preserve opaque change IDs | P1 | Resolved | Fixed in `5d17b48cda1b`: opaque IDs survive both adapter directions. |
| `collector.go` — preserve waiting without authoritative liveness | P0 | Resolved | Fixed in `55a3ef9db18e`: non-local sources retain transcript-derived waiting status when liveness is absent. |
| `record.go` — bound aggregate nested items | P0 | Resolved | Fixed in `7d6f527da4b6`: the token prepass now enforces aggregate item and nesting budgets. |
| `adapter.go` — preserve null arrays | P1 | Resolved | Fixed in `7d6f527da4b6`: the reported null-to-empty conversions now use nil-preserving clones; the reverse direction is a separate finding below. |
| `adapter.go` — preserve canonical usage ordering | P1 | Resolved | Fixed in `7d6f527da4b6`: usage arrays must be strictly sorted, matching adapter output. |
| `measure/main.go` — bound measurement memory | P1 | Resolved | Fixed in `7d6f527da4b6`: files are byte-limited and only sizes and counters persist across the corpus. |

## Review summary

The portable-record architecture is generally sound, and all 18 prior scenarios remain fixed. New correctness gaps remain in producer/consumer compatibility and lossless portable enrichment. The 26-file change exceeds the review budget, so prior findings and production paths were prioritized; tests, documentation, and generated fixtures were assessed for adequacy rather than inspected line by line.

Review profile: Milan

Requested in [PR comment](https://github.com/centauri-ai/coslash/pull/203#issuecomment-5720013134)

collector/fullsession/v1/record.go
luhe19001: Added a shared aggregate collection budget to typed validation and Decode, including nested arrays. Freeze and Marshal now reject records exceeding MaxItems, with regression coverage.

Resolved in `5a35e9ceae87e3eef2a020dd9e8c43b69983951b`.

collector/internal/fullsessionrecord/adapter.go
luhe19001: AttachCost now populates each priced ModelTokens cost before freezing while preserving unknown models in unpricedModels. Added cost-enrichment coverage.

Resolved in `5a35e9ceae87e3eef2a020dd9e8c43b69983951b`.

collector/internal/collector/collector.go
luhe19001: Name resolution now processes every composed family member, so standalone descendant records receive parsed or enriched names.

Resolved in `5a35e9ceae87e3eef2a020dd9e8c43b69983951b`.

collector/internal/fullsessionrecord/adapter.go
luhe19001: The adapters now preserve null versus empty structural collections, including nested command, usage, and file-change arrays. Added revision round-trip coverage.

Resolved in `5a35e9ceae87e3eef2a020dd9e8c43b69983951b`.

luhe19001:
@agent-review

Resolver pushed `5a35e9ceae87e3eef2a020dd9e8c43b69983951b` and resolved the addressed review feedback. Please review the updated head.

collector/internal/session/models.go
```diff
@@ -116,17 +124,21 @@ func UnpricedModels(tokens map[string]ModelTokens) []string {
 }
 
 func AttachCost(s *Session, recorded *float64) {
+	for model, used := range s.Tokens {
+		if cost, ok := estimatedModelCost(model, used); ok {
+			used.Cost = cost
+			s.Tokens[model] = used
+		}
+	}
+	s.UnpricedModels = UnpricedModels(s.Tokens)
```
milanshen: **[P0] Do not mark authoritative costs as incomplete**

`UnpricedModels` is now populated before the `recorded != nil` branch. OpenCode supplies an authoritative recorded total, so whenever it reports a model absent from the embedded table, session cards show that total with the false warning “No pricing info … excluded from this estimate.” Preserve an empty unpriced list for authoritative totals, or carry cost provenance so the UI can distinguish recorded totals from partial estimates.

collector/fullsession/v1/record.go
```diff
@@ -0,0 +1,559 @@
...
+// The revision is the SHA-256 of the canonical record with revisionId empty.
+func Freeze(record Record) (Record, error) {
+	record = cloneRecord(record)
+	record.SchemaVersion = SchemaVersion
+	record.RevisionID = ""
+	for editIndex := range record.Session.FileEdits {
+		for changeIndex := range record.Session.FileEdits[editIndex].Changes {
+			change := &record.Session.FileEdits[editIndex].Changes[changeIndex]
+			if change.ID == "" {
+				change.ID = fmt.Sprintf("change-%06d-%06d", editIndex, changeIndex)
```
milanshen: **[P1] Avoid collisions when generating change IDs**

An empty first change receives `change-000000-000000`, but that same string is also a valid opaque ID for a later supplied change because no namespace is reserved. Freeze then rejects the otherwise valid record as containing duplicate IDs. Collect supplied IDs first and generate deterministic IDs that skip occupied values, or formally reserve and validate a generated-ID namespace.

collector/internal/fullsessionrecord/adapter.go
```diff
@@ -0,0 +1,297 @@
...
+		if err != nil {
+			return nil, err
+		}
+		records = append(records, record)
+	}
+	return records, nil
+}
+
+func FromSession(sourceID string, value session.Session) (fullsessionv1.Record, error) {
+	return fromSession(sourceID, "", value)
```
milanshen: **[P1] Preserve parent identity through the adapter**

`ToSession` accepts a valid child record, but the private session carries no `ParentSessionID`, and `FromSession` always restores an empty value. A child record passed through the advertised record-to-session-to-record adapter therefore becomes a root and receives a different revision. Preserve the envelope identity alongside the private model or expose a parent-aware conversion.

collector/internal/session/clone.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+	cloned := *value
+	return &cloned
+}
+
+func cloneSlice[T any](value []T) []T {
+	if value == nil {
+		return nil
+	}
+	return append([]T(nil), value...)
```
milanshen: **[P1] Preserve non-null empty slices in portable clones**

Appending an empty slice to a nil destination returns nil. `FromParsedFamily` uses this clone, so Codex collections deliberately represented as `[]` by the parser accumulators become `null` before freezing; `cloneParsedFamily` repeats the same conversion for child commands. Use a non-nil zero-length destination so portable composition preserves the source collection representation.

collector/internal/collector/collector.go
```diff
@@ -113,25 +113,28 @@ func finalizeSessions(
...
 	composition := composeSessions(parsed)
 	promoteFamilyActivity(composition)
-	enrichSubagents(composition, metadata, claude.WorkflowAgentsSource(source, composition.parsed))
+	enrichSubagents(composition, metadata, claude.WorkflowAgentsSource(source, composition.parsed), preserveSubagentText, useLiveStatus)
 	for _, p := range composition.parsed {
 		removeUnresolvedSpawnRows(p.Session)
 	}
-	resolveNames(composition.roots, metadata)
-	resolveStatus(composition.roots, metadata, source == vendors.LocalReadSource)
+	resolveNames(composition.parsed, metadata)
```
milanshen: **[P1] Resolve descendant names before projecting them**

Name resolution still runs after `enrichSubagents`. When a child's winning name exists only in session metadata, the standalone child record receives that name here, but the parent record has already copied the child ID into both its subagent entry and spawn digest. Resolve all family names before projection and have `subagentFrom` consume the resolved name.

milanshen (commented):
**Carried over:** 22 resolved, 0 closed, 0 unresolved (22 prior findings).

| --- | --- | --- | --- |
| `adapter.go` — preserve unknown costs | P0 | Resolved | Fixed in `44adfb9`: nullable aggregate costs remain preserved. |
| `collector.go` — deterministic timestamps | P1 | Resolved | Fixed in `44adfb9`: portable records require source-derived timing. |
| `collector.go` — shared finalizer | P2 | Resolved | Fixed in `44adfb9`: both paths use the parameterized finalizer. |
| `record.go` — redundant recomputation | P2 | Resolved | Fixed in `44adfb9`: Freeze validates once and performs one final size encoding. |
| `adapter.go` — complete subagent text | P0 | Resolved | Fixed in `f6af405`: portable projection bypasses display truncation. |
| `collector.go` — volatile portable liveness | P0 | Resolved | Fixed in `f6af405`: portable statuses ignore live and wall-clock signals. |
| `README.md` — canonical JSON rules | P1 | Resolved | Fixed in `f6af405`: ordering and escaping are specified with a fixture. |
| `record.go` — caller-owned storage mutation | P1 | Resolved | Fixed in `f6af405`: Freeze deep-clones nested storage. |
| `record.go` — oversized arrays before decode | P0 | Resolved | Fixed in `5d17b48`: token-level limits precede typed decoding. |
| `collector.go` — remote live status | P0 | Resolved | Fixed in `5d17b48`: ListRemote consumes available live metadata. |
| `record.go` — complete child sessions | P0 | Resolved | Fixed in `5d17b48`: every family member receives a linked full record. |
| `adapter.go` — parsed-family mutation | P0 | Resolved | Fixed in `5d17b48`: portable composition operates on a clone. |
| `adapter.go` — opaque change IDs | P1 | Resolved | Fixed in `5d17b48`: supplied IDs survive both adapter directions. |
| `collector.go` — waiting without liveness | P0 | Resolved | Fixed in `55a3ef9`: non-local sources preserve transcript-derived waiting. |
| `record.go` — aggregate nested decode bound | P0 | Resolved | Fixed in `7d6f527`: the token prepass enforces a global item budget. |
| `adapter.go` — null arrays | P1 | Resolved | Fixed in `7d6f527`: nil string collections remain nil. |
| `adapter.go` — usage ordering | P1 | Resolved | Fixed in `7d6f527`: sorted usage is validated and emitted. |
| `measure/main.go` — measurement memory | P1 | Resolved | Fixed in `7d6f527`: reads are bounded and records are measured incrementally. |
| `record.go` — typed aggregate budget | P1 | Resolved | Fixed in `5a35e9c`: typed validation applies the shared aggregate limit. |
| `adapter.go` — per-model costs | P1 | Resolved | Fixed in `5a35e9c`: priced token rows receive model costs. |
| `collector.go` — descendant names | P1 | Resolved | Fixed in `5a35e9c`: name resolution includes every family member. |
| `adapter.go` — empty structural arrays | P1 | Resolved | Fixed in `5a35e9c`: direct adapter round trips preserve non-null empty collections. |

## Review summary

The portable full-session record design is coherent, and all 22 prior findings remain resolved at the updated head. However, shared cost enrichment now introduces an unintended OpenCode UI warning on authoritative costs; four additional record-integrity issues also remain. The measured 28-file, 2,327-line scope exceeded the review budget, so this re-review prioritized every prior finding and changed production paths while assessing tests, documentation, and generated artifacts for adequacy.

**Findings:** 1 blocking (P0), 4 should-fix (P1), 0 nice-to-have (P2) — see inline comments.

Review profile: Milan

Requested in [PR comment](https://github.com/centauri-ai/coslash/pull/203#issuecomment-5720254406)

collector/internal/session/models.go
luhe19001: Updated cost attachment so authoritative recorded totals always carry an empty unpriced-model list, avoiding partial-estimate warnings.

Resolved in `36aac0107f8604d2679df9e212aba848913f62fb`.

collector/fullsession/v1/record.go
luhe19001: Freeze now inventories supplied change IDs first and generates deterministic IDs that skip occupied values.

Resolved in `36aac0107f8604d2679df9e212aba848913f62fb`.

collector/internal/fullsessionrecord/adapter.go
luhe19001: Preserved parent identity as non-JSON private-session state and restored it when converting back to a record.

Resolved in `36aac0107f8604d2679df9e212aba848913f62fb`.

collector/internal/session/clone.go
luhe19001: Updated portable cloning to preserve non-nil empty slices, including parsed child command collections.

Resolved in `36aac0107f8604d2679df9e212aba848913f62fb`.

collector/internal/collector/collector.go
luhe19001: Moved family name resolution before subagent projection and made projections consume the resolved session name.

Resolved in `36aac0107f8604d2679df9e212aba848913f62fb`.

luhe19001:
@agent-review

Resolver pushed `36aac0107f8604d2679df9e212aba848913f62fb` and resolved the addressed review feedback. Please review the updated head.

collector/internal/session/models.go
```diff
@@ -116,17 +124,22 @@ func UnpricedModels(tokens map[string]ModelTokens) []string {
 }
 
 func AttachCost(s *Session, recorded *float64) {
+	for model, used := range s.Tokens {
+		if cost, ok := estimatedModelCost(model, used); ok {
+			used.Cost = cost
```
milanshen: **[P1] Preserve provider-reported model costs**

OpenCode accumulates each provider-reported `message.Cost` into `ModelTokens`, but this assignment replaces every recognized model's recorded value with a LiteLLM estimate before the authoritative aggregate branch returns. An OpenCode session can consequently retain its authoritative total while exposing a conflicting, non-authoritative per-model breakdown.

Only populate estimated model costs when no authoritative cost is present, or track model-cost provenance explicitly.

milanshen (commented):
**Carried over:** 27 resolved, 0 closed, 0 unresolved (27 prior findings).

| --- | --- | --- | --- |
| `adapter.go` — unknown costs | P0 | Resolved | Fixed in `44adfb9da`: nullable aggregate costs survive adapter round trips. |
| `collector.go` — deterministic timestamps | P1 | Resolved | Fixed in `44adfb9da`: portable records require source-derived timing. |
| `collector.go` — shared finalizer | P2 | Resolved | Fixed in `44adfb9da`: display and portable paths share the parameterized finalizer. |
| `record.go` — redundant recomputation | P2 | Resolved | Fixed in `44adfb9da`: Freeze removed the redundant validation/hash pass. |
| `adapter.go` — complete subagent text | P0 | Resolved | Fixed in `f6af405ef`: portable projection preserves full task and result text. |
| `collector.go` — volatile portable liveness | P0 | Resolved | Fixed in `f6af405ef`: portable statuses ignore live and wall-clock signals. |
| `README.md` — canonical JSON rules | P1 | Resolved | Fixed in `f6af405ef`: ordering and escaping are specified with an escaping fixture. |
| `record.go` — caller-owned storage mutation | P1 | Resolved | Fixed in `f6af405ef`: Freeze deep-clones nested record storage. |
| `record.go` — oversized arrays before decode | P0 | Resolved | Fixed in `5d17b48cd`: token-level limits precede typed decoding. |
| `collector.go` — remote live status | P0 | Resolved | Fixed in `5d17b48cd`: ListRemote consumes available live metadata. |
| `record.go` — complete child sessions | P0 | Resolved | Fixed in `5d17b48cd`: every family member receives a linked full record. |
| `adapter.go` — parsed-family mutation | P0 | Resolved | Fixed in `5d17b48cd`: parsed sessions are deeply cloned before composition. |
| `adapter.go` — opaque change IDs | P1 | Resolved | Fixed in `5d17b48cd`: supplied IDs survive both adapter directions. |
| `collector.go` — waiting without liveness | P0 | Resolved | Fixed in `55a3ef9db`: non-local sources preserve transcript-derived waiting. |
| `record.go` — aggregate nested decode bound | P0 | Resolved | Fixed in `7d6f527da`: the token prepass enforces a global item budget. |
| `adapter.go` — null arrays | P1 | Resolved | Fixed in `7d6f527da`: nil string collections remain nil. |
| `adapter.go` — usage ordering | P1 | Resolved | Fixed in `7d6f527da`: sorted usage is validated and emitted. |
| `measure/main.go` — measurement memory | P1 | Resolved | Fixed in `7d6f527da`: reads are bounded and records are measured incrementally. |
| `record.go` — typed aggregate budget | P1 | Resolved | Fixed in `5a35e9cea`: typed validation applies the shared aggregate limit. |
| `adapter.go` — per-model costs | P1 | Resolved | Fixed in `5a35e9cea`: normal priced token rows receive model costs. |
| `collector.go` — descendant record names | P1 | Resolved | Fixed in `5a35e9cea`: name resolution covers every family member. |
| `adapter.go` — empty structural arrays | P1 | Resolved | Fixed in `5a35e9cea`: adapter round trips preserve non-null empty collections. |
| `models.go` — authoritative-cost warning | P0 | Resolved | Fixed in `36aac0107`: recorded totals now carry an empty unpriced-model list. |
| `record.go` — generated-ID collision | P1 | Resolved | Fixed in `36aac0107`: generated IDs skip every supplied ID. |
| `adapter.go` — parent identity | P1 | Resolved | Fixed in `36aac0107`: parentSessionId is retained through the private session. |
| `clone.go` — portable empty slices | P1 | Resolved | Fixed in `36aac0107`: clones preserve non-null zero-length slices. |
| `collector.go` — names before projection | P1 | Resolved | Fixed in `36aac0107`: names resolve before subagent and digest projection. |

## Review summary

The portable-record approach is coherent, and all 27 prior findings remain resolved. Three should-fix correctness and interoperability gaps remain. The 31-file, 2,397-line scope exceeded the nominal budget, so the re-review prioritized all prior findings and changed production paths across four parallel passes; tests, documentation, and generated artifacts were assessed for adequacy.

**Findings:** 0 blocking (P0), 3 should-fix (P1), 2 nice-to-have (P2) — see inline comments.

Review profile: Milan

Requested in [PR comment](https://github.com/centauri-ai/coslash/pull/203#issuecomment-5720478013)

collector/internal/fullsessionrecord/adapter.go
```diff
@@ -0,0 +1,298 @@
...
+			cost := *item.RecordedCost
+			copy.RecordedCost = &cost
+		}
+		cloned = append(cloned, &copy)
+	}
+	return cloned
+}
+
+func micros(value float64) int64  { return int64(math.Round(value * 1_000_000)) }
+func dollars(value int64) float64 { return float64(value) / 1_000_000 }
```
milanshen, line 264: **[P1] Keep micro-USD values exact through the adapter**

The record permits nonnegative `int64` costs, but converting them through `float64` is lossy above 2^53. For example, a valid `9_007_199_254_740_993` micro-USD value returns as `9_007_199_254_740_992`, changing the canonical record and revision; values near MaxInt64 can become invalid on conversion back.

Preserve integer micro-USD state in the private representation or validate and document an exactly representable upper bound.

collector/fullsession/v1/README.md
```diff
@@ -0,0 +1,87 @@
...
+revision field. Each change body independently declares its UTF-8 byte count
+and SHA-256.
+
+Every family member has its own complete record. `parentSessionId` is empty for
+a root and identifies the direct parent for a child, preserving descendants
+without reducing child sessions to the display-oriented `subagents` list.
+
+## Canonical JSON
```
milanshen, line 23: **[P1] Specify the complete record-validity contract**

The language-neutral contract defines canonical encoding but omits the 64 MiB record limit, text/string limits, aggregate item budget, identifier and agent constraints, and nonnegative-count requirements enforced by Go. A non-Go consumer can therefore accept a 100,001-item or oversized record that the reference decoder rejects, making cache validity implementation-dependent.

Document the normative validation rules and publish representative boundary fixtures.

collector/internal/fullsessionrecord/adapter.go
```diff
@@ -0,0 +1,298 @@
...
+	"github.com/centauri-ai/coslash/collector/internal/collector"
+	"github.com/centauri-ai/coslash/collector/internal/session"
+	"github.com/centauri-ai/coslash/collector/internal/vendors"
+)
+
+// FromParsedFamily runs portable composition on a clone before freezing one
+// complete record for every rooted family member.
+func FromParsedFamily(sourceID, vendor string, source vendors.ReadSource, parsed []*vendors.ParsedSession, metadata *vendors.SessionMetadata) ([]fullsessionv1.Record, error) {
+	composed := collector.ComposePortable(source, map[string]vendors.RemoteCollection{
+		vendor: {Sessions: cloneParsedFamily(parsed), Metadata: metadata},
```
milanshen, line 19: **[P2] Clone metadata before portable enrichment**

The parsed family is cloned, but caller-owned metadata is passed through unchanged. When metadata supplies `Usage.Tokens`, enrichment aliases that map into the clone and `AttachCost` writes estimated costs back into it, so record composition mutates reusable metadata and concurrent consumers can race.

Clone metadata token maps before composition, or copy enrichment values when applying them.

collector/fullsession/v1/record.go
```diff
@@ -0,0 +1,572 @@
...
+			if len(stack) >= MaxItems {
+				return fmt.Errorf("%w: collection nesting exceeds item limit", ErrInvalid)
+			}
+			stack = append(stack, container{kind: delim})
+		}
+	}
+}
+
+func Validate(record Record) error {
+	if err := validate(record, true); err != nil {
```
milanshen, line 353: **[P2] Enforce the record byte limit in Validate**

`Validate` does not enforce `MaxRecordBytes`. A correctly hashed record containing several individually legal change bodies can exceed 64 MiB and pass this function—and therefore `ToSession`—while `Marshal` and `Decode` reject the same record as oversized.

Apply the byte limit in full validation, or make `ToSession` use a validation path that includes serialized size.

collector/internal/session/models.go
luhe19001: Updated AttachCost to estimate per-model costs only when no authoritative aggregate cost exists, preserving provider-reported model costs. Added regression coverage.

Resolved in `5f010c7a846daae5160b030cd85a5d5041fd2ea2`.

collector/internal/fullsessionrecord/adapter.go
luhe19001: Added and enforced a documented maximum micro-USD value guaranteed to round-trip exactly through the float64 adapter, with maximum-boundary round-trip coverage.

Resolved in `5f010c7a846daae5160b030cd85a5d5041fd2ea2`.

collector/fullsession/v1/README.md
luhe19001: Documented the complete normative validity contract and published valid/invalid fixtures covering identifier, cost, timestamp, item-budget, and negative-count boundaries.

Resolved in `5f010c7a846daae5160b030cd85a5d5041fd2ea2`.

collector/internal/fullsessionrecord/adapter.go
luhe19001: Metadata token maps are now copied during enrichment, preventing portable composition from mutating caller-owned metadata. Added regression coverage.

Resolved in `5f010c7a846daae5160b030cd85a5d5041fd2ea2`.

collector/fullsession/v1/record.go
luhe19001: Validate now rejects canonical records exceeding MaxRecordBytes, so ToSession enforces the same limit as Marshal and Decode. Added an oversized-record regression test.

Resolved in `5f010c7a846daae5160b030cd85a5d5041fd2ea2`.

luhe19001:
@agent-review

Resolver pushed `5f010c7a846daae5160b030cd85a5d5041fd2ea2` and resolved the addressed review feedback. Please review the updated head.

collector/fullsession/v1/record.go
```diff
@@ -0,0 +1,586 @@
...
+		return err
+	}
+	canonical, err := json.Marshal(record)
```
milanshen, line 356: **[P2] Reuse canonical bytes during validation**

This added marshal creates and discards a complete canonical buffer; `Validate` then marshals the revision preimage, while both `Marshal` and `Decode` immediately marshal the final record again. A valid record near 64 MiB therefore incurs three full JSON encodes per operation. Have the internal validation path return the canonical bytes for callers to reuse, or derive the final size from the preimage plus the fixed 64-byte revision value.

milanshen (commented):
**Carried over:** 32 resolved, 0 closed, 0 unresolved (32 prior findings).

| --- | --- | --- | --- |
| `adapter.go` — unknown costs | P0 | Resolved | Fixed in `44adfb9da3353a34e2e0a6d9483fd0269088b67b`: nullable costs survive round trips. |
| `collector.go` — deterministic timestamps | P1 | Resolved | Fixed in `44adfb9da3353a34e2e0a6d9483fd0269088b67b`: portable composition rejects missing source timing. |
| `collector.go` — duplicated finalizer | P2 | Resolved | Fixed in `44adfb9da3353a34e2e0a6d9483fd0269088b67b`: paths share a parameterized finalizer. |
| `record.go` — redundant Freeze work | P2 | Resolved | Fixed in `44adfb9da3353a34e2e0a6d9483fd0269088b67b`: redundant post-derivation validation was removed. |
| `adapter.go` — truncated subagent text | P0 | Resolved | Fixed in `f6af405ef2e7df0494707228a77cce49f8ab1e26`: portable projection preserves complete text. |
| `collector.go` — volatile portable liveness | P0 | Resolved | Fixed in `f6af405ef2e7df0494707228a77cce49f8ab1e26`: portable statuses ignore live and wall-clock signals. |
| `README.md` — canonical JSON rules | P1 | Resolved | Fixed in `f6af405ef2e7df0494707228a77cce49f8ab1e26`: ordering and escaping are specified with a fixture. |
| `record.go` — caller storage mutation | P1 | Resolved | Fixed in `f6af405ef2e7df0494707228a77cce49f8ab1e26`: Freeze deep-clones nested storage. |
| `record.go` — oversized arrays before decode | P0 | Resolved | Fixed in `5d17b48cda1b821bc68878e0ab28d279dbb764da`: token-level limits precede typed decoding. |
| `collector.go` — remote live status | P0 | Resolved | Fixed in `5d17b48cda1b821bc68878e0ab28d279dbb764da`: display composition consumes live metadata. |
| `record.go` — incomplete child sessions | P0 | Resolved | Fixed in `5d17b48cda1b821bc68878e0ab28d279dbb764da`: each family member receives a linked complete record. |
| `adapter.go` — parsed-family mutation | P0 | Resolved | Fixed in `5d17b48cda1b821bc68878e0ab28d279dbb764da`: portable composition operates on a deep clone. |
| `adapter.go` — opaque change IDs | P1 | Resolved | Fixed in `5d17b48cda1b821bc68878e0ab28d279dbb764da`: supplied IDs survive both adapter directions. |
| `collector.go` — waiting without liveness | P0 | Resolved | Fixed in `55a3ef9db18ef2522697c2f8ccba5c33327ac8a0`: non-local sources preserve parsed waiting state. |
| `record.go` — aggregate nested decode bound | P0 | Resolved | Fixed in `7d6f527da4b6b3e389ad2053e89debc98e5188ff`: the token prepass enforces a global budget. |
| `adapter.go` — null arrays | P1 | Resolved | Fixed in `7d6f527da4b6b3e389ad2053e89debc98e5188ff`: nil string collections remain nil. |
| `adapter.go` — usage ordering | P1 | Resolved | Fixed in `7d6f527da4b6b3e389ad2053e89debc98e5188ff`: ascending model order is validated and emitted. |
| `measure/main.go` — measurement memory | P1 | Resolved | Fixed in `7d6f527da4b6b3e389ad2053e89debc98e5188ff`: reads are bounded and bodies are not retained. |
| `record.go` — typed aggregate budget | P1 | Resolved | Fixed in `5a35e9ceae87e3eef2a020dd9e8c43b69983951b`: typed validation shares the aggregate limit. |
| `adapter.go` — per-model estimated costs | P1 | Resolved | Fixed in `5a35e9ceae87e3eef2a020dd9e8c43b69983951b`: priced token rows receive costs. |
| `collector.go` — descendant names | P1 | Resolved | Fixed in `5a35e9ceae87e3eef2a020dd9e8c43b69983951b`: name resolution covers every family member. |
| `adapter.go` — empty structural arrays | P1 | Resolved | Fixed in `5a35e9ceae87e3eef2a020dd9e8c43b69983951b`: non-null empty collections survive round trips. |
| `models.go` — authoritative-cost warning | P0 | Resolved | Fixed in `36aac0107f8604d2679df9e212aba848913f62fb`: recorded totals carry no partial-estimate warning. |
| `record.go` — generated-ID collision | P1 | Resolved | Fixed in `36aac0107f8604d2679df9e212aba848913f62fb`: generated IDs skip supplied IDs. |
| `adapter.go` — parent identity | P1 | Resolved | Fixed in `36aac0107f8604d2679df9e212aba848913f62fb`: parent identity is retained in private state. |
| `clone.go` — portable empty slices | P1 | Resolved | Fixed in `36aac0107f8604d2679df9e212aba848913f62fb`: clones preserve non-null empty slices. |
| `collector.go` — names before projection | P1 | Resolved | Fixed in `36aac0107f8604d2679df9e212aba848913f62fb`: names resolve before subagent/digest projection. |
| `models.go` — provider model costs | P1 | Resolved | Fixed in `5f010c7a846daae5160b030cd85a5d5041fd2ea2`: authoritative per-model costs are no longer overwritten. |
| `adapter.go` — exact micro-USD values | P1 | Resolved | Fixed in `5f010c7a846daae5160b030cd85a5d5041fd2ea2`: an exact round-trip maximum is enforced. |
| `README.md` — complete validity contract | P1 | Resolved | Fixed in `5f010c7a846daae5160b030cd85a5d5041fd2ea2`: normative limits and boundary fixtures are documented. |
| `adapter.go` — metadata mutation | P2 | Resolved | Fixed in `5f010c7a846daae5160b030cd85a5d5041fd2ea2`: enrichment copies metadata token maps. |
| `record.go` — Validate byte limit | P2 | Resolved | Fixed in `5f010c7a846daae5160b030cd85a5d5041fd2ea2`: Validate enforces `MaxRecordBytes`. |

## Review summary

The portable full-record design and adapter direction are sound, and all prior scenarios remain resolved at the updated head. The scope exceeded the nominal budget; because parallel reviewers were unavailable, this used a single-pass fallback covering all changed production paths and prior findings, while tests, fixtures, and documentation were assessed for adequacy rather than line-by-line bug hunting. Only one minor efficiency cleanup remains.

**Findings:** 0 blocking (P0), 0 should-fix (P1), 1 nice-to-have (P2) — see inline comments.

Review profile: Milan

Requested in [PR comment](https://github.com/centauri-ai/coslash/pull/203#issuecomment-5720870992)

milanshen (approved):

PR #204, opened 2026-09-17
Add agent-facing coSlash CLI commands

calvintvu:
## Context

coSlash needs a stable command interface that coding agents can invoke without handling loopback credentials themselves. This adds the local-session CLI foundation for session discovery, canonical handoffs, target-agent launches, agent-selected reviews, and offline diagnostics.

## Changes

- Add compact JSON session search, canonical Markdown handoff output, Claude/Codex launch commands, and Claude/Codex/OpenCode review launches.
- Reuse the existing review API so session resolution, reviewer availability, prompt generation, naming, duplicate prevention, and launch behavior remain backend-owned.
- Discover the running app’s custom loopback port and rotating token automatically while preserving authenticated API access.
- Keep the commands local-session only and return actionable errors when the app is stopped or input is invalid.
- Document the command and loopback API contracts.

## Test

- `cd collector && make test` — passed
- `cd collector && make build` — passed
- Live CLI smoke test:

<details>
<summary>Command output</summary>

```console
$ coslash sessions cvu/coslash-plugin-claude-codex --json
[{"id":"<session-id>","name":"Define Coslash agent skills","agent":"codex","status":"busy","repo":"github.com/centauri-ai/coslash","branch":"cvu/coslash-plugin-claude-codex","mtime":<timestamp>}]

$ coslash handoff <session-id>
# Handoff — readme

## Objective (first prompt)
...

## Current state
...

## Key decisions
- —

## Timeline
...

## Files
- —

## Commits
- —

## Next steps
- —

## Environment
- Vendor: Codex
- Repository: github.com/centauri-ai/coslash
- Branch: main
...

$ coslash send <session-id> --to codex
Success: started codex for session <session-id>

$ coslash review <session-id> --with codex
Success: started codex review for session <session-id>

$ coslash sessions 'Review — Sessions skill validation' --json
[{"id":"<review-session-id>","name":"Review — Sessions skill validation (<session-prefix>)","agent":"codex","status":"busy","repo":"github.com/centauri-ai/coslash","branch":"cvu/coslash-plugin-claude-codex","mtime":<timestamp>}]

$ coslash review <session-id> --with codex
Error: review already running
[exit 1]

$ coslash doctor --json
{"version":"<version>","platform":{"os":"darwin","arch":"arm64","terminalLaunchSupported":true},"storage":{"writable":true},"sources":[{"agent":"claude","state":"ok"},{"agent":"codex","state":"ok"},{"agent":"opencode","state":"ok"}],"checks":[{"id":"source.claude","status":"ok"},{"id":"source.codex","status":"ok"},{"id":"source.opencode","status":"ok"},{"id":"storage","status":"ok"},{"id":"synthesis","status":"ok"},{"id":"opencode.plugin","status":"ok"}]}
```

</details>

Local paths, session identifiers, timestamps, and lengthy handoff content are abbreviated for the public PR.


collector/internal/launch/launch.go
```diff
@@ -173,15 +184,19 @@ func cliCommand(agent, sessionID, mode, handoff string) (string, string, error)
...
 	switch agent {
 	case vendors.AgentClaude:
 		path, err := writeHandoffFile(context)
 		if err != nil {
 			return "", "", err
 		}
-		return withCleanup(shellJoin(cli, "--append-system-prompt-file", path), path), path, nil
+		arguments := []string{cli, "--append-system-prompt-file", path}
+		if prompt != "" {
+			arguments = append(arguments, prompt)
```
milanshen: **[P0] Terminate option parsing before appending the prompt**

`runSend` accepts arbitrary message text, but this appends it directly after the agent's options. Shell quoting does not stop CLI option parsing: for example, `coslash send <id> --to claude --dangerously-skip-permissions` passes a safety-disabling Claude option instead of that literal prompt; Codex has analogous options. This changes both behavior and security without an intentional flag-forwarding contract.

Insert `--` before the positional prompt for both Claude and Codex, including the prompt-only path, or use an explicit mechanism that guarantees literal prompt delivery.

milanshen (commented):
## Review summary

This PR adds authenticated local CLI commands and server-generated session handoffs. The overall structure is coherent, but initial task text must be forwarded literally instead of being interpreted as agent CLI options. Several request-lifecycle and error-boundary issues also need attention.

**Findings:** 1 blocking (P0), 4 should-fix (P1), 0 nice-to-have (P2) — see inline comments.

Review profile: Milan

collector/cmd/coslash/cli.go
```diff
@@ -0,0 +1,258 @@
...
+
+func newLocalAPIClient() (*localAPIClient, error) {
+	baseURL, token, err := readRuntime()
+	if err != nil {
+		return nil, err
+	}
+	return &localAPIClient{
+		baseURL: baseURL,
+		token:   token,
+		client:  &http.Client{Timeout: 30 * time.Second},
```
milanshen, line 115: **[P1] Do not let a timed-out send launch later**

A targeted `send` currently runs `collector.List(0)` across every historical session. If that exceeds this 30-second deadline, the client reports that coSlash is unreachable, but the server does not honor request cancellation and can subsequently open the terminal. Retrying then launches duplicate agents even though the first invocation appeared to fail.

Load only the requested session family, and check the request context before performing the launch side effect.

collector/cmd/coslash/api.go
```diff
@@ -348,6 +350,97 @@ func handleLaunch(w http.ResponseWriter, r *http.Request, settingsStore *setting
...
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+	if err := open(
+		state.Config.Launch.Terminal,
+		target,
+		found.WorkingDirectory,
+		found.ID,
+		launch.NewSession,
+		handoffcontext.Build(found),
```
milanshen: **[P1] Enforce the handoff size limit on generated content**

This passes the generated handoff directly to the launcher, bypassing the existing `launch.MaxHandoffBytes` check in `readHandoff`. Long sessions can exceed 64 KiB because digest recaps and list counts are unbounded; sufficiently large Codex handoffs can then exceed argument limits after command substitution, while the API has already reported a successful terminal launch.

Build the handoff once, reject it when its byte length exceeds `MaxHandoffBytes`, and only then call the launcher.

collector/cmd/coslash/main.go
```diff
@@ -130,6 +134,9 @@ func main() {
 		log.Printf("write API token: %v", err)
 	}
 	baseURL := "http://" + listener.Addr().String()
+	if err := writeRuntime(baseURL); err != nil {
```
milanshen, line 149: **[P1] Bind runtime discovery to the owning process**

The token and runtime descriptor are separate global files with no ownership or generation binding. If instance A is running and instance B starts on another supported port, B overwrites both files; after B exits, every data command targets B's stale endpoint even though A remains healthy. Concurrent startups can also interleave the two writes into a mismatched endpoint/token pair.

Define a singleton instance or store generation/owner metadata and conditionally restore or remove only the descriptor owned by the exiting process.

collector/cmd/coslash/api.go
```diff
@@ -348,6 +350,97 @@ func handleLaunch(w http.ResponseWriter, r *http.Request, settingsStore *setting
...
+	open promptLauncher,
+) {
+	target := r.URL.Query().Get("to")
+	if target != vendors.AgentClaude && target != vendors.AgentCodex {
+		http.Error(w, "target must be claude or codex", http.StatusBadRequest)
+		return
+	}
+	state := settingsStore.State()
+	if !state.Valid {
+		http.Error(w, state.Error+"; open Settings to repair it", http.StatusConflict)
```
milanshen: **[P1] Keep raw settings errors out of API responses**

This returns `state.Error` verbatim. Read failures include `os.PathError` text and can expose an absolute home path. This directly violates `collector/CLAUDE.md`: “Keep internal errors in server logs and return fixed, non-sensitive messages to clients.”

Log `state.Error` and return a fixed message such as `settings are invalid; open Settings to repair them`.

collector/internal/launch/launch.go
calvintvu: Fixed in b19a3f2. `cliCommandWithPrompt` now inserts `--` before flag-shaped prompts for both prompt-only and handoff-backed Claude/Codex launches. The regression covers both agents and both paths.

collector/cmd/coslash/cli.go
calvintvu: Fixed in b19a3f2. Canonical resolution now uses `collector.GetSessionForPreview` to load only the requested family, and `handleSend` checks the request context immediately before opening the terminal. The regression verifies that a canceled request cannot launch an agent.

collector/cmd/coslash/api.go
calvintvu: Fixed in b19a3f2. `handleSend` logs the internal settings error and returns the fixed client message `settings are invalid; open Settings to repair them`. The regression verifies that decoder details are not returned.

collector/cmd/coslash/api.go
calvintvu: Fixed in b19a3f2. `handleSend` builds the canonical handoff once, rejects content larger than `launch.MaxHandoffBytes` with HTTP 413, and never calls the launcher in that case. The oversized-handoff regression covers the boundary.

collector/cmd/coslash/main.go
calvintvu: Fixed in b19a3f2. The server now acquires a non-blocking runtime lock scoped to `COSLASH_HOME` before writing the token or runtime descriptor, so another instance cannot replace the owning process’s discovery files. The lock regression verifies exclusion and release.

calvintvu:
@agent-reviewer

milanshen:
@agent-review

milanshen:

collector/internal/launch/launch.go
milanshen: **[P0 — RESOLVED]** Fixed in `e882cffbc4681f0557462533bd12d71fb496be48`: prompt-only and handoff-backed Claude/Codex commands now insert `--` before the prompt, so flag-shaped task text remains positional.

collector/cmd/coslash/cli.go
milanshen: **[P1 — RESOLVED]** Fixed in `e882cffbc4681f0557462533bd12d71fb496be48`: send resolution now loads only the requested session family, and `handleSend` checks request cancellation immediately before opening the terminal.

collector/cmd/coslash/api.go
milanshen: **[P1 — RESOLVED]** Fixed in `e882cffbc4681f0557462533bd12d71fb496be48`: `handleSend` builds the handoff once, rejects content over `MaxHandoffBytes` with 413, and does not invoke the launcher.

collector/cmd/coslash/main.go
milanshen: **[P1 — RESOLVED]** Fixed in `e882cffbc4681f0557462533bd12d71fb496be48`: startup acquires a nonblocking lock scoped to `COSLASH_HOME` and holds it for the server lifetime, preventing concurrent owners from replacing discovery files.

collector/cmd/coslash/api.go
milanshen: **[P1 — RESOLVED]** Fixed in `e882cffbc4681f0557462533bd12d71fb496be48`: the internal settings error is logged while `/api/send` returns a fixed, non-sensitive repair message.

collector/cmd/coslash/api.go
```diff
@@ -353,6 +355,103 @@ func handleLaunch(w http.ResponseWriter, r *http.Request, settingsStore *setting
...
+
+func handleSend(
+	w http.ResponseWriter,
+	r *http.Request,
+	settingsStore *settings.Store,
+	getSession func(string) (*session.Session, error),
+	open promptLauncher,
+) {
+	target := r.URL.Query().Get("to")
+	if target != vendors.AgentClaude && target != vendors.AgentCodex {
```
milanshen, line 415: **[P1] Verify the target CLI before reporting success**

This validates the target name but never checks whether its executable is available. If a user with only Claude installed sends a session to Codex, AppleScript successfully opens the terminal, the API returns 204, and the CLI prints `Success: started codex` even though the shell only reports `codex: command not found`.

Check target availability before opening the terminal and return a fixed, actionable error when it is missing.

collector/cmd/coslash/cli.go
```diff
@@ -0,0 +1,297 @@
...
+	}
+	reviewer := args[2]
+	if reviewer != "claude" && reviewer != "codex" && reviewer != "opencode" {
+		return fmt.Errorf("--with must be claude, codex, or opencode")
+	}
+	client, err := newLocalAPIClient()
+	if err != nil {
+		return err
+	}
+	path := "/api/reviews?source=local&id=" + url.QueryEscape(args[0]) + "&reviewer=" + url.QueryEscape(reviewer)
```
milanshen: **[P1] Do not start reviews after the client gives up**

The new client has a 30-second timeout, but `handleReview` does not check `r.Context()` after loading the session and before calling `startReview`. If loading a large family crosses that timeout—or the user interrupts the command—the CLI reports failure while the server subsequently starts the reviewer anyway.

Check request cancellation immediately before `startReview`, as `handleSend` already does, and propagate cancellation into loading where possible.

collector/cmd/coslash/main.go
```diff
@@ -133,6 +141,9 @@ func main() {
...
```
milanshen, line 149: **[P1] Require successful CLI discovery publication**

Both `writeToken` and this runtime-descriptor write merely log failures, although every new data command requires both files. For example, if an existing `runtime.lock` remains openable but the coSlash directory becomes unwritable or the disk is full, the app continues running with missing or stale discovery data and every CLI command fails or targets a stale endpoint.

Publish the token and descriptor as a required unit and abort startup before advertising the server when either write fails.

collector/cmd/coslash/cli.go
```diff
@@ -0,0 +1,297 @@
...
+		}
+	}
+	if !jsonOutput {
+		return fmt.Errorf("usage: coslash sessions [query] --json")
+	}
+	client, err := newLocalAPIClient()
+	if err != nil {
+		return err
+	}
+	data, err := client.request(http.MethodGet, "/api/sessions", nil)
```
milanshen, line 197: **[P1] Avoid timing out while fetching discarded session details**

This requests unrestricted `/api/sessions`, which runs `collector.List(0)` over the complete archive and serializes full session records. Filtering and projection happen only after the response arrives, so a large history can exceed the fixed 30-second timeout and be reported as “app is not reachable” even though the server is healthy.

Use a lightweight server-side listing contract with filtering or pagination, or otherwise give this operation a timeout appropriate to full-history collection.

milanshen (commented):
## Review summary

This PR adds authenticated CLI access to session listing, handoffs, agent launches, and reviews. The overall design is coherent, but several failure and timeout paths can still produce misleading CLI outcomes or unintended work.

**Prior findings:** 5 resolved, 0 closed, 0 open.
**This review:** 5 resolved, 0 closed, 0 reopened, 4 new.

Review profile: Milan

Requested in [PR comment](https://github.com/centauri-ai/coslash/pull/204#issuecomment-5724354525)

collector/cmd/coslash/api.go
calvintvu: Agree. I verified that handleSend validated the target name but TerminalWithPrompt only checked terminal availability, so a missing agent CLI could still produce a false success. Fixed in 680700c by checking the existing executable-discovery path before loading or launching and returning a fixed 400 response. The regression verifies that an unavailable target cannot launch.

collector/cmd/coslash/cli.go
calvintvu: Agree. I verified that handleReview called startReview after session loading without rechecking the request context, unlike handleSend. Fixed in 680700c by checking cancellation immediately before the side effect. The regression verifies that a canceled request cannot start a review.

collector/cmd/coslash/main.go
calvintvu: Agree. I verified that token and runtime publication errors were only logged, allowing startup to continue with unusable or stale discovery data. Fixed in 680700c by making either publication failure fatal before the app logs or opens its access URL and before the server starts serving.

collector/cmd/coslash/cli.go
calvintvu: Partially agree. The 30-second generic timeout is too short for collector.List(0), but filtering or pagination layered over the current collector would still parse the full archive because there is no lightweight index. Fixed the misleading timeout in 680700c with a sessions-only three-minute timeout matching the server full-history write limit, while retaining the short timeout for handoff and side-effecting commands.

calvintvu:
@agent-review

collector/cmd/coslash/api.go
luhe19001: **[P1 — RESOLVED]** Fixed in `680700cbb652dd25519ca58a5e9f4478708e18fb`: `/api/send` now verifies target executable availability before loading the session or opening a terminal, returning a fixed error when unavailable.

collector/cmd/coslash/cli.go
luhe19001: **[P1 — RESOLVED]** Fixed in `680700cbb652dd25519ca58a5e9f4478708e18fb`: `handleReview` now checks request cancellation immediately before calling `startReview`.

collector/cmd/coslash/main.go
luhe19001: **[P1 — RESOLVED]** Fixed in `680700cbb652dd25519ca58a5e9f4478708e18fb`: token and runtime-descriptor publication failures are now fatal before the server starts serving or advertises its access URL.

collector/cmd/coslash/cli.go
luhe19001: **[P1 — RESOLVED]** Fixed in `680700cbb652dd25519ca58a5e9f4478708e18fb`: session listing now uses a dedicated three-minute client timeout aligned with the server's full-history response window.

collector/cmd/coslash/main.go
```diff
@@ -194,6 +205,9 @@ func routes(
 ) *http.ServeMux {
 	mux := http.NewServeMux()
 	api := http.NewServeMux()
+	getCanonicalSession := func(id string) (*session.Session, error) {
+		return canonicalSession(id, mgr, collector.GetSessionForPreview)
```
luhe19001: **[P1] Preserve agent identity when resolving CLI sessions**

Session identity is agent-qualified—the collector tests explicitly support Claude and Codex sessions sharing an ID—but handoff, send, and CLI-triggered review transmit only the bare ID. `GetSessionForPreview` then returns the first matching vendor, so selecting a Codex session whose ID also exists in Claude can print the wrong transcript or launch work in the wrong directory.

Add an agent-qualified session selector to these CLI/API requests and resolve it with `GetSessionForPreviewByAgent`.

collector/cmd/coslash/api.go
```diff
@@ -353,6 +355,108 @@ func handleLaunch(w http.ResponseWriter, r *http.Request, settingsStore *setting
...
+	}
+	handoff := handoffcontext.Build(found)
+	if len(handoff) > launch.MaxHandoffBytes {
+		http.Error(w, "handoff context is too large", http.StatusRequestEntityTooLarge)
+		return
+	}
+	if r.Context().Err() != nil {
+		return
+	}
+	if err := open(
```
luhe19001, line 457: **[P1] Cancel an in-progress terminal launch**

The request context is checked before `open`, but the launcher itself is context-free and invokes `osascript` with `exec.Command`. If an automation permission prompt or unresponsive terminal blocks for more than the client's 30-second timeout, the CLI reports that coSlash is unreachable while the server continues waiting; approving the prompt later can still launch the agent, and a retry can create a duplicate session.

Pass the request context through the launcher and use a cancellation-aware subprocess invocation.

luhe19001 (commented):
## Review summary

This PR adds authenticated CLI access for sessions, handoffs, agent launches, and reviews. The approach is coherent, but session identity is ambiguous across agents and an in-progress terminal launch can outlive the client's deadline.

**Findings:** 0 blocking (P0), 2 should-fix (P1), 0 nice-to-have (P2) — see inline comments.

**Prior findings:** 9 resolved, 0 closed, 0 open.
**This review:** 4 resolved, 0 closed, 0 reopened, 2 new.

Review profile: Helu

Requested in [PR comment](https://github.com/centauri-ai/coslash/pull/204#issuecomment-5732672372)

collector/cmd/coslash/main.go
calvintvu: Agree. The collector already permits different agents to share a session ID, while handoff, send, and CLI review dropped the agent before lookup. Fixed in 2b0e643: sessions now emit an agent-qualified selector, all three CLI commands require and transmit it, and the APIs resolve through GetSessionForPreviewByAgent. Focused tests cover selector parsing and agent propagation, and the README documents the updated contract.

collector/cmd/coslash/api.go
calvintvu: Agree. The preflight context check did not cover a blocked AppleScript subprocess because the shared launch path used exec.Command without a context. Fixed in 2b0e643 by carrying the request context through local and remote terminal launch helpers and invoking osascript with exec.CommandContext. The regression verifies cancellation reaches the AppleScript runner.

calvintvu:
@agent-review

collector/cmd/coslash/main.go
luhe19001: **[P1 — STILL OPEN]** `2b0e643549ab45fc1b68a38202359dda5f78a158` qualifies the base-session lookup, but review still calls `mgr.LookupLatest(found.ID)` and the synthesis cache is keyed only by bare ID. Same-ID Claude and Codex sessions can therefore attach the other agent's cached outcome, so identity is not preserved end-to-end.

collector/cmd/coslash/api.go
luhe19001: **[P1 — RESOLVED]** Fixed in `2b0e643549ab45fc1b68a38202359dda5f78a158`: the request context now flows through local and remote terminal helpers into the macOS adapter, where `osascript` is launched with `exec.CommandContext`.

collector/cmd/coslash/api.go
```diff
@@ -353,6 +355,119 @@ func handleLaunch(w http.ResponseWriter, r *http.Request, settingsStore *setting
...
```
luhe19001, line 457: **[P1] Validate the working directory before launching**

An archived session can retain a nonempty working-directory path after that directory has been deleted, renamed, or replaced by a file. `TerminalWithPrompt` only checks that the string is nonempty, while AppleScript successfully queues `cd '<cwd>' && <agent>`; the `cd` then fails inside the terminal, but the API returns 204 and the CLI reports that the agent started.

Verify immediately before launch that the path exists and is a directory, and return a fixed actionable error when it is unavailable.

luhe19001 (commented):
## Review summary

This PR adds authenticated local CLI workflows for session discovery, handoff, send, and review. Most of the hardening is sound, but one prior identity boundary remains incomplete and one launch-readiness edge case still needs correction.

**Prior findings:** 10 resolved, 0 closed, 1 open.
**This review:** 1 resolved, 0 closed, 0 reopened, 1 new.

Review profile: Helu

Requested in [PR comment](https://github.com/centauri-ai/coslash/pull/204#issuecomment-5733163712)

collector/cmd/coslash/main.go
calvintvu: Agree. Agent-qualified API lookup was incomplete because synthesis cache storage, in-flight work, cooldown/failure state, and LookupLatest were still keyed by bare ID. Fixed in e8a4145: all synthesis state is keyed by agent and ID, and review, list, and canonical lookups pass both. The regression runs same-ID Claude and Codex synthesis and verifies distinct outcomes.

collector/cmd/coslash/api.go
calvintvu: Agree. A nonempty archived working-directory path can be missing or name a file, and TerminalWithPrompt previously passed it through. Fixed in e8a4145: the shared local launch boundary now requires an existing directory, and send returns an actionable 409 response. Regressions cover both missing and file paths plus the API response.

calvintvu:
@agent-review

collector/cmd/coslash/main.go
milanshen: **[P1 — RESOLVED]** Fixed in `e8a4145ba5742292624ea10b4e271526cad1e3aa`: synthesis cache storage, in-flight work, failure state, and CLI review lookups are now keyed by both agent and session ID.

collector/cmd/coslash/api.go
milanshen: **[P1 — RESOLVED]** Fixed in `e8a4145ba5742292624ea10b4e271526cad1e3aa`: local launches now verify that the archived working path exists and is a directory before creating the command or opening a terminal.

collector/cmd/coslash/api.go
```diff
@@ -244,12 +246,12 @@ func handleSynthesis(w http.ResponseWriter, id string, mgr *synthesis.Manager) {
 	}{}
 	revision := found.LastActivityTime
 	if revision > 0 {
-		response.Synthesis = mgr.Lookup(found.ID, revision)
+		response.Synthesis = mgr.Lookup(found.Agent, found.ID, revision)
```
milanshen, line 253: **[P1] Keep synthesis requests agent-qualified**

`handleSynthesis` still receives only an ID and calls `GetSessionFacts(id)`, while the frontend requests `/api/synthesis?id=...`. If Claude and Codex share an ID, selecting the later vendor loads, looks up, and potentially generates synthesis for the first vendor instead.

Include the agent in this API contract and resolve facts by `(agent, id)`.

collector/internal/synthesis/cache.go
```diff
@@ -22,36 +23,44 @@ type Cache struct {
...
 
-func (c *Cache) Load(id string) (Record, error) {
-	if value, ok := c.records.Load(id); ok {
+func (c *Cache) Load(agent, id string) (Record, error) {
+	key := cacheKey{agent: agent, id: id}
+	if value, ok := c.records.Load(key); ok {
 		return value.(Record), nil
 	}
-	data, err := os.ReadFile(c.recordPath(id))
+	data, err := os.ReadFile(c.recordPath(agent, id))
```
milanshen, line 107: **[P1] Migrate existing synthesis cache entries**

Existing installations store records at `summaries/<id>.json`, but this now reads only `summaries/<agent>/<id>.json`. After upgrading, cached summaries disappear; inactive historical sessions are not swept, while recent sessions can consume additional metered synthesis work regenerating unchanged results.

Migrate legacy entries whose IDs map uniquely to one agent, while discarding ambiguous collisions.

collector/internal/handoff/handoff.go
```diff
@@ -0,0 +1,193 @@
...
+	lines = append(lines, "# Handoff — "+title, "", "## Objective ("+goalSource+")")
+	lines = appendListOrText(lines, goals)
+	lines = append(lines, "", "## Current state", current, "", "## Key decisions")
+	lines = appendBullets(lines, decisions)
+	lines = append(lines, "", "## Timeline")
+	if len(value.Digest) == 0 {
+		lines = append(lines, "- —")
+	} else {
+		for _, entry := range value.Digest {
+			lines = append(lines, fmt.Sprintf("- [%s · turn %d] %s", entry.Category, entry.Turn, entry.Description))
```
milanshen, line 41: **[P1] Preserve recorded answers in handoffs**

Question digest entries carry `DigestEntry.Answer`, and the inspector displays it, but this renderer includes only the question. After an `AskUserQuestion` or `request_user_input` interaction, the successor can miss the user's selected answer and repeat or contradict that decision.

Render every nonempty answer with its question.

collector/internal/launch/launch.go
```diff
@@ -144,15 +147,23 @@ func (buffer *boundedBuffer) Write(data []byte) (int, error) {
...
+}
+
+func TerminalWithPrompt(ctx context.Context, terminal, agent, workingDirectory, sessionID, mode, handoff, prompt string) error {
 	if workingDirectory == "" {
 		return fmt.Errorf("launch: session has no working directory")
 	}
-	command, handoffPath, err := cliCommand(agent, sessionID, mode, handoff)
+	info, err := os.Stat(workingDirectory)
+	if err != nil || !info.IsDir() {
+		return ErrWorkingDirectoryUnavailable
```
milanshen, line 160: **[P1] Normalize unavailable-directory failures**

This sentinel is handled by `/api/send`, but `/api/launch` maps it to a generic 500. Empty working directories also return a different error, so `send` reports those as internal failures too. Deleted, file-valued, and empty archived paths therefore produce inconsistent and non-actionable responses.

Return one unavailable-directory sentinel for every unusable path and map it consistently to an actionable conflict response in both handlers.

collector/internal/handoff/handoff.go
```diff
@@ -0,0 +1,193 @@
...
+	if value.Name != nil {
+		title = *value.Name
+	}
+	goals, goalSource := sessionGoals(value)
+	current := "—"
+	if value.Summary != nil {
+		current = *value.Summary
+	}
+	if value.Synthesis != nil {
+		current = value.Synthesis.Outcome
```
milanshen: **[P2] Retain the summary when synthesis has no outcome**

Synthesis normalization permits a goals-only result with an empty outcome. In that case this assignment replaces a usable session summary with an empty current-state section.

Only prefer `Synthesis.Outcome` when it is nonblank.

milanshen (commented):
## Review summary

This PR adds authenticated CLI workflows and strengthens launch and session identity handling. The core direction is sound, but synthesis identity remains incomplete, existing caches are abandoned on upgrade, and generated handoffs can omit important context.

**Findings:** 0 blocking (P0), 4 should-fix (P1), 1 nice-to-have (P2) — see inline comments.

**Prior findings:** 12 resolved, 0 closed, 0 deferred, 0 open.
**This review:** 2 resolved, 0 closed, 0 deferred, 0 reopened, 5 new.

Review profile: Milan

Requested in [PR comment](https://github.com/centauri-ai/coslash/pull/204#issuecomment-5733788111)

collector/cmd/coslash/api.go
calvintvu: Agree. The frontend and route still used a bare ID, so the newly agent-qualified cache could still load or generate synthesis for the wrong vendor. Fixed in 6045612: requests now carry the agent, the route requires it, and the collector facts lookup filters by agent. Frontend, route, and collector regressions cover the contract.

collector/internal/synthesis/cache.go
calvintvu: Agree. Moving storage without migration made valid flat cache records unreachable, and legacy records cannot identify their own agent. Fixed in 6045612: startup migration resolves each legacy ID against supported agents, moves unique matches, discards ambiguous entries, and preserves records on resolver or I/O failure. Regressions cover unique and ambiguous IDs.

collector/internal/handoff/handoff.go
calvintvu: Agree. Digest answers were retained in session data and shown by the inspector, but the handoff renderer omitted them. Fixed in 6045612 by rendering each nonblank answer beneath its timeline question, with focused regression coverage.

collector/internal/launch/launch.go
calvintvu: Agree. Empty paths returned a different error, and /api/launch did not map the unavailable-directory sentinel. Fixed in 6045612: empty local and remote paths use the same sentinel, and both launch handlers share the actionable HTTP 409 mapping. The path and API regressions cover the behavior.

collector/internal/handoff/handoff.go
calvintvu: Agree. A goals-only synthesis can have a blank outcome, so unconditional preference erased a usable summary. Fixed in 6045612 by preferring only a nonblank trimmed synthesis outcome. The regression verifies the summary remains the current state.

calvintvu:
@agent-review

collector/cmd/coslash/api.go
milanshen: **[P1 — RESOLVED]** Fixed in `6045612340727a40356095741eb72f6534e54ecf`: synthesis requests now carry the agent through the frontend route, API handler, collector lookup, and cache operations.

collector/internal/synthesis/cache.go
milanshen: **[P1 — RESOLVED]** Fixed in `6045612340727a40356095741eb72f6534e54ecf`: startup migrates uniquely attributable flat cache records into agent-qualified storage and discards ambiguous collisions.

collector/internal/handoff/handoff.go
milanshen: **[P1 — RESOLVED]** Fixed in `6045612340727a40356095741eb72f6534e54ecf`: every nonblank digest answer is now rendered beneath its timeline question.

collector/internal/launch/launch.go
milanshen: **[P1 — RESOLVED]** Fixed in `6045612340727a40356095741eb72f6534e54ecf`: empty local and remote paths now return the shared unavailable-directory sentinel, which both launch handlers map to an actionable 409 response.

collector/internal/handoff/handoff.go
milanshen: **[P2 — RESOLVED]** Fixed in `6045612340727a40356095741eb72f6534e54ecf`: handoffs now prefer synthesis outcomes only when they are nonblank, preserving the existing summary otherwise.

milanshen (commented):
## Review summary

This PR adds authenticated CLI workflows, canonical handoffs, cancellation-aware launches, and agent-qualified synthesis caching. The direction is sound, but the review workflow still lacks one launch precondition. The exact diff was reviewed using the single-pass fallback because parallel reviewers were unavailable.

**Prior findings:** 17 resolved, 0 closed, 0 deferred, 0 open.
**This review:** 5 resolved, 0 closed, 0 deferred, 0 reopened, 1 new.

Review profile: Milan

Requested in [PR comment](https://github.com/centauri-ai/coslash/pull/204#issuecomment-5734294320)

collector/cmd/coslash/api.go
```diff
@@ -432,6 +559,9 @@ func handleReview(
 	}
 	name := reviewpkg.Name(originName, found.ID)
 	prompt := reviewpkg.Prompt(found)
+	if r.Context().Err() != nil {
```
milanshen, line 558: **[P1] Validate the review directory before accepting the request**

For an archived session whose nonempty working-directory path was deleted or replaced by a file, this proceeds to `startReview`, returns 202, and makes `coslash review` print success. The background `exec.CommandContext` then fails when applying `command.Dir`, so no review actually starts.

Check immediately before `startReview` that the path exists and is a directory, and return an actionable conflict response; preferably reuse the launch-path validation used by send.

calvintvu: Agreed. handleReview accepted any nonempty path and returned 202 before the background reviewer applied command.Dir. Fixed in 01a9277: review requests now reuse the shared local working-directory validator immediately before startReview, return an actionable 409 for empty, missing, or file paths, and launch.Review enforces the same boundary. Focused regressions cover the false-success response and process boundary.

calvintvu:
@agent-review

luhe19001:
Review is waiting for merge conflicts to be resolved.

collector/cmd/coslash/api.go
milanshen: **[P1 — RESOLVED]** Fixed in `01a92773d626df67b740a72fb38e7813838606d3`: `handleReview` now validates the archived path before `startReview`, maps empty, missing, and non-directory paths to the actionable 409 response, and `launch.Review` enforces the same boundary.

collector/cmd/coslash/cli.go
milanshen: **[P1 — REOPENED]** `e3239dd44c7fbc7f7c05da22ae42abdbee1c8a8e` inserts the context-free `os.Stat` validation between the cancellation check and `startReview`. If a working directory is on a slow or unavailable mounted filesystem and the request times out during that stat, the detached review still starts after stat eventually succeeds.

calvintvu:
@agent-review

luhe19001 (commented):
## Review summary

This PR adds authenticated CLI handoff, send, and review workflows plus agent-qualified synthesis caching. The overall direction is sound, but discovery trust, redirect handling, cache migration behavior, and divergent handoff renderers still need attention; one prior cancellation race also remains open.

**Findings:** 0 blocking (P0), 5 should-fix (P1), 0 nice-to-have (P2) — see inline comments.

**Prior findings:** 17 resolved, 0 closed, 0 deferred, 1 open.
**This review:** 0 resolved, 0 closed, 0 deferred, 0 reopened, 5 new.

Review profile: Helu

Requested in [PR comment](https://github.com/centauri-ai/coslash/pull/204#issuecomment-5736582612)

collector/cmd/coslash/cli.go
```diff
@@ -0,0 +1,329 @@
...
+
+func newLocalAPIClient() (*localAPIClient, error) {
+	baseURL, token, err := readRuntime()
+	if err != nil {
+		return nil, err
+	}
+	return &localAPIClient{
+		baseURL: baseURL,
+		token:   token,
+		client:  &http.Client{Timeout: 30 * time.Second},
```
luhe19001: **[P1] Keep CLI redirects on loopback**

The default HTTP client follows redirects. Go forwards the custom `X-Coslash-Token` header across hosts, and a 307/308 also replays `send`'s `bytes.Buffer` task body. If the recorded loopback listener redirects to an external URL, both the token and task text leave the machine.

Reject redirects, or revalidate every redirect destination and never forward credentials or bodies outside the original loopback endpoint.

collector/internal/synthesis/cache.go
```diff
@@ -22,36 +26,108 @@ type Cache struct {
...
+				if err := cache.Store(matches[0], id, record); err != nil {
+					failures = append(failures, fmt.Errorf("migrate legacy synthesis %q: %w", id, err))
+					continue
+				}
+			} else if err != nil {
+				failures = append(failures, fmt.Errorf("inspect synthesis destination %q: %w", id, err))
+				continue
+			}
+		}
+		if err := os.Remove(legacyPath); err != nil {
```
luhe19001, line 98: **[P1] Preserve unresolved legacy cache entries**

A zero-match migration falls through to this unconditional removal. Missing transcript roots produce a successful empty scan, and inaccessible subtrees can be skipped without returning an error, so upgrading while a source is temporarily unavailable permanently deletes otherwise valid cached synthesis.

Keep zero-match entries for a later retry; remove only after a successful migration or when multiple matches prove the entry ambiguous.

collector/cmd/coslash/cli.go
```diff
@@ -0,0 +1,329 @@
...
+	}
+	if err := temporary.Close(); err != nil {
+		return err
+	}
+	return os.Rename(temporaryPath, filepath.Join(home, runtimeFilename))
+}
+
+func readRuntime() (string, string, error) {
+	home := settings.Home()
+	data, err := os.ReadFile(filepath.Join(home, runtimeFilename))
```
luhe19001, line 143: **[P1] Reject stale runtime discovery data**

`runtime.json` and the token remain after the server exits, and `readRuntime` does not verify that `runtime.lock` is currently owned. If another process later binds the recorded port and returns 2xx, `send` and `review` print success although nothing launched, and `send` delivers its task text to that unrelated process.

Verify the live lock/owner before trusting the descriptor and conditionally remove discovery files when the owning server shuts down.

collector/cmd/coslash/main.go
```diff
@@ -99,6 +107,11 @@ func main() {
 	if err := synthesis.EnsureDirs(); err != nil {
 		log.Printf("initialize synthesis cache: %v", err)
 		mgr.SetRunner(nil)
+	} else if err := synthesis.MigrateLegacyCache(func(agent, id string) (bool, error) {
```
luhe19001, line 112: **[P1] Resolve migration identities with one source scan**

For every legacy summary, migration invokes this callback up to three times. Claude and Codex implement `GetSessionFactsByAgent` by enumerating their complete transcript tree first, making upgrade startup O(summary count × transcript count) before the listener is created. Large histories can therefore make the app appear not to start for minutes.

Build an agent-qualified ID index once per source, or provide a direct indexed existence lookup, and migrate against that snapshot.

collector/internal/handoff/handoff.go
```diff
@@ -0,0 +1,196 @@
...
+	lines = appendListOrText(lines, goals)
+	lines = append(lines, "", "## Current state", current, "", "## Key decisions")
+	lines = appendBullets(lines, decisions)
+	lines = append(lines, "", "## Timeline")
+	if len(value.Digest) == 0 {
+		lines = append(lines, "- —")
+	} else {
+		for _, entry := range value.Digest {
+			lines = append(lines, fmt.Sprintf("- [%s · turn %d] %s", entry.Category, entry.Turn, entry.Description))
+			if answer := strings.TrimSpace(entry.Answer); answer != "" {
```
luhe19001, line 42: **[P1] Keep UI handoffs aligned with the canonical renderer**

The new Go renderer preserves question answers and ignores blank synthesis outcomes, but the existing frontend `handoffBrief` still drops `DigestEntry.answer` and prefers an empty outcome via `??`. Consequently CLI/API handoffs retain decisions such as “Postgres,” while UI Copy and Start fresh omit them or erase the summary for the same session.

Update the frontend renderer to match these rules, preferably routing UI handoffs through the canonical implementation so the formats cannot drift again.

collector/cmd/coslash/cli.go
calvintvu: Agree. Fixed in d28f1f8f. The CLI HTTP client now rejects redirects, so the token and task body remain on the original loopback request. The regression invokes coslash send against a 307 response and verifies that the redirect target receives nothing.

collector/internal/synthesis/cache.go
calvintvu: Agree. Fixed in d28f1f8f. Legacy entries with zero identity matches are now retained for a later startup retry. Only a successful unique migration or a proven multi-agent collision removes the flat entry. The regression verifies the unresolved file remains.

collector/cmd/coslash/cli.go
calvintvu: Agree. Fixed in d28f1f8f. Runtime discovery is trusted only while another file descriptor owns runtime.lock, and normal shutdown removes runtime.json and token only when both still match the current server. Regressions cover stale discovery rejection and conditional cleanup.

collector/cmd/coslash/main.go
calvintvu: Agree. Fixed in d28f1f8f. Migration now lazily builds one agent-qualified ID index by scanning each source once, then resolves every legacy entry from that snapshot. The regression verifies one scan per source and excludes child sessions.

collector/internal/handoff/handoff.go
calvintvu: Partially agree. Fixed the behavioral divergence in d28f1f8f: the frontend now ignores blank synthesis outcomes and renders nonblank digest answers exactly like the Go handoff renderer. I did not route the UI through the local handoff endpoint because this renderer also handles remote session details, while that endpoint accepts only local sessions. The focused frontend regression covers both rules.

collector/cmd/coslash/cli.go
calvintvu: Agree. Fixed in d28f1f8f. handleReview retains the early cancellation check and now checks the request context again after working-directory validation, immediately before startReview. A timeout during os.Stat can no longer start a detached review.

calvintvu:
@agent-review

collector/cmd/coslash/cli.go
luhe19001: **[P1 — RESOLVED]** Fixed in `d28f1f8f7358a42141c3d9db920968981d0b901b`: `handleReview` now checks request cancellation again after working-directory validation and immediately before `startReview`.

collector/cmd/coslash/cli.go
luhe19001: **[P1 — RESOLVED]** Fixed in `d28f1f8f7358a42141c3d9db920968981d0b901b`: the CLI HTTP client now returns `http.ErrUseLastResponse` for redirects, preventing credential and request-body forwarding.

collector/internal/synthesis/cache.go
luhe19001: **[P1 — RESOLVED]** Fixed in `d28f1f8f7358a42141c3d9db920968981d0b901b`: legacy entries with zero identity matches are retained, while only successful or provably ambiguous migrations remove them.

collector/cmd/coslash/cli.go
luhe19001: **[P1 — STILL OPEN]** After an unclean exit, a restart acquires `runtime.lock` before migration but does not replace the stale descriptor and token until later. During that window `readRuntime` sees a held lock and trusts the previous endpoint, so `send` can still reach an unrelated process that reused the old port. Clear stale discovery when acquiring the lock or bind the descriptor to the lock owner's generation.

collector/cmd/coslash/main.go
luhe19001: **[P1 — RESOLVED]** Fixed in `d28f1f8f7358a42141c3d9db920968981d0b901b`: `SessionIDIndex` scans each source once, and `sync.OnceValues` reuses that snapshot for every legacy entry.

collector/internal/handoff/handoff.go
luhe19001: **[P1 — RESOLVED]** Fixed in `d28f1f8f7358a42141c3d9db920968981d0b901b`: the frontend renderer now trims and preserves digest answers and falls back to the session summary for blank synthesis outcomes.

luhe19001 (commented):
### Review 8 Summary

**Score: 4/5** — coverage incomplete

The endpoint delta improves cancellation, redirects, cache migration, runtime discovery, and handoff rendering. Runtime discovery remains vulnerable to stale data during restart because the lock and descriptor share no generation identity. The restricted delta and affected callers were inspected via a single-pass fallback, but unknown original coverage keeps overall coverage partial.

**Prior findings:** 22 resolved, 0 closed, 0 deferred, 1 open.
**This review:** 5 resolved, 0 closed, 0 deferred, 0 reopened, 0 new.

Review profile: Helu

Requested in [PR comment](https://github.com/centauri-ai/coslash/pull/204#issuecomment-5736873533)

collector/cmd/coslash/cli.go
calvintvu: Agree. Fixed in 7731cee9. Exclusive runtime-lock acquisition now removes any previous runtime descriptor and token before startup migration begins, so CLI commands cannot trust the old endpoint during the restart window. The end-to-end regression acquires the restart lock over stale discovery, invokes coslash sessions, and verifies that the stale endpoint receives no request. The complete Go suite and go vet pass on the current rebased PR head.

calvintvu:
@agent-review

collector/cmd/coslash/cli.go
luhe19001: **[P1 — STILL OPEN]** `7731cee96a3f09eb31dc90b6295544e168e0a2b3` narrows but does not eliminate the race: after `Flock` succeeds and before the first `os.Remove`, a concurrent `readRuntime` sees an active owner and can still load the stale descriptor and token. The regression invokes the CLI only after `acquireRuntimeLock` returns, so it does not cover this interleaving. Bind discovery to the owner's generation or add a readiness state that readers validate.

luhe19001 (commented):
### Review 9 Summary

The new cleanup substantially narrows the stale-discovery window. However, a CLI can still read stale credentials after the lock is acquired but before cleanup runs, so the prior P1 remains open. No new delta findings were identified.

**Prior findings:** 22 resolved, 0 closed, 0 deferred, 1 open.
**This review:** 0 resolved, 0 closed, 0 deferred, 0 reopened, 0 new.

**Scope:** prior findings and new changes only; coverage complete.



Review profile: Helu

Requested in [PR comment](https://github.com/centauri-ai/coslash/pull/204#issuecomment-5737230897)

collector/cmd/coslash/cli.go
calvintvu: Agree. Fixed in 76c2e386. The runtime lock now encodes readiness: startup keeps it exclusive, which CLI readers reject, and the server downgrades it to shared only immediately before Serve. This closes the interval between lock acquisition and stale-file removal without adding a separate readiness file or generation. The end-to-end regression recreates stale discovery while the startup lock is exclusive, invokes coslash sessions, and verifies that the old endpoint receives no request. The complete Go suite and go vet pass.

calvintvu:
@agent-review

collector/cmd/coslash/cli.go
luhe19001: **[P1 — RESOLVED]** Fixed in `76c2e386d30d66442b6448d2f5229f15f4dff7a2`: startup now retains the exclusive lock through stale-discovery cleanup and publication, while CLI readers reject discovery until the owner marks it ready, eliminating the original restart window.

collector/cmd/coslash/cli.go
```diff
@@ -0,0 +1,394 @@
...
+		if err := os.Remove(filepath.Join(home, name)); err != nil && !errors.Is(err, os.ErrNotExist) {
+			file.Close()
+			return nil, fmt.Errorf("clear stale runtime discovery: %w", err)
+		}
+	}
+	return file, nil
+}
+
+func markRuntimeReady(file *os.File) error {
+	return unix.Flock(int(file.Fd()), unix.LOCK_SH)
```
luhe19001: **[P1] Preserve singleton ownership while publishing readiness**

An exclusive-to-shared `flock` conversion is not guaranteed to be atomic: exclusive ownership can be released before the shared lock is acquired. If a second coSlash process starts during that gap, it can acquire the exclusive lock and remove the first process's token and runtime descriptor. On the default port it then fails to bind and exits, leaving the first server running but undiscoverable; with distinct ports, both processes can serve concurrently.

Keep the lifetime ownership lock exclusive and publish readiness through a separate generation or state that readers validate.

**Change evidence:** Since `7731cee96a3f`, `collector/cmd/coslash/cli.go:54` (RIGHT): This line newly downgrades the lifetime singleton lock to shared. The previous head retained the exclusive lock until process exit, so a second startup could not acquire it.

luhe19001 (commented):
### Review 10 Summary

**Active findings:** 0 blocking (P0), 1 should-fix (P1), 0 nice-to-have (P2) — new and unresolved prior findings.

The readiness gate resolves the previously reported stale-endpoint window during startup. One new concurrency defect remains in the readiness mechanism.


**Prior findings:** 23 resolved, 0 closed, 0 deferred, 0 open.
**This review:** 1 resolved, 0 closed, 0 deferred, 0 reopened, 1 new.

**Scope:** prior findings and new changes only; coverage complete.



Review profile: Helu

Requested in [PR comment](https://github.com/centauri-ai/coslash/pull/204#issuecomment-5737537228)

collector/cmd/coslash/cli.go
calvintvu: Verified: flock(2) conversion releases the existing lock before applying the new mode, so the singleton gap is real. Fixed in 851ffe66 by retaining the exclusive runtime.lock for the server lifetime and publishing readiness through a separate runtime.ready lock. Added regression coverage that fails on the parent implementation and confirms readiness publication cannot release singleton ownership.

calvintvu:
@agent-review

collector/cmd/coslash/cli.go
luhe19001: **[P1 — RESOLVED]** Fixed in `851ffe6693bacb7773fd6c9388af53546657eecf`: readiness now uses a separate exclusively held `runtime.ready` lock while `runtime.lock` remains exclusive for the server lifetime, so publishing readiness no longer releases singleton ownership.

luhe19001 (commented):
### Review 11 Summary

The follow-up separates lifetime ownership from readiness, eliminating the lock-conversion gap while keeping discovery unavailable during startup. The prior concurrency defect is fixed, and no new issues were confirmed.


**Prior findings:** 24 resolved, 0 closed, 0 deferred, 0 open.
**This review:** 1 resolved, 0 closed, 0 deferred, 0 reopened, 0 new.

**Scope:** prior findings and new changes only; coverage complete.



Review profile: Helu

Requested in [PR comment](https://github.com/centauri-ai/coslash/pull/204#issuecomment-5737745686)

milanshen (approved):

PR #214, opened 2026-09-18
Exclude old live sessions from recent time windows

calvintvu:
## Context

Live sessions bypassed the selected time cutoff, which allowed sessions with old activity timestamps to appear under filters such as “This week.” Recent time windows should reflect recorded session activity, while “All” should remain unconstrained.

## Changes

- Apply named time windows strictly to each session’s last activity time, regardless of live status.
- Keep the “All” window unconstrained and add regression coverage for old live sessions.
- Explicitly exclude local review state from portable full-session records, restoring the collector inventory check.

## Test

- `npm test` - passed, 116 tests
- `make release` - passed
- `make test` - passed
- `make check` - passed
- Manual: resumed a September 9 OpenCode session while filtering by “This week”; verified it appeared before the fix and was excluded afterward.

### Screenshots

- [x] OpenCode “This week” before - old live session is incorrectly visible
eng-1429-browser-before

- [x] OpenCode “This week” after - old live session is excluded
eng-1429-browser-after


calvintvu:
Closing because the confirmed product behavior keeps genuinely running sessions visible even when their last activity falls outside the selected time window. ENG-1378 addresses the stale OpenCode status that caused the original false-positive observation.

PR #215, opened 2026-09-18
Clear stale OpenCode turn status

calvintvu:
## Context

OpenCode transcripts can retain historical assistant messages without completion timestamps or with running tool state after later turns finish. Treating those old records as current can keep completed sessions Active without a live TUI and surface them in recent activity windows.

## Changes

- Derive busy and waiting status from the current OpenCode turn instead of accumulating state across the full transcript.
- Clear superseded question and task state while preserving historical task metadata.
- Keep current incomplete responses Active and current unanswered questions Waiting.

## Test

- `go test ./... -count=1` - passed, 500 tests
- `make check` - passed
- Manual: with an isolated August transcript, verified unchanged `main` showed the completed session as Active in This week; the fixed build excluded it from This week and retained it as inactive history under All.

### Screenshots

Before:

Old completed OpenCode session incorrectly shown Active in This week

After:

Same session excluded from This week after stale status cleared


collector/internal/vendors/opencode/parse.go
```diff
@@ -82,7 +83,9 @@ func parse(tx *sql.Tx, row storedSession) (parsedSession, error) {
 			if prompt == "" {
 				continue
 			}
+			busy = false
```
luhe19001: **[P1] Keep a newly submitted turn busy**

When a session has an older incomplete assistant message and then persists a new real user prompt, collection can occur before the new assistant row is created. This assignment clears the only busy state, so `parse` returns `InTurn=false` with no `StatusHint`. Non-TUI clients such as `opencode run` have no external live metadata to correct that result, leaving an unanswered prompt without a busy status; the base code retained busy for this interrupted/resumed sequence.

Set `busy = true` when a non-empty user prompt opens a turn, while still clearing `waiting` and `activeTasks`, and let the following assistant message close or refine the state. Add a test whose transcript ends at the second user message.

luhe19001 (commented):
## Review summary

This PR scopes OpenCode status to the latest meaningful turn and adds focused parser coverage. The approach is generally sound, but the user-boundary reset can temporarily misclassify an active non-TUI session and should be corrected before merge.

Review profile: Helu

collector/internal/vendors/opencode/parse.go
calvintvu: Agreed and fixed in 2cf076a. A non-empty user prompt now opens a busy current turn while still clearing superseded waiting and task state. The regression includes an older incomplete assistant record followed by a real user text part with no assistant row; it failed with InTurn=false before the fix and now passes. The full collector suite passes with 498 tests, and go vet remains clean.

calvintvu:
@agent-review

collector/internal/vendors/opencode/parse.go
luhe19001: **[P1 — RESOLVED]** Fixed in `2cf076a8eb06a6ff4df92318755691faf8bfed78`: a non-empty user prompt now sets `busy = true`, and the added transcript-ending regression verifies `InTurn` with a `busy` hint before an assistant row exists.

collector/internal/vendors/opencode/parse.go
```diff
@@ -182,6 +188,7 @@ func parse(tx *sql.Tx, row storedSession) (parsedSession, error) {
 						}
 						link.status = part.State.Status
 						tasks[childID] = link
+						activeTasks[childID] = part.State.Status
```
luhe19001, line 199: **[P1] Settle spawns when task state is superseded**

`activeTasks` is cleared at a newer user or assistant boundary, but the corresponding historical `tasks` and `spawns` entries remain `running` and `Completed=false`. If that task has a child with a completed transcript and the parent later completes another turn, the root becomes inactive, but `source.go` still assigns the child a spawn key and `subagentStatus` consequently reports the finished child as running indefinitely.

When active task state is superseded, also settle its spawn/task linkage as returned or aborted, or make composition honor a terminal child transcript. Add coverage that composes the parent with its completed child.

luhe19001 (commented):
## Review summary

This PR scopes OpenCode persisted status to the latest meaningful turn, and the follow-up correctly repairs the previously reported persistence gap. The approach is sound overall, but lifecycle cleanup remains incomplete at two status boundaries; the child-composition defect should be fixed before merge.

**Findings:** 0 blocking (P0), 1 should-fix (P1), 1 nice-to-have (P2) — see inline comments.

**Prior findings:** 1 resolved, 0 closed, 0 deferred, 0 open.
**This review:** 1 resolved, 0 closed, 0 deferred, 0 reopened, 2 new.

Review profile: Helu

Requested in [PR comment](https://github.com/centauri-ai/coslash/pull/215#issuecomment-5736560975)

collector/internal/vendors/opencode/parse.go
```diff
@@ -82,7 +83,9 @@ func parse(tx *sql.Tx, row storedSession) (parsedSession, error) {
 			if prompt == "" {
 				continue
 			}
+			busy = true
```
luhe19001, line 91: **[P2] Expire abandoned prompt-only busy hints**

If OpenCode persists the user row and is killed before creating its assistant row, this sets a durable `StatusHint="busy"` that no subsequent message clears. Non-TUI sessions lack matching live metadata, and `resolveStatus` does not age persisted hints, so the abandoned session remains busy across later collections.

Bound prompt-only busy hints by liveness or recency, or consume a terminal process marker, while preserving the intended transient busy state before the assistant row appears.

calvintvu: Partially agreed and addressed in 54476ff. The durable prompt-only hint is possible, but non-TUI runs expose no terminal marker or authoritative process liveness, so the collector cannot prove abandonment. The fix narrowly expires only the prompt-before-assistant phase after two minutes; assistant and tool-derived current-turn status remains unchanged. Tests cover both a recent prompt remaining busy and an old prompt becoming inactive.

collector/internal/vendors/opencode/parse.go
calvintvu: Agreed and fixed in 54476ff. Tasks absent from the current assistant state now settle their retained spawn metadata as completed. Parser coverage verifies the superseded spawn flag, and composition coverage verifies a terminal OpenCode child renders as returned rather than running.

calvintvu:
@agent-review

collector/internal/vendors/opencode/parse.go
luhe19001: **[P1 — RESOLVED]** Fixed in `54476ffd6f010e4b2a59b9b387fcdfb038c1e758`: tasks absent from the current assistant state now have their retained spawn marked completed, so composition returns a terminal child instead of reporting it as running.

collector/internal/vendors/opencode/parse.go
luhe19001: **[P2 — RESOLVED]** Fixed in `54476ffd6f010e4b2a59b9b387fcdfb038c1e758`: the parser now identifies the prompt-before-assistant phase and clears its busy hint after two minutes, while preserving busy status for recent prompts.

luhe19001 (commented):
### Review 3 Summary

**Score: 5/5** — coverage incomplete

The endpoint delta correctly settles superseded OpenCode task spawns and expires abandoned prompt-only busy hints. Both outstanding findings are demonstrably fixed, and no new issues were found in the authorized delta or affected composition path. Original coverage was unknown, so this restricted round cannot certify the entire PR.

**Prior findings:** 3 resolved, 0 closed, 0 deferred, 0 open.
**This review:** 2 resolved, 0 closed, 0 deferred, 0 reopened, 0 new.

Review profile: Helu

Requested in [PR comment](https://github.com/centauri-ai/coslash/pull/215#issuecomment-5736835439)

milanshen (approved):

PR #225, opened 2026-09-20
Fix Codex compaction counts

calvintvu:
## Context

Codex rollouts record compactions as top-level `compacted` rows. coSlash only recognized the legacy nested event format, leaving compaction counts and timeline entries at zero.

## Changes

- count current top-level Codex compaction rows
- share compaction bookkeeping with the legacy event format
- add a regression test covering sequential compactions

## Test

- `cd collector && go test ./...` - 679 tests passed
- `cd collector && make check` - passed
- Manual: verified an isolated two-compaction rollout in the table, inspector, and timeline with no browser console errors

#### Before

To be uploaded.

#### After

To be uploaded.


luhe19001 (commented):
### Review 1 Summary

This PR correctly adds support for top-level Codex compaction records while preserving legacy handling through a shared helper. The focused regression test validates the intended count behavior.


Review profile: Helu

luhe19001 (approved):

PR #226, opened 2026-09-20
Revert "Refresh the model table from LiteLLM"

milanshen:
Reverts centauri-ai/coslash#224

Undesirably deleted 4 repo-owned Composer model overrides, causing [CI failures](https://github.com/centauri-ai/coslash/actions/runs/35540857882/job/106158113319?pr=212) on main


collector/internal/session/models.json
```diff
@@ -15907,7 +15898,7 @@
     "output_cost_per_token": 0.000002016
   },
   "replicate/google/gemini-2.5-flash": {
-    "input_cost_per_token": 3E-7,
+    "input_cost_per_token": 0.0000025,
```
luhe19001, line 15901: **[P1] Keep the corrected Gemini input rate**

This changes `replicate/google/gemini-2.5-flash` input pricing from $0.30/M tokens back to $2.50/M—the output-token rate. Consequently, a session with one million input tokens is estimated at $2.50 instead of $0.30, an 8.3× overestimate.

Keep the refreshed `3E-7` input rate while selectively restoring the intended custom model entries.

collector/internal/session/models.json
```diff
@@ -17520,14 +17511,12 @@
     "output_cost_per_token": 3E-7
   },
   "vercel_ai_gateway/google/gemini-2.5-flash": {
-    "cache_read_input_token_cost": 3E-8,
```
luhe19001, line 17523: **[P1] Preserve the Gemini cache-read prices**

Removing these cache rates makes `loadModels` substitute the full input rate. One million cached Vercel Gemini 2.5 Flash tokens therefore rise from $0.03 to $0.30, while Pro rises from $0.125 to $2.50 because its ordinary input rate is also doubled here.

Retain the refreshed cache-read fields and Pro input rate instead of reverting the complete generated table.

collector/internal/session/models.json
```diff
@@ -11829,12 +11854,6 @@
     "max_input_tokens": 1000000,
     "output_cost_per_token": 3.84e-06
   },
-  "opencode/deepseek-v4.1-flash": {
```
luhe19001, line 11832: **[P1] Retain the newly imported OpenCode model rows**

The collector records OpenCode usage as `provider/model`, so deleting this row and the new `jev-1.13{,-free}` and `qwen3.8-flash` rows leaves sessions using those current gateway models unmatched. For example, `opencode/deepseek-v4.1-flash` has no unqualified fallback, causing its cost to be excluded and its known one-million-token context window to disappear.

Preserve the newly fetched OpenCode rows and revert only the catalog entries that caused the original refresh problem.

luhe19001 (commented):
### Review 1 Summary

**Active findings:** 0 blocking (P0), 3 should-fix (P1), 0 nice-to-have (P2) — new and unresolved prior findings.

The revert restores manually maintained Composer pricing, but also rolls back unrelated current catalog data. Preserve the refreshed provider rows and rates while restoring only the intended custom entries.

Review profile: Helu

luhe19001 (approved):

PR #216, opened 2026-09-18
Add full-session v2 collector upload support

luhe19001:
## Context

Complete cached SSH sessions need a bounded collector path that can preview and upload the exact immutable revision accepted by Hub. This adds the collector and local API boundary without changing existing v1 sharing behavior.

## Changes

- serialize one validated full-session record into the canonical v2 envelope with exact byte count, hash, and repository binding
- expose loopback preview and share routes that verify pairing and Hub capability before loading complete content
- bind consent to the revision, bytes, destination, repository, and audience while preserving the reviewed request and idempotency key across retryable failures
- upload one bounded gzip request and reconcile ambiguous outcomes without creating a second revision

## Test

- `cd collector && go test ./...` — passed
- `cd collector && go vet ./...` — passed
- `git diff --check` — passed
- canonical 2,620-byte client/server request fixture comparison — passed

PR #212, opened 2026-09-18
Rebuild the session board with configurable grouping, and fix remote host status reporting

milanshen:
## Context

The session board was a fixed matrix with no way to change how sessions were organized. Remote machine health also used inconsistent dots, messages, banners, and retry behavior, while headers intended to remain visible did not actually pin inside long result lists.

## Changes

1. **Configurable board grouping**
   - Add persisted **Columns** and **Rows** controls covering status, readiness, repository, branch, agent, and machine, including `Rows: None`.
   - Replace the fixed matrix with collapsible full-width row bands, sparse nonempty intersections, pinned labels and totals, and stable per-dimension collapse state.
   - Keep repositories distinct by full identity while showing concise labels and full paths on hover.
   - Clamp long titles, withhold review when no workspace exists, and retain the selected-card, remote-machine, readiness, cost, and subagent treatments.

2. **Consistent remote machine status**
   - Treat backend state as authoritative: connection errors are red and bannered, stale or incomplete data is amber, truncated-but-complete history is green, and disabled collection is gray.
   - Keep the status dot informational and inert; retry remains available in its tooltip, the banner, and Settings according to the machine state.
   - Make Settings consume the terminal result of a retry so its private status immediately matches the board.

3. **Pinned result headers**
   - Constrain the result pane so table headers, board columns, and row bands remain visible while scrolling.
   - Measure wrapped table-header height so section bands pin directly below it.

## Test

- `npm test` — passed, 161 tests across 25 files
- `npm run build` — passed
- `npm run lint` — passed with four pre-existing React Compiler warnings in `CoslashPage.tsx`
- `npm run format:check` — passed
- `make release` — passed with the embedded frontend and helper assets
- `make smoke` — passed against the embedded release binary
- Manual: restarted the embedded backend, loaded 87 sessions in Table and Board, exercised Status/Repository and Branch/Branch grouping, confirmed the high-cardinality board rendered 15 sparse cells with no console errors, and verified Settings transitions through retry to its terminal machine status

### Screenshots

Before publication, take screenshots and generate sanitized images that contain no private session titles, summaries, paths, costs, token counts, credentials, or internal-only content.

- [X] Board at `Columns: Status`, `Rows: Repository`
exec-d4f497d5-1cc3-480b-bf52-d794a9033d90

- [X] Sparse Branch × Branch board
exec-7fab2659-92cb-4b10-9a91-c39909f4dfdd

- [X] Settings → Machines before, during, and after Retry
[...]

PR #211, opened 2026-09-18
Redesign session views for faster triage

milanshen:
## Context

Session-heavy environments need a faster way to scan activity across agents, repositories, folders, and machines. This redesign introduces focused list and board views in a shared page layout.

## Changes

- Shared page layout: top navigation, sidebar filters, search, time-range and view controls.
- Compact list view alongside the board view, surfacing readiness, activity, usage, remote-machine state, and review actions.
- Sidebar facets derived from the sessions present — agents from the vendors in the list, not the reviewer CLIs installed locally — with view, filter, density, and sort preferences persisted per browser session.
- Grouping by detected repository or folder; a root-level `cwd` gets its own group. Local-only folders key on the repository root recovered from `cwd`, not the basename, so two folders of the same name stay distinct and are labelled by their parent.
- Header and per-status rollups sum `sessionsForAggregates(...)` with `sumKnown`, matching `SessionBoard`; repository groups key on the full canonical remote so same-named forks stay distinct.
- Inspector docks at ≥1720px and no longer dismisses on outside clicks; one `(min-width: 1720px)` query drives both JS and CSS.
- Theme saves route through `initialSettingsDraft()` so a first-run save can't persist unsanitized synthesis defaults.
- Machine status reuses the former `MachineActivity` logic from `lib/machine-status.ts`: the sidebar dot carries its tone (amber for `limited`/`stale`, clay for a failed connector, pulsing while checking), and the banner is narrowed to a failed connector or an outright error. Diagnostics status folded into its `aria-label`.
- Retry is offered only where a refresh can help: an offline host gets a focusable retry button beside its facet (a sibling control, not a click handler nested inside the filter button), while a failed connector gets an Open Settings action, since `Manager.Retry` never runs the consented setup flow. A failed credential keeps the refresh path — Settings has no re-test for a configured host, so routing it there would end at a screen with only Remove.
- Review is offered only for a local session with a recorded working directory; without one the request would fail with `409 session has no working directory`.
- A failed `/api/settings` load raises its own banner and disables the theme controls, which previously looked live while the handler returned early.
- `Start fresh` reports the condition that triggered it — context pressure, compaction count, or commits behind — rather than always reporting context.
- Sort keeps stale rows below current ones, and A-Z sorts on the title the row actually renders; search matches that same title, so an unnamed session is findable by its first prompt. The field's accessible label and empty state name the fields actually searched rather than promising outcomes, files and context.
[...]

PR #205, opened 2026-09-17
Add explicit full-session sharing review

luhe19001:
## Context

Complete cached Codex revisions can contain prompts, commands, paths, subagent output, and file-change bodies. Authors need to inspect the exact content, destination, and audience before approving a full-session upload.

This PR contains the frontend review flow and builds on the collector preview and upload API in the preceding PR. Existing v1 sharing remains unchanged.

## Changes

- offer full sharing only for eligible exact SSH Codex revisions
- render the complete canonical envelope, destination, audience, revision, repository, byte counts, hash, and embedded-secret warning before approval
- bind approval to the reviewed content and invalidate it when relevant session or destination state changes
- preserve the reviewed request and idempotency key for retryable outcomes, then show the canonical accepted revision route

## Test

- `cd frontend && npm test` — 140 tests passed
- `cd frontend && npm run lint` — passed with existing warnings
- `cd frontend && npm run format:check` — passed
- `cd frontend && npm run build` — passed with existing chunk-size advisory
- manual desktop and 390×844 keyboard flow — zero requests before approval and one validated request after approval

### Screenshots

- [ ] Complete full-session review — desktop and 390×844
- [ ] Accepted exact revision result

PR #228, opened 2026-09-21
Strengthen agent guidance and make it load in every agent

milanshen:
## Why

Agent-authored PRs in this repo take several review rounds. I pulled the **280 severity-tagged review findings** (33 P0, 177 P1, 70 P2) from PRs #186–#227 and checked the existing guidance against them, then closed the gaps the data showed.

Two problems: the guidance had blind spots, and most of it was not being loaded at all.

## The loading problem

`collector/CLAUDE.md` and `frontend/CLAUDE.md` held most of the substantive guidance, but only Claude Code reads `CLAUDE.md`:

| Agent | Root `AGENTS.md` | Per-area `CLAUDE.md` |
|---|---|---|
| Codex | Yes, root → cwd | **No** — reads `AGENTS.md` / `AGENTS.override.md` only |
| Cursor | Yes, root and nested | **No** — `.cursor/rules` + `AGENTS.md` |
| opencode | Yes, traversing up from cwd | **No** — `CLAUDE.md` is a fallback only when no `AGENTS.md` exists, and ours does |
| Claude Code | Only when no `CLAUDE.md` is in cwd or above | Yes, on demand |

Worse, Claude Code reads the root `AGENTS.md` *only when no `CLAUDE.md` sits in the working directory or above it*. [CONTRIBUTING.md](CONTRIBUTING.md) tells everyone to `cd collector`, and from there `collector/CLAUDE.md` suppressed the repository-wide rules entirely.

Both files are now `AGENTS.md`, which all four agents read, and `opencode.json` declares them because opencode does not discover nested instruction files on its own.

## New guidance

- **Trust boundaries** (`collector/AGENTS.md`) — untrusted text entering agent prompts, stdin over argv with `--` terminating option parsing, hardened CLI launches, sanitized logging, server-side consent bound to the exact audience. Six P0s sat in this class with no rule covering any of them: *Treat recorded review context as untrusted data* (#194), *Disable project code when launching reviewers* (#194), *Terminate option parsing before appending the prompt* (#204), *Sanitize Hub error details before logging* (#216), *Enforce share eligibility before returning the full record* and *Bind consent to the actual audience rather than its size* (#205).
- **Keep tests honest** (`AGENTS.md`) — do not delete a test to make a change pass; a rewrite carries over the invariants the replaced code guarded; no dependence on the developer's machine, installed CLIs, timing, or map order. From *`sortSessions` lost the "stale rows sort last" invariant, and the test that pinned it was deleted* (#211) and *Isolate the executable-rejection test from system Cursor installs* (#189).
- **Cost that scales with session count**, in both areas — from the quadratic board grid and per-row facet rescans (#212), and per-poll rehashing and per-click environment probes on the collector side.

## Structural fixes

- The contract README list named three of seven and omitted `snapshot/v1/`, the one that self-describes as the public contract. It now states that every `README.md` under `collector/` is normative for its boundary.
- Abstractions are anchored to real symbols — `sessionKey`, `sameSession`, `displayStatusLabel`, `boardStatusKey`, `groupSessions`, `boardGroupKey`, `sessionMatchesSearchTerm`, `internal/session/clone.go`, `internal/fullsessionrecord/adapter.go`, `internal/sessionexport` — so each rule is greppable instead of a principle.
- Renamed `## Model lifecycle and external operations explicitly`, which read as a noun phrase about AI models in a package holding `models.json` and `models.go`.
- Unwrapped the hard-wrapped bullets to match every surrounding file.

## Verification

- `npm run format:check` passes from `frontend/` (Prettier lints `frontend/AGENTS.md`; `proseWrap` defaults to `preserve`, so unwrapping is safe).
- All 17 referenced paths and symbols machine-checked against `main` — including `lib/session-grouping.ts`, which only exists after rebasing onto `7012b1b`.
- `opencode.json` validates and both declared paths resolve.
- Markdown and config only; no Go or TypeScript touched.

## Not covered here

The `models.json` refresh loop (#224 merged → broke CI on main → reverted by #226 → #227 re-raising it) is a generator gap: `collector/scripts/models.jq` has no override overlay, so repo-owned pricing keeps getting deleted. No wording in these files prevents it, so it is tracked separately.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


frontend/AGENTS.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Working from designs
+
+- Treat HTML mockups as visual references, not implementation templates. Match the requested visual contract without copying prototype chrome, sample data, or interactions that are outside the task.
+- For pixel-matching work, render the reference and implementation at the same viewport. Verify computed dimensions and the relevant responsive breakpoints in addition to inspecting screenshots.
+- After the first correct render, perform a simplification pass before running checks. Pay particular attention to net-new CSS and one-use abstractions.
+
+## Verification
+
+- Use the repository's installed Prettier rather than manually tuning wrapping or indentation. Run `npm run format`, inspect the resulting diff, then run `npm run format:check`.
```
luhe19001, line 51: **[Optional] Scope the mutating formatter to task-owned files**

Root `AGENTS.md` requires agents to “Preserve concurrent or unrelated changes,” but this line mandates `npm run format`, which expands to `prettier --write .`. If the frontend contains an unrelated unformatted edit, following this instruction modifies that work before the agent can inspect the resulting diff. Format only task-owned files, or require a clean frontend worktree before running the whole-tree formatter.

frontend/AGENTS.md
```diff
@@ -0,0 +1,52 @@
...
+- Verify frontend changes with `npm run lint`, `npm test`, `npm run build`, and `npm run format:check` from `frontend/`. If dependencies are unavailable and the exact formatter cannot run, leave formatting unchanged instead of guessing and report the missing verification.
```
luhe19001, line 52: **[Optional] Run the full-tree formatting check only once**

Line 51 already requires `npm run format:check`, and this verification list requires the identical full-tree scan again. A compliant agent therefore performs two equivalent Prettier checks for every frontend change without an intervening source-writing operation. Keep a single `format:check` invocation at the end of the verification sequence.

luhe19001 (commented):
### Review 1 Summary

This PR successfully consolidates agent guidance into scoped `AGENTS.md` files and preserves the deleted `CLAUDE.md` rules. The frontend verification workflow needs adjustment before merge to avoid touching state outside the task.

Review profile: Helu

luhe19001 (approved):

calvintvu (approved):

PR #231, opened 2026-09-21
Add cancelable collection primitives

calvintvu:
## Context

Shared transcript parsing and repository probes need context-aware entry points so callers can stop expensive work when a request is canceled.

## Changes

- Add context-aware filesystem discovery, JSONL parsing, and bounded concurrent parsing while preserving existing compatibility wrappers.
- Add cancelable repository history, branch, and filesystem probes.
- Add deterministic tests covering cancellation during parsing and source traversal.

## Test

- `go test ./... -count=1` - passed
- `go vet ./...` - passed


collector/internal/vendors/readsource.go
```diff
@@ -261,6 +292,9 @@ func ParseJSONLSource[T any](source ReadSource, path string) ([]T, error) {
 	decoder := json.NewDecoder(file)
 	var records []T
 	for {
+		if err := ctx.Err(); err != nil {
```
milanshen, line 298: **[P1] Make blocking source I/O cancelable**

The new APIs only poll `ctx` around `ReadSource.Stat`, `ReadDir`, `Open`, and `decoder.Decode`; those operations themselves remain context-free. If an SFTP-backed source or network-mounted file stalls during `Open`, `ReadDir`, or `Decode`, cancellation never reaches this check, and `ParseFilesContext` waits for the blocked worker indefinitely. The new decode test uses an in-memory reader and cancellation triggered by polling, so it does not exercise this case. Add context-aware source operations or arrange for cancellation to close in-flight readers and interrupt directory operations.

collector/internal/vendors/readsource.go
```diff
@@ -269,6 +303,9 @@ func ParseJSONLSource[T any](source ReadSource, path string) ([]T, error) {
 			}
 			return nil, fmt.Errorf("%w: %w", ErrInvalidData, err)
 		}
+		if err := ctx.Err(); err != nil {
```
milanshen: **[P2] Check cancellation before classifying decode errors**

If the context is canceled while `Decode` is running and the reader then returns `io.EOF` or `io.ErrUnexpectedEOF`, the error branch breaks before this post-decode check and returns accumulated records with a nil error. Other decode errors similarly mask cancellation as `ErrInvalidData`. A reader that cancels the context immediately before returning EOF reproduces this deterministically. Check `ctx.Err()` immediately after `Decode`, before classifying its result.

collector/internal/vendors/readsource.go
```diff
@@ -240,8 +261,11 @@ func walkReadSourceEntry(
 	}
 	sort.Slice(entries, func(i, j int) bool { return entries[i].Name() < entries[j].Name() })
 	for _, child := range entries {
+		if err := ctx.Err(); err != nil {
```
milanshen, line 267: **[P2] Recheck cancellation after an empty directory read**

The post-`ReadDir` cancellation check only runs inside the child loop. If `ReadDir` completes with an empty listing after the context has been canceled, the loop is skipped and `ScanSourceContext` reports a successful empty scan. Check `ctx.Err()` immediately after `ReadDir` succeeds so cancellation is propagated even for empty directories.

milanshen (commented):
### Review 1 Summary

**Active findings:** 0 blocking (P0), 1 should-fix (P1), 2 nice-to-have (P2) — new and unresolved prior findings.

This PR adds context-aware Git probes, source discovery, JSONL parsing, and concurrent parsing while preserving existing APIs. The worker-pool direction is sound, but blocking source I/O can still defeat cancellation, and terminal I/O paths can incorrectly report canceled work as successful.

Review profile: Milan

collector/internal/vendors/readsource.go
calvintvu: Partially agree. The generic ReadSource interface cannot interrupt an arbitrary in-flight OS filesystem syscall. The production SFTP source is already interruptible at the transport boundary: OpenSession derives the SSH process from the caller context with exec.CommandContext, so cancellation kills SSH and closes the SFTP pipes used by Open, ReadDir, Stat, and Decode. Wrapping local calls in goroutines would only leak blocked work, and a context-shaped interface would not make os.Stat cancellable. I did fix the two verified cancellation-masking gaps in follow-up commit 6266bf3d and added deterministic regressions. Full collector tests (713) and go vet pass.

collector/internal/vendors/readsource.go
calvintvu: Agree. Follow-up commit 6266bf3d checks ctx.Err immediately after Decode, before EOF or invalid-data classification. TestParseJSONLReturnsCancellationWhenDecodeReturnsEOF reproduces the masked-cancellation case and now passes.

collector/internal/vendors/readsource.go
calvintvu: Agree. Follow-up commit 6266bf3d checks ctx.Err immediately after a successful ReadDir, including an empty result. TestWalkReadSourceReturnsCancellationAfterEmptyDirectoryRead covers this case.

calvintvu:
@agent-review

collector/internal/vendors/readsource.go
luhe19001: **[P1 — CLOSED]** The generic interface cannot preempt arbitrary filesystem syscalls, while the supported SFTP source is interrupted at the transport boundary because cancellation kills its context-owned SSH process and closes the SFTP pipes. Waiting for already-started callbacks is the documented behavior.

collector/internal/vendors/readsource.go
luhe19001: **[P2 — RESOLVED]** Fixed in `6266bf3d6ce6851b950330b91b457281a9d9d456`: `ParseJSONLSourceContext` now checks `ctx.Err()` immediately after `Decode`, before EOF or invalid-data classification.

collector/internal/vendors/readsource.go
luhe19001: **[P2 — RESOLVED]** Fixed in `6266bf3d6ce6851b950330b91b457281a9d9d456`: the walker now checks `ctx.Err()` immediately after every successful `ReadDir`, including empty results.

luhe19001 (commented):
### Review 2 Summary

This PR adds context-aware cancellation to Git probes, source discovery, JSONL parsing, and concurrent parsing. The follow-up correctly preserves cancellation errors, and supported remote I/O is interrupted through the context-owned SSH process. No issues found by automated review.


**Prior findings:** 2 resolved, 1 closed, 0 deferred, 0 open.
**This review:** 2 resolved, 1 closed, 0 deferred, 0 reopened, 0 new.

Review profile: Helu

Requested in [PR comment](https://github.com/centauri-ai/coslash/pull/231#issuecomment-5767046211)

milanshen (approved):

PR #233, opened 2026-09-21
Preserve repository-owned model pricing overrides

calvintvu:
## Context

Model table regeneration only retained entries from LiteLLM and OpenCode, removing repository-owned Cursor compatibility models. This caused missing pricing and context metadata for active and historical Composer sessions.

## Changes

- Preserve the four Composer compatibility entries during model regeneration.
- Update Composer 2.5 pricing, cached-input pricing, and context windows.
- Refresh the generated model table while keeping the Composer pricing regression test.

## Test

- `make models` - passed
- `go test ./internal/session -run '^TestComposer25FastPricing$' -count=1` - passed
- `make test` - passed
- `make check` - passed


collector/scripts/models.jq
```diff
@@ -39,3 +39,38 @@ with_entries(
       | with_entries(select(.value != null))
     )
   )
+| . + {
```
milanshen, line 42: **[P2] Let upstream data replace compatibility fallbacks**

`jq` object addition lets the right-hand object win, so this merge always overwrites the LiteLLM/OpenCode records built above it. If `opencode/jev-latest` is reintroduced with revised pricing or context metadata, the daily refresh will silently retain these hard-coded values, leaving estimates stale even though the refresh succeeds. Add these entries only when their keys are missing, or merge the compatibility object on the left so they remain fallbacks.

milanshen (commented):
### Review 1 Summary

**Active findings:** 0 blocking (P0), 0 should-fix (P1), 1 nice-to-have (P2) — new and unresolved prior findings.

The PR preserves six legacy model records and refreshes generated pricing data; the compatibility approach is sound overall. One nonblocking issue remains: local defaults override any future upstream reappearance, potentially retaining stale price or context metadata.

Review profile: Milan

luhe19001 (approved):

collector/scripts/models.jq
calvintvu: compatibility fallbacks are strictly for models where we have no upstream source (ie proprietary cursor models). daf70c0074258bda85d5d3fbb95272bf191118d9 updated models.jq to only include cursor models

PR #236, opened 2026-09-21
Make Cursor collection cancelable

calvintvu:
## Context

Cursor collection must stop transcript discovery, parsing, live-state lookup, and metadata enrichment when its caller is canceled.

## Changes

- Propagate context through Cursor discovery, parsing, live-state lookup, metadata, and source collection.
- Add deterministic cancellation coverage for Cursor parsing and metadata operations.

## Test

- `go test ./... -count=1` - 688 passed


collector/internal/vendors/cursor/source.go
```diff
@@ -373,7 +464,7 @@ func selectCursorFilesSourceWithMetadata(source vendors.ReadSource, files []stri
 			}
 			return modified
 		})
-	return selected
+	return selected, ctx.Err()
```
luhe19001: **[P2] Check cancellation during candidate limiting**

When more than 2,000 files are eligible, `LimitNewestFileFamilies` performs a `Stat` for every file plus grouping and sorting without checking `ctx`. If cancellation occurs during that pass, collection continues potentially thousands of filesystem operations before this final check returns the error, which can noticeably delay a disconnected request on a large or slow home directory. Use a context-aware limiter, or precompute modification times with cancellation checks between files.

luhe19001 (commented):
### Review 1 Summary

**Active findings:** 0 blocking (P0), 0 should-fix (P1), 1 nice-to-have (P2) — new and unresolved prior findings.

The context-aware collection approach is sound and preserves existing best-effort metadata behavior. One nonblocking gap can delay cancellation while large candidate sets are limited.

Review profile: Helu

collector/internal/vendors/cursor/source.go
calvintvu: Agree. Follow-up commit 75754161 adds a context-aware shared family limiter and uses it from Cursor. The regression test cancels during the second file-stat pass and verifies no further Stat calls occur.

milanshen (approved):

PR #238, opened 2026-09-21
Cancel session collection when requests disconnect

calvintvu:
## Context

Canceled session requests can leave collection, finalization, and repository probes running after the client disconnects. This change connects request cancellation to the shared and vendor-specific cancellation layers introduced by the preceding PRs.

## Changes

- Pass the HTTP request context through local session collection.
- Stop collection and finalization loops when cancellation is observed.
- Cancel repository probes and discard incomplete results.
- Avoid reporting request cancellation as an ordinary collection failure.
- Add deterministic coverage for HTTP disconnection and collection cancellation.

## Test

- `go test ./... -count=1` - 695 passed
- `go vet ./...` - passed
- Targeted race suite - passed
- Manual: canceled all-history requests through the paired in-app browser and verified that a subsequent bounded request completed normally.

### Screenshots

| Canceled Share request | Windowed board afterward |
| --- | --- |
| Share request canceled during collection | Windowed board remains healthy after cancellation |

| Share: 7 days | Share: all history |
| --- | --- |
| Share dialog with bounded 7-day results | Share dialog with completed all-history results |


collector/internal/collector/collector.go
```diff
@@ -232,8 +249,24 @@ func finalizeLocalSessions(
...
+	parsed []*vendors.ParsedSession,
+	metadata map[string]*vendors.SessionMetadata,
+	preserveSubagentText bool,
+) ([]*vendors.ParsedSession, error) {
+	roots, err := finalizeSessionsSourceContext(ctx, parsed, metadata, vendors.LocalReadSource, true, true, true)
+	if err != nil {
+		return nil, err
+	}
 	for _, root := range roots {
+		if err := ctx.Err(); err != nil {
```
milanshen, line 267: **[P2] Make detail-revision generation cancellable**

For a large session, cancellation immediately after this check still leaves `LocalDetailRevision` cloning the full session, hashing every change, and marshaling the complete payload without observing the context. Add a context-aware revision path so a disconnected list request can stop this CPU and allocation work.

collector/cmd/coslash/api.go
```diff
@@ -92,19 +93,28 @@ func handleList(
...
+		if r.Context().Err() != nil {
+			return
+		}
 		session.Synthesis = mgr.Lookup(session.Agent, session.ID, session.LastActivityTime)
 		state := reviewManager.Status(reviewpkg.Key(session.Agent, session.ID))
 		session.ReviewPending = state.Pending
 		session.ReviewError = state.Error
 	}
 	if r.URL.Query().Get("sourceAware") != "1" {
+		if r.Context().Err() != nil {
```
milanshen, line 115: **[P2] Stop JSON encoding when the request is canceled**

If the browser aborts after this check or while `writeJSON` is running, `json.Encoder.Encode` still marshals the complete session response before the failed write exposes the disconnect. Stream large list responses with context checks between entries, or otherwise make response serialization cancellation-aware; the source-aware response has the same gap.

milanshen (commented):
### Review 1 Summary

**Active findings:** 0 blocking (P0), 0 should-fix (P1), 2 nice-to-have (P2) — new and unresolved prior findings.

The PR soundly propagates request cancellation through collection and probing. Two low-severity gaps can still consume resources after cancellation for large responses.

Review profile: Milan

collector/cmd/coslash/api.go
calvintvu: Agree. Follow-up commit 8f3ba2b2 streams both plain and source-aware list responses one session at a time with context checks between entries. Tests verify cancellation stops before the next entry and both JSON response shapes remain compatible.

collector/internal/collector/collector.go
calvintvu: Agree. Follow-up commit 8f3ba2b2 adds LocalDetailRevisionContext, observes cancellation while assigning change IDs and normalizing subagents, checks around clone and marshal work, and propagates cancellation from finalization. A focused regression cancels during change hashing.

milanshen (approved):

